PR #12, opened 2021-02-23
Fix typos in README.md

dakerfp:


scardracs:
You may want to close that PR as README.md is a lot different now.

Issue #8, opened 2021-02-13
Choose a license


nico-abram:
I apologize if you already have a license and I missed it (I couldn't find one)

rui314:
I haven't made my mind yet as to the license of this software. I want to make money from this software if I can, but at the same time I want this to be used as widely as possible. Ideally, if I can find a sponsor who will "purchase" this entire program (not a copy of it) and release it under the MIT license or something, it would be great. But if I release this under the MIT license from the beginning, no one has to do that in the first place. So, this is not an easy problem for me.

At this moment, I'm not too worried about not choosing a license. License matters when a program becomes something that people want to use, and IMO mold hasn't reached that point. This is still just my hobby, experimental project.

xyproto:
I can't package `mold` for Arch Linux unless it has a license.

rui314:
Beside the license issue, I believe it is too early to create a package for mold. It's too experimental.

tux3:
A note to any hypothetical person in the business of pedantically following copyright law: according to [GitHub's choosealicense.com](https://choosealicense.com/no-permission/) and [this StackOverflow Q/A](https://opensource.stackexchange.com/questions/1720), the GitHub ToS only guarantee that the code can be viewed publicly, and forked _on GitHub_. 

In theory mondane things like making a local git clone or running the program are not allowed without acquiring a personal license.

Of course there are disturbing reports that international copyright law silliness rarely stops individual users, but I couldn't possibly confirm or deny that =)

xyproto:
In addition to this, Linux distros need to be careful not to become Pirate Linux distros. A license file helps a lot.

ssokolow:
Just a data point here. I'm very particular about using only free/libre tooling (ie. tooling under licenses that meet criteria like the Debian Free Software Guidelines) for my modern projects, so I'm waiting to see what license this winds up under before touching the code at all.

(I say "my modern projects" because I made an exception for Open Watcom C/C++'s "too copyleft" license for DOS hobby projects and I'm still split on whether to do my Windows 3.1 retro-hobby programming exclusively in Open Watcom C/C++ or to use my New Old Stock copy of Borland Delphi 1.x for rapid prototyping and risk not having time to rewrite in C when I release the source for others hobbyists to hack on.)

rui314:
[...]

PR #13, opened 2021-02-23
Gurantee user's rights

CrazyPython:
License it under the GNU General Public License Version 3 so that users always have the right to run, understand, modify, and contribute to the software. Make people give back.

This does not require that programs that invoke `mold`, such as compilers, to be GPL-licensed. It also permits being distributed with permissively-licensed software, such as part of `clang`.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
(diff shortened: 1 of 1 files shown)


Tarliton:
https://github.com/rui314/mold/issues/8#issuecomment-778651932

seanpianka:
Instead of GPL3, why not a more permissive license such as Apache or MIT? Especially if this is a hobby project.

CrazyPython:
> Instead of GPL3, why not a more permissive license such as Apache or MIT? Especially if this is a hobby project.

Do you want an entity to take `mold` and make a proprietary version? I don't. GPL doesn't "infect" other programs that invoke `mold` with the GPL. It merely keeps mold itself open-source.

rui314:
I settled with the GNU AGPLv3 for now, with an option to change it to the MIT license in the future.

PR #24, opened 2021-03-01
Handle many sections

Flakebi:
From the manpage:

If the number of entries in the section header table is
larger than or equal to SHN_LORESERVE (0xff00), e_shnum
holds the value zero and the real number of entries in the
section header table is held in the sh_size member of the
initial entry in section header table.  Otherwise, the
sh_size member of the initial entry in the section header
table holds the value zero.

Compiling the PassBuilder from LLVM results in 66135 sections for me.

Linking [PassBuilder.cpp.o](https://flakebi.de/flakebi/mold-PassBuilder.cpp.o.xz) still fails with this change, but it gets a little further:
```python
compiler/llpc/llvm/lib/libLLVMPasses.a(PassBuilder.cpp.o): SHT_SYMTAB_SHNDX section is not supported
```

### DIFF
--- a/object_file.cc
+++ b/object_file.cc
@@ -58,6 +58,11 @@ InputFile::InputFile(MemoryMappedFile *mb) : mb(mb), name(mb->name) {
 
   u8 *sh_begin = mb->data() + ehdr.e_shoff;
   u8 *sh_end = sh_begin + ehdr.e_shnum * sizeof(ElfShdr);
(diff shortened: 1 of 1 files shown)


rui314:
I couldn't review commits until I land a CONTRIBUTORS file to the project. I'm sorry about that. I independently found the same issue and already submitted a fix, so I'm closing this pull request.

Issue #25, opened 2021-03-17
Misc. build issues


deliciouslytyped:
# Failure to clone xxHash repo
```
$ nix-shell -p cmake git clang_10 xclip --run 'pth="$(mktemp -d)"; echo "$pth" | xclip -i -selection clipboard -f; pushd "$pth"; echo Working in: "$pth"; git clone --recursive https://github.com/[67/521]
old.git; cd mold; make submodules; make'                                                                                
/run/user/1000/tmp.HGvrS5rbhx                       
/run/user/1000/tmp.HGvrS5rbhx /tmp/tmp.tMWaLqN3Se                                                                                                                                                                                             
Working in: /run/user/1000/tmp.HGvrS5rbhx
Cloning into 'mold'...                                                                                                                                                                                                                        
remote: Enumerating objects: 591, done.
remote: Counting objects: 100% (591/591), done.
remote: Compressing objects: 100% (333/333), done.                                                                                                                                                                                            
remote: Total 7737 (delta 417), reused 423 (delta 258), pack-reused 7146
Receiving objects: 100% (7737/7737), 1.32 MiB | 6.54 MiB/s, done.                                                                                                                                                                             
Resolving deltas: 100% (5655/5655), done.                                                                              
Submodule 'mimalloc' (https://github.com/microsoft/mimalloc.git) registered for path 'mimalloc'
Submodule 'oneTBB' (https://github.com/oneapi-src/oneTBB.git) registered for path 'oneTBB'
Submodule 'xxHash' (git@github.com:Cyan4973/xxHash.git) registered for path 'xxHash'                                   
Cloning into '/run/user/1000/tmp.HGvrS5rbhx/mold/mimalloc'...
remote: Enumerating objects: 69, done.                                                                                                                                                                                                        
remote: Counting objects: 100% (69/69), done.        
remote: Compressing objects: 100% (50/50), done.                                                                       
remote: Total 8395 (delta 41), reused 37 (delta 19), pack-reused 8326        
Receiving objects: 100% (8395/8395), 3.07 MiB | 6.59 MiB/s, done.
Resolving deltas: 100% (6460/6460), done.      
Cloning into '/run/user/1000/tmp.HGvrS5rbhx/mold/oneTBB'... 
remote: Enumerating objects: 95, done.                                                                                 
remote: Counting objects: 100% (95/95), done.                                                                          
remote: Compressing objects: 100% (89/89), done.        
[...]

PR #35, opened 2021-05-05
feature: add Dockerfile

ZaMaZaN4iK:
Hi!

Thanks for the blazing fast linker :)

I just found that it can be a little bit difficult for other users to build the linker from scratch (low-speed machines, lack of experience, wasting time for building, etc), so I've prepared the Docker image with `mold`.

I completely understand that can be some overhead with starting the `mold` binary in the container.

The Dockerfile is based on this one: https://github.com/rui314/mold/issues/25#issuecomment-812869216

I hope it will be merged since for you and other users will be easier for making their own Docker images or use an existing one.

Some possible ways for customizations:
* Use Alpine or Distroless as a base image instead of Debian Slim
* Prepare Docker image for released `mold` versions (even better will be integrating it into a CI)

I hope it'll help you.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,18 @@
+FROM ubuntu:20.04 as builder
+
+ENV DEBIAN_FRONTEND noninteractive
+RUN DEBIAN_FRONTEND=noninteractive apt-get update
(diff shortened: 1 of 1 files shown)


ZaMaZaN4iK:
Also probably will be nice to mention this Dockerfile somewhere in README :)

rui314:
I couldn't review commits until I land a CONTRIBUTORS file to the project. I'm sorry about that. I independently implemented a similar feature as `build-static.sh`, so I'm closing this pull request.

Issue #21, opened 2021-02-26
Project cannot be built with both recent gcc and clang


tesuji:
I cannot build this project with both gcc and clang on latest commit as this time: 66307818eebfe471213bd1b9b27a73b504b8ad13.

I'm happy to answer more questions about my build environment if that helps.

## Expected Behavior
Build shall pass.

## Current Behavior
Build fails.

With clang:
<details>`make -j32`

```c++
archive_file.cc:39:17: error: no matching constructor for initialization of 'std::string' (aka 'basic_string<char>')
    std::string name(start, strstr(start, "/\n"));
                ^    ~~~~~~~~~~~~~~~~~~~~~~~~~~~
archive_file.cc:70:12: error: no viable overloaded '='
      name = {start, strstr(start, "/\n")};
      ~~~~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
</details>

With gcc by changing CC and CXX in makefile.
<details>`make -j32`

```c++
subprocess.cc: In function ‘std::function<void()> fork_child()’:
subprocess.cc:33:34: error: taking address of temporary array
   33 |     if (read(pipefd[0], (char[1]){}, 1) == 1)
      |                                  ^~
subprocess.cc: In lambda function:
subprocess.cc:48:44: error: taking address of temporary array
   48 |   return [=]() { write(pipefd[1], (char []){1}, 1); };
      |                                            ^~~
subprocess.cc: In function ‘bool resume_daemon(char**, i64*)’:
subprocess.cc:153:31: error: taking address of temporary array
  153 |   i64 r = read(conn, (char[1]){}, 1);
      |                               ^~
subprocess.cc: In lambda function:
subprocess.cc:218:47: error: taking address of temporary array
  218 |   *on_complete = [=]() { write(conn, (char []){1}, 1); };
      |                                               ^~~
icf.cc:69:19: warning: ‘template<class Key> class tbb::tbb_hash’ is deprecated: tbb::tbb_hash is deprecated, use std::hash [-Wdeprecated-declarations]
   69 | template<> struct tbb_hash<Digest> {
      |                   ^~~~~~~~
In file included from oneTBB/include/tbb/concurrent_hash_map.h:35,
                 from mold.h:21,
                 from icf.cc:53:
oneTBB/include/tbb/internal/_tbb_hash_compare_impl.h:86:74: note: declared here
[...]

PR #38, opened 2021-05-14
Fix string ctor param from strstr

gjimye:
When using the (begin, end) constructor for string, the types of begin
and end must match, including cv qualifiers. Depending on the library
headers used, strstr(const char *, const char *) can potentially return
char * or const char *, the former being a call to the C function strstr
and the latter being a call to C++ std::strstr.

This commit fixes the clang portion of issue #21, by explicitly casting
the return value of strstr to match the 1st param of the string ctor so
that the ctor can be instantiated regardless of the header used.

### DIFF
--- a/archive_file.cc
+++ b/archive_file.cc
@@ -37,7 +37,7 @@ read_thin_archive_members(Context<E> &ctx, MemoryMappedFile<E> *mb) {
       Fatal(ctx) << mb->name << ": filename is not stored as a long filename";
 
     const char *start = strtab.data() + atoi(hdr.ar_name + 1);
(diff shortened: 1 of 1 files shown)


rui314:
Thank you for your patch! Could you read https://github.com/rui314/mold/blob/main/CONTRIBUTING.md and add a Signed-off-by line to your commit message?

gjimye:
> Thank you for your patch! Could you read https://github.com/rui314/mold/blob/main/CONTRIBUTING.md and add a Signed-off-by line to your commit message?

Done, thanks!

rui314:
Thanks! I merged your commit. You are the first contributor to this project!

PR #44, opened 2021-06-01
Fix bug in sort_symbols. Likely leftover from a previous code refactor.

3a2l:
The code previously used `symbols` directly, for which `begin() + 1` makes sense. Now that `vec` is `symbols[1, size]`, this `+1` should be dropped.

### DIFF
--- a/output_chunks.cc
+++ b/output_chunks.cc
@@ -864,11 +864,11 @@ void DynsymSection<E>::sort_symbols(Context<E> &ctx) {
     vec[i] = {symbols[i], i, 0};
 
   // In any ELF file, local symbols should precede global symbols.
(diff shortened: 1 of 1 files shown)


rui314:
The first entry of an ELF symbol table is always a null symbol which has zero values in all fields. I think +1 is to skip that null symbol.

3a2l:
You're right, this is not a correct fix for this.

However, I do note that when `symbols` is empty (e.g. `basic.sh` test), this isn't strictly correct. As I am compiling on MSVC, this causes the iterator checks to fail (setting `_ITERATOR_DEBUG_LEVEL` to 0 "fixes" this).

I will close this request, and maybe reopen when I have a better fix. Thanks!

PR #47, opened 2021-06-06
Refactor sort symbols to deal with empty symbol table

gjimye:
PR #44 highlights that most generic algorithms use `first_it != last_it`
to determine when to stop -- this breaks if `first_it > last_it`.

This commit refactors the function to do noops when the symbol table is
empty. It also experimentally leverages C++17 Parallel STL using TBB to
replace the explicit calls to TBB parallel algorithms, and in doing so
removing the need to use a temporary vector.

I don't have a machine that has enough parallelism to benchmark this,
but it would be interesting to compare this implementation against the
existing one. If this approach performs well, it could pave the way to
make linking to TBB optional, making the algorithms fall back to their
sequenced versions where TBB is not available.

Signed-off-by: Jim Ye <gjimye@gmail.com>

### DIFF
--- a/output_chunks.cc
+++ b/output_chunks.cc
@@ -875,58 +875,54 @@ template <typename E>
 void DynsymSection<E>::sort_symbols(Context<E> &ctx) {
   Timer t(ctx, "sort_dynsyms");
 
-  struct T {
-    Symbol<E> *sym;
-    i32 idx;
-    u32 hash;
-
-    bool is_local() const {
-      return sym->esym().st_bind == STB_LOCAL;
-    }
-  };
-
-  std::vector<T> vec(symbols.size());
-
-  for (i32 i = 1; i < symbols.size(); i++)
-    vec[i] = {symbols[i], i, 0};
+  // The first entry of an ELF symbol table is a null symbol
+  const auto first_nonnull_symbol =
+      std::min(std::begin(symbols) + 1, std::end(symbols));
 
   // In any ELF file, local symbols should precede global symbols.
-  tbb::parallel_sort(vec.begin() + 1, vec.end(), [](const T &a, const T &b) {
-    return std::tuple(a.is_local(), a.idx) < std::tuple(b.is_local(), b.idx);
-  });
-
-  auto first_global = std::partition_point(vec.begin() + 1, vec.end(),
-                                           [](const T &x) {
-    return x.is_local();
-  });
+  constexpr auto is_local = [](const auto &sym) {
+    return sym->esym().st_bind == STB_LOCAL;
+  };
+  const auto first_global = std::stable_partition(
+      std::execution::par, first_nonnull_symbol, std::end(symbols), is_local);
 
   // In any ELF file, the index of the first global symbols can be
   // found in the symtab's sh_info field.
-  this->shdr.sh_info = first_global - vec.begin();
+  this->shdr.sh_info = std::distance(std::begin(symbols), first_global);
 
   // If we have .gnu.hash section, it imposes more constraints
   // on the order of symbols.
   if (ctx.gnu_hash) {
-    i64 num_globals = vec.end() - first_global;
+    const auto num_globals = std::distance(first_global, std::end(symbols));
     ctx.gnu_hash->num_buckets = num_globals / ctx.gnu_hash->LOAD_FACTOR + 1;
-    ctx.gnu_hash->symoffset = first_global - vec.begin();
+    ctx.gnu_hash->symoffset = std::distance(std::begin(symbols), first_global);
 
-    tbb::parallel_for_each(first_global, vec.end(), [&](T &x) {
-      x.hash = djb_hash(x.sym->name()) % ctx.gnu_hash->num_buckets;
-    });
-
-    tbb::parallel_sort(first_global, vec.end(), [&](const T &a, const T &b) {
-      return std::tuple(a.hash, a.idx) < std::tuple(b.hash, b.idx);
-    });
+    const auto compare_hash = [num_buckets = ctx.gnu_hash->num_buckets](
+                                  const auto &sym1, const auto &sym2) {
+      return djb_hash(sym1->name()) % num_buckets <
+             djb_hash(sym2->name()) % num_buckets;
+    };
+    std::stable_sort(std::execution::par, first_global, std::end(symbols),
+                     compare_hash);
   }
 
   ctx.dynstr->dynsym_offset = ctx.dynstr->shdr.sh_size;
 
-  for (i64 i = 1; i < symbols.size(); i++) {
-    symbols[i] = vec[i].sym;
-    symbols[i]->set_dynsym_idx(ctx, i);
-    ctx.dynstr->shdr.sh_size += symbols[i]->name().size() + 1;
-  }
+  constexpr auto symbol_name_len = [](const auto &sym) {
+    return sym->name().size() + 1;
+  };
+  ctx.dynstr->shdr.sh_size += std::transform_reduce(
+      std::execution::par_unseq, first_nonnull_symbol, std::end(symbols),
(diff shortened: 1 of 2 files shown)


output_chunks.cc
```diff
@@ -875,58 +875,52 @@ template <typename E>
...
-
-  for (i32 i = 1; i < symbols.size(); i++)
-    vec[i] = {symbols[i], i, 0};
+  // The first entry of an ELF symbol table is a null symbol
+  const auto first_nonnull_symbol =
+      std::min(std::begin(symbols) + 1, std::end(symbols));
 
   // In any ELF file, local symbols should precede global symbols.
-  tbb::parallel_sort(vec.begin() + 1, vec.end(), [](const T &a, const T &b) {
-    return std::tuple(a.is_local(), a.idx) < std::tuple(b.is_local(), b.idx);
```
gjimye, line 895: Does this do what the comment says it does? Because `cmp( (1, idx) , (0, idx)` puts the local symbols _after_ the global symbols if I'm reading this correctly

rui314: This piece of code looks odd in the first place, as the dynamic symbol table contains only global symbols. I don't think that putting any local symbol to .dynsym makes sense.

rui314 (commented):
This patch contains two logical changes -- one is to fix an out-of-bound access to an empty vector, and the other is to introduce C++17 Parallel STL. They should be separated into two physical patches.

As to Parallel STL, how mature are std c++ libraries? Intel TBB has been battle-tested parallel library, and it is hard to imagine that std C++ libraries caught up so soon.

output_chunks.cc
gjimye: Not familiar with linkers (reading MattPD's resources and learning as I work on this project), but this would explain why the existing code didn't exhibit any notable defects.
We could move the local symbol logic to an debug level assert, and assume we're only dealing with global symbols

gjimye:
> This patch contains two logical changes -- one is to fix an out-of-bound access to an empty vector, and the other is to introduce C++17 Parallel STL. They should be separated into two physical patches.

Agreed. Since this was discovered by @3a2l I'm happy to let him handle the out-of-bound access fix.

> As to Parallel STL, how mature are std c++ libraries? Intel TBB has been battle-tested parallel library, and it is hard to imagine that std C++ libraries caught up so soon.

These are the [implementations](https://software.intel.com/content/www/us/en/develop/articles/get-started-with-parallel-stl.html) that Intel originally developed in 2018 based off of TBB, which were [provided](https://isocpp.org/blog/2019/05/cpp17-stl-parallel-algorithms-with-gcc-9.1-and-intel-tbb-on-linux-and-macos) to gcc and clang some time afterwards. When using `-ltbb` with supported compiler/stdlib, these parallel execution policies kick and presumably use some TBB backed implementation.

I don't know how mature these libraries are; however I think it's worth benchmarking, which I would do if I had access to a machine with many cores. Alas, I'm stuck with a laptop.

rui314:
> > This patch contains two logical changes -- one is to fix an out-of-bound access to an empty vector, and the other is to introduce C++17 Parallel STL. They should be separated into two physical patches.
> 
> Agreed. Since this was discovered by @3a2l I'm happy to let him handle the out-of-bound access fix.

Fixing an issue is another thing than finding it, so it doesn't have to be done by the same person. I'll commit a fix for the issue.

> > As to Parallel STL, how mature are std c++ libraries? Intel TBB has been battle-tested parallel library, and it is hard to imagine that std C++ libraries caught up so soon.
> 
> These are the [implementations](https://software.intel.com/content/www/us/en/develop/articles/get-started-with-parallel-stl.html) that Intel originally developed in 2018 based off of TBB, which were [provided](https://isocpp.org/blog/2019/05/cpp17-stl-parallel-algorithms-with-gcc-9.1-and-intel-tbb-on-linux-and-macos) to gcc and clang some time afterwards. When using `-ltbb` with supported compiler/stdlib, these parallel execution policies kick in and presumably use some TBB backed implementation ([godbolt](https://godbolt.org/z/EbP1dhdqW)).
> 
> I don't know how mature these libraries are; however I think it's worth benchmarking, which I would do if I had access to a machine with many cores. Alas, I'm stuck with a laptop.

It looks like Parallel STL doesn't provide any concurrent container such as concurrent hashmap, so we can't remove a dependency to TBB.

output_chunks.cc
```diff
@@ -875,58 +875,54 @@ template <typename E>
...
-      return std::tuple(a.hash, a.idx) < std::tuple(b.hash, b.idx);
-    });
+    ctx.gnu_hash->symoffset = std::distance(std::begin(symbols), first_global);
+
+    const auto compare_hash = [num_buckets = ctx.gnu_hash->num_buckets](
+                                  const auto &sym1, const auto &sym2) {
+      return djb_hash(sym1->name()) % num_buckets <
+             djb_hash(sym2->name()) % num_buckets;
+    };
+    std::stable_sort(std::execution::par, first_global, std::end(symbols),
```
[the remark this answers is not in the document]
rui314: Yes, there's a chance of a hash collision, as the DJB hash is not a cryptographically-safe hash function.

output_chunks.cc
gjimye: Hey, it looks like in the `i386-tls-gd` test case, `mold`@`fc5469` generates a local symbol in the executable `exe` -- this is not the case for the default clang linker
```
(base) gjye@gjye:~/clone/mold/test/tmp/i386-tls-gd$ readelf --dyn-syms exe

Symbol table '.dynsym' contains 7 entries:
   Num:    Value  Size Type    Bind   Vis      Ndx Name
     0: 00000000     0 NOTYPE  LOCAL  DEFAULT  UND 
     1: 00200570     4 OBJECT  GLOBAL DEFAULT  UND _IO_stdin_used
     2: 00000000     4 TLS     LOCAL  DEFAULT   20 bar
readelf: Warning: local symbol 2 found at index >= .dynsym's sh_info value of 1
     3: 00000000     0 TLS     GLOBAL DEFAULT  UND foo
     4: 00201010     0 FUNC    GLOBAL DEFAULT  UND printf@GLIBC_2.0 (3)
     5: 00201020     0 FUNC    GLOBAL DEFAULT  UND __libc_start_main@GLIBC_2.0 (3)
     6: 00201030     0 FUNC    GLOBAL DEFAULT  UND ___tls_get_addr@GLIBC_2.3 (2)
(base) gjye@gjye:~/clone/mold/test/tmp/i386-tls-gd$ readelf --dyn-syms exe-ld

Symbol table '.dynsym' contains 7 entries:
   Num:    Value  Size Type    Bind   Vis      Ndx Name
     0: 00000000     0 NOTYPE  LOCAL  DEFAULT  UND 
     1: 00000000     0 FUNC    GLOBAL DEFAULT  UND printf@GLIBC_2.0 (2)
     2: 00000000     0 FUNC    GLOBAL DEFAULT  UND ___tls_get_addr@GLIBC_2.3 (3)
     3: 00000000     0 NOTYPE  WEAK   DEFAULT  UND __gmon_start__
     4: 00000000     0 FUNC    GLOBAL DEFAULT  UND __libc_start_main@GLIBC_2.0 (2)
     5: 00000000     0 TLS     GLOBAL DEFAULT  UND foo
     6: 0804a004     4 OBJECT  GLOBAL DEFAULT   17 _IO_stdin_used

3a2l:
> Agreed. Since this was discovered by @3a2l I'm happy to let him handle the out-of-bound access fix.
Thank you for mentioning me, @gjimye but as @rui314 has said, finding and fixing a bug are two separate things and I'm happy if someone else fixes it.

I see that a commit has already been made on main to address this problem, I will check it later. Thanks!

gjimye:
Looks like the latest commit to `main` addresses most of these issues.
I'll put out a separate PR to parallelize the last section of the renamed `finalize` function.

PR #48, opened 2021-06-06
Refactor finalize dynsym for loop to parallelize workload

gjimye:
This commit refactors the for loop in finalize to use C++17 Parallel STL
(which is backed by TBB).

Signed-off-by: Jim Ye <gjimye@gmail.com>

### DIFF
--- a/mold.h
+++ b/mold.h
@@ -9,6 +9,7 @@
 #include <atomic>
 #include <cassert>
 #include <cstdint>
+#include <execution>
 #include <fstream>
 #include <functional>
 #include <iostream>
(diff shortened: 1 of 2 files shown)


rui314:
This part of code spends less than 1 milliseconds on my machine using a single thread for 10k dynamic symbols, so this doesn't have to be optimized. In fact, parallelizing it may make this part of code slightly slower due to the cost of inter-thread communication. I want to optimize only bottlenecks and leave other code as simple as possible to avoid premature optimizations.

FYI, this is how I find a bottleneck. We have bunch of timers at various places in mold, just like `Timer t(ctx, "DynsymSection::finalize");` in that particular function. With this instrumentation, you can see how much time each piece of code takes by running mold with `--perf` option. I find a bottleneck using the timers and optimize it if it's worth it (i.e. if the benefit of optimizing it clearly exceeds the cost of optimizing it, including the cost of additional code complexity).

PR #49, opened 2021-06-06
Add CMake support.

3a2l:
This adds basic CMake support for Mold.

I noticed that you had previously added `mimalloc` as a Git submodule but then removed it, in case the code is not a Git checkout (for example a .zip download), which I agree with.

This change will use CMake for fetching dependencies if they are needed and builds them (and Mold) using the same flags used when building using the existing Makefile. I have tested this by running the tests and they all pass (I did need to copy over `mold` and `mold-wrapper.so` from my "build" directory to the source directory because the shell scripts reference that location.

I am planning on expanding this to maybe integrate the existing tests with CTest. Also, I think this will be a good step in the direction of supporting other platforms (namely Windows). I know this is a non-goal for Mold at the moment.

Do let me know if this direction is welcome or if you would like to postpone it (or discard it all-together). Thanks!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,204 @@
+cmake_minimum_required(VERSION 3.13)
+
+project(mold
+        LANGUAGES C CXX
+        VERSION 0.1.1)
+
+###############################################################################
+# Configuration options
+###############################################################################
+
+option(MOLD_ASAN "Use address sanitizer." OFF)
+option(MOLD_TSAN "Use thread sanitizer." OFF)
+
+# If this is a Git checkout, use Git to define the GIT_HASH that will be
+# appended to the version string.
+
+find_package(Git QUIET)
+
+if(Git_FOUND)
+  execute_process(
+    COMMAND ${GIT_EXECUTABLE} rev-parse --git-dir
+    WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
+    RESULT_VARIABLE   NOT_GIT_DIR
+    OUTPUT_QUIET
+    ERROR_QUIET)
+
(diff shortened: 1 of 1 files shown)


rui314:
This is an interesting experiment, but I don't think that now is the right time to move to CMake. Make has been working fine for us, so there's no immediate reward to migrate to CMake, though it would add yet another dependency to build mold. For the same reason, I'm writing tests as shell scripts -- it may not look very sophisticated, but it "just works" and doesn't require external tools to run tests.

Once mold becomes useful not only for Linux but for other Unix-like operating systems and we need to deal with more diversified build environments, we might want to migrate to CMake. But for now, I'd like to stick with the plain Make.

3a2l:
No problem, that does make sense. I wanted to have this as an alternative method of build, the default remaining make. But like you said the benefits don't outweigh the costs just yet.

I will close this PR now, hopefully it can come in handy sometime in the future. Thanks!

rui314:
mold is quickly becoming pretty stable. Since this patch has been proposed, I tweaked project's directory structure and build scheme. Now I'd like to merge this patch to see if CMake works better than make for us.

3a2l:
This is great to hear. I have been a bit busy over the past few weeks, please let me know if it works out and I will take a look to see if there are changes/fixes that need to be made.

rui314:
I managed to port your CMakeFiles.txt to the most recent version of mold.
https://github.com/rui314/mold/commit/fe26dad74483f3bdb5b6db7f2b9a345f4b7f259c

My plan is to remove the existing Makefile and completely migrate to cmake once the new CMake config is verified to work.

PR #40, opened 2021-05-21
Correct some spelling mistakes in `README.md`

expenses:

PR #51, opened 2021-06-23
Fix a small typo

romanz:


romanz:
Already addressed in https://github.com/rui314/mold/pull/40.

Issue #56, opened 2021-06-29
Version number in Makefile needs a bump to 0.9


Maato:

rui314:
Ouch! I'll create a new release with version 0.9.1 to fix it. What a silly mistake... Thank you for pointing that out.

rui314:
Fixed in 9a09c777d9460ebe7eb498d1cc0327915f8bbcdc

PR #58, opened 2021-06-29
[Makefile] Compute MOLD_VERSION automatically

nehaljwani:

Resolves https://github.com/rui314/mold/issues/56

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,10 +3,11 @@ CXX = clang++
 
 MIMALLOC_LIB = mimalloc/out/release/libmimalloc.a
 GIT_HASH = $(shell [ -d .git ] && git rev-parse HEAD)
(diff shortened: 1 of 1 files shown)


rui314:
The directory is not guaranteed to be a git checkout, so we can't solely rely on it to get a version number. For example, it needs to work with a tarball which is created by `wget https://github.com/rui314/mold/archive/main.tar.gz`

nehaljwani:
Ah, the tarball contains the submodule now? Cool. Closing!

rui314:
mimalloc is no longer a submodule but a subtree, which is just a subdirectory in a git repository. i.e. I just copied mimalloc into my repo as a subdirectory. In hindsight, I should have done this from the beginning, as it is much easier to handle.

PR #62, opened 2021-06-30
Various fixes in Makefile

Logarithmus:
Huge thanks to you for creating this amazing program!

This PR contains various small fixes to `Makefile`, namely:
- assign `GIT_HASH` only if it hasn't been assigned yet (useful when building an extracted tarball instead of `git clone`, e. g. [Void Linux package](https://github.com/void-linux/void-packages/blob/9cd75f36185341ec664e06072c10c6fa353fa1fb/srcpkgs/mold/template))
- Respect `LDFLAGS`
- Add `SYSTEM_MIMALLOC` variable to link against system-wide installed `mimalloc` shared library (useful for packaging)
- Replace `CFLAGS` with `CXXFLAGS` (why did you even use CFLAGS, `mold` is written in C++ after all)
- Respect `$DESTDIR`, more info: https://www.gnu.org/prep/standards/html_node/DESTDIR.html (again useful for packaging)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,14 +2,14 @@ CC = clang
 CXX = clang++
 
 MIMALLOC_LIB = mimalloc/out/release/libmimalloc.a
-GIT_HASH = $(shell [ -d .git ] && git rev-parse HEAD)
+GIT_HASH ?= $(shell [ -d .git ] && git rev-parse HEAD)
 
 CPPFLAGS = -g -Imimalloc/include -pthread -std=c++20 \
            -Wno-deprecated-volatile \
            -DMOLD_VERSION=\"0.9.1\" \
            -DGIT_HASH=\"$(GIT_HASH)\" \
 	   $(EXTRA_CPPFLAGS)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -51,8 +55,8 @@ endif
 
 all: mold mold-wrapper.so
 
-mold: $(OBJS) $(MIMALLOC_LIB)
```
Alcaro, line 54: Won't this make it not compile the bundled mimalloc even if it should? Maybe unset $(MIMALLOC_LIB) if using the system one, instead of this change.

Logarithmus: Hmm probably yes, I'll fix it now

rui314:
Thanks! All changes make sense to me. Can you read CONTRIBUTORS.md and add a Signed-off-by line if you agree?

Logarithmus:
@Logarithmus I agree to my code being dual-licensed over AGPL 3.0 & MIT and added sign-off.

PR #63, opened 2021-06-30
README.md: add guide to using mold with GCC via -B option

Logarithmus:
Signed-off-by: Artur Sinila <freesoftware@logarithmus.dev>


rui314:
This is extremely informative! Thank you for writing it up!

PR #68, opened 2021-07-02
[oneTBB] build only required target(s)

nehaljwani:
The 'all' target builds `tbbmalloc` and `tbbmalloc_proxy`
which mold does not care about.


rui314:
Thanks!

PR #70, opened 2021-07-03
Fix compilation error with GCC 11

nehaljwani:
```
mold.h:128:19: error: explicit specialization of
  ‘template<class Key> class tbb::detail::d1::tbb_hash_compare’
  outside its namespace must use a nested-name-specifier [-fpermissive]
  128 | template<> struct tbb_hash_compare<std::string_view> {
      |                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/mold.h
+++ b/mold.h
@@ -124,8 +124,7 @@ inline u64 hash_string(std::string_view str) {
   return XXH3_64bits(str.data(), str.size());
 }
 
-namespace tbb {
-template<> struct tbb_hash_compare<std::string_view> {
+template<> struct tbb::tbb_hash_compare<std::string_view> {
   static size_t hash(const std::string_view &k) {
     return hash_string(k);
   }
@@ -134,7 +133,6 @@ template<> struct tbb_hash_compare<std::string_view> {
     return k1 == k2;
   }
 };
-}
 
 template<typename ValueT> class ConcurrentMap {
 public:


rui314:
Is this a GCC's bug? Usually, a `detail` sub-namespace used in a header is intended to be private. Even though it is visible, no one except the library itself should directly use anything in a `detail` namespace.

If we have to do this to workaround GCC's bug, this is OK, but in general we want to avoid referreing classes in other library's `detail`.

nehaljwani:
There is a report at https://gcc.gnu.org/bugzilla/show_bug.cgi?id=92598

rui314:
Thanks for the URL!

I wonder if it still fails with this patch:

```
diff --git a/mold.h b/mold.h
index be28eb14..37d4f9c9 100644
--- a/mold.h
+++ b/mold.h
@@ -124,8 +124,7 @@ inline u64 hash_string(std::string_view str) {
   return XXH3_64bits(str.data(), str.size());
 }

-namespace tbb {
-template<> struct tbb_hash_compare<std::string_view> {
+template<> struct tbb::tbb_hash_compare<std::string_view> {
   static size_t hash(const std::string_view &k) {
     return hash_string(k);
   }
@@ -134,7 +133,6 @@ template<> struct tbb_hash_compare<std::string_view> {
     return k1 == k2;
   }
 };
-}

 template<typename ValueT> class ConcurrentMap {
 public:
```

nehaljwani:
It does! Patch updated.

PR #75, opened 2021-07-04
[tests] Improve test for -spare-dynamic-tags

nehaljwani:
The default size and offsets of the .dynamic section are not necessarily
same with different builds of clang. For example, some builds can make
clang pass --hash-style=gnu to the linker and some --hash-style=both.

This change updates the test to compute the difference in size of the
.dynamic section before and after using the flag -spare-dynamic-tags,
making it independent of the compiler build.


rui314:
I was trying to avoid doing computation in tests, but it looks like we can't avoid doing it in this case. Thank you for creating this change!

PR #77, opened 2021-07-04
Optimize CI build (enable ccache & run tests in parallel)

Logarithmus:
Some tests invoke `ld` instead of `mold` by a mistake. This creates an illusion that they run successfully for `mold`. Which isn't true for `copyrel-protected` & `ifunc-dso`.

Things we have to fix in `mold`:
- handling of `protected` when creating shared libraries
- handling of `ifunc`

Honestly, I'm not a linker expert, so don't think I can help. @rui314

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
     - name: make
       run: make -j$(nproc)
     - name: make test
-      run: make test
+      run: make test -j$(nproc)
   build-gcc11:
     runs-on: ubuntu-20.04
     container: gcc:11.1.0
@@ -25,4 +25,4 @@ jobs:
     - name: make
       run: make CC=gcc CXX=g++ -j$(nproc)
     - name: make test
-      run: make test
+      run: make test -j$(nproc)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,18 +11,30 @@ jobs:
     - uses: actions/checkout@v2
     - name: apt-get
       run: sudo dpkg --add-architecture i386 && sudo apt-get update && sudo apt-get install build-essential libstdc++-10-dev cmake clang libssl-dev zlib1g-dev libxxhash-dev libtbb-dev git bsdmainutils dwarfdump libc6-dev:i386 lib32gcc-10-dev libstdc++-10-dev-arm64-cross gcc-aarch64-linux-gnu g++-aarch64-linux-gnu qemu-user
+    - name: ccache
+      uses: hendrikmuhs/ccache-action@v1
     - name: make
-      run: make -j$(nproc)
+      run: |
+        export PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH"
+        make -j$(nproc)
     - name: make test
-      run: make test -j$(nproc)
+      run: |
+        export PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH"
+        make test -j$(nproc)
   build-gcc11:
     runs-on: ubuntu-20.04
     container: gcc:11.1.0
     steps:
     - uses: actions/checkout@v2
     - name: apt-get
-      run: dpkg --add-architecture i386 && apt-get update && apt-get install -y libssl-dev zlib1g-dev libxxhash-dev cmake clang dwarfdump bsdextrautils libgcc-9-dev gcc-aarch64-linux-gnu g++-aarch64-linux-gnu libc6-dev:i386 lib32gcc-10-dev qemu-user
+      run: dpkg --add-architecture i386 && apt-get update && apt-get install -y sudo libssl-dev zlib1g-dev libxxhash-dev cmake clang dwarfdump bsdextrautils libgcc-9-dev gcc-aarch64-linux-gnu g++-aarch64-linux-gnu libc6-dev:i386 lib32gcc-10-dev qemu-user
+    - name: ccache
+      uses: hendrikmuhs/ccache-action@v1
     - name: make
-      run: make CC=gcc CXX=g++ -j$(nproc)
+      run: |
+        export PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH"
+        make CC=gcc CXX=g++ -j$(nproc)
     - name: make test
-      run: make test -j$(nproc)
+      run: |
+        export PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH"
+        make test -j$(nproc)


.github/workflows/ci.yml
```diff
@@ -10,19 +10,48 @@ jobs:
...
+        sudo dpkg --add-architecture i386 && sudo apt-get update && \
+        sudo apt-get install build-essential libstdc++-10-dev cmake clang \
+          libssl-dev zlib1g-dev libxxhash-dev libtbb-dev git bsdmainutils \
+          dwarfdump libc6-dev:i386 lib32gcc-10-dev libstdc++-10-dev-arm64-cross \
+          gcc-10-aarch64-linux-gnu g++-10-aarch64-linux-gnu && \
+        sudo update-alternatives --install /usr/local/bin/aarch64-linux-gnu-gcc \
+          aarch64-linux-gnu-gcc /usr/bin/aarch64-linux-gnu-gcc-10 20 && \
+        sudo update-alternatives --install /usr/local/bin/aarch64-linux-gnu-g++ \
+          aarch64-linux-gnu-g++ /usr/bin/aarch64-linux-gnu-g++-10 20
+    - name: ccache
```
rui314, line 23: I don't want to use a third party tool here unless it is necessary, as I prefer keeping things as simple as possible. For example, that's why I stick with the good old make instead of a sophisticated build system.

In this case, ccache is used for making build faster on a buildbot. How much faster is this with this change? I don't care about bots performance too much.

test/as-needed-weak.sh
```diff
@@ -13,11 +13,11 @@ _start:
         call fn1@PLT
 EOF
 
-cat <<EOF | cc -o $t/b.so -shared -fPIC -Wl,-soname,libfoo.so -xc -
```
rui314, line 16: I intentionally used the default linker instead of mold to build most shared object files to make sure that mold is interoperable with the system linker.

.github/workflows/ci.yml
Logarithmus: Looking at https://github.com/rui314/mold/actions/runs/999535407, roughly 2 min 58 s vs 2 s (clang build was with hot ccache and and gcc without). 3 minutes gain is quite significant IMO. It's not too much but it adds up and speeds up iterations when doing multiple force pushes into PR branch with small fixes.
If you are concerned about using third-party Github Action, then I can implement it using bare GitHub's `cache action. Though I think it would be just a code duplication and wheel reinvention.

test/as-needed-weak.sh
Logarithmus: @rui314 I think it would be nice to test both scenarios: link shared with mold & with ld and then test.

Logarithmus:
@rui314 any ideas on how to fix those 2 failing tests?

.github/workflows/ci.yml
rui314: Hm, 3 min vs 2s. I didn't expect that the difference is so huge. I have to agree with you that this is quite significant. I'm convinced that we should use ccache.

rui314:
I have no idea why these tests are failing on CI. I checked out your branch from GitHub and ran tests locally, and all tests passed. Weird...

Logarithmus:
> I have no idea why these tests are failing on CI. I checked out your branch from GitHub and ran tests locally, and all tests passed. Weird...

These 2 tests also fail on my Void Linux. What distro do you use on your local machine? If Ubuntu (like CI) then it's even more weird.

Probably you are somehow missing `ld` symlink to `mold` linker inside your clonned repo. And `cc -B`pwd`/.. ` just silently proceeds to use `/usr/bin/ld` instead of `mold`.

Also, does mold have support for setting protected visibility in generated shared libraries at all?

nehaljwani:
In `copyrel-protected`, when `b.so` is generated with mold, the visibility for the `foo` symbol is incorrect in the `.dynsym` section. I've submitted a fix at https://github.com/rui314/mold/pull/82, but I'm not sure if it is correct.

nehaljwani:
In `ifunc-dso`, when `b.so` is generated with mold, `.dynsym` the symbol type is `FUNC`:
```
5: 000000000000112c     0 FUNC    GLOBAL DEFAULT   15 foobar
```
But with bfd:
```
6: 000000000000111c     0 IFUNC   GLOBAL DEFAULT   11 foobar
```
But with gold:
```
9: 000000000000068c     0 IFUNC   GLOBAL DEFAULT   12 foobar
```
But with lld:
```
7: 00000000000015dc     0 <OS specific>: 10 GLOBAL DEFAULT   11 foobar
```
For this, the following seems to make the test pass:
```diff
diff --git a/output_chunks.cc b/output_chunks.cc
index 2bad032..6ef4981 100644
--- a/output_chunks.cc
+++ b/output_chunks.cc
@@ -978,8 +978,6 @@ void DynsymSection<E>::copy_buf(Context<E> &ctx) {
 
     memset(&esym, 0, sizeof(esym));
     esym.st_type = sym.esym().st_type;
-    if (esym.st_type == STT_GNU_IFUNC)
-      esym.st_type = STT_FUNC;
     esym.st_size = sym.esym().st_size;
 
     if (sym.is_weak)
```
However, it contradicts the commit message in https://github.com/rui314/mold/commit/dd043573

rui314:
@nehaljwani https://github.com/rui314/mold/commit/dd043573 might have been just wrong, feel free to revert.

nehaljwani:
Opened https://github.com/rui314/mold/pull/83 to partially revert https://github.com/rui314/mold/commit/dd043573

nehaljwani:
This patch needs to be rebased now.

Logarithmus:
@nehaljwani thank you for digging into this issue.

rui314:
@Logarithmus It looks like the same tests are failing. Did you rebase?

Logarithmus:
@rui314 I didn't push yet. What should I do: make all tests use `mold` or both `ld` & `mold`? https://github.com/rui314/mold/pull/77#discussion_r663620853

Logarithmus:
@rui314 also I think I can drop this `update-alternatives` commit (https://github.com/rui314/mold/pull/77/commits/019bffe062fe99fb33817b2521b2c77a532b58da), because it turned that gcc 9.3.0 is sufficient to compile this small test.
P. S. Maybe in future we'll enable running all the tests on aarch64, which will need building mold for aarch64, and then we'll definitely have to use gcc 10, but that's sure out of the scope of this PR.

Logarithmus:
@rui314 I've pushed the changes. For now all the tests use `mold` as linker.

rui314:
@Logarithmus As to making all tests to use `mold`, I want to keep the current behavior. I.e. I want to use the system-default linker to create some .so files.

In order to test the scenario that we always use mold, we can run the same tests with `mold -run`. I mean we can make a change to `make test` so that `make test` invokes `mold -run make test` so that it tests both scenarios. But this should be done in a different commit and should be done after 0.9.2.

nehaljwani:
I am eagerly waiting for 0.9.2 :smiley: !

@Logarithmus Could you please confirm whether void-linux is happy now?

Logarithmus:
@rui314 OK, so I've dropped the commit responsible for using `mold` in all tests.
@nehaljwani it builds perfectly fine on my laptop now. Though I still need a new release to make a package.

rui314:
@Logarithmus Looks like the bots are happy with this change, so I'll merge your commits. Thank you very much for your wonderful contributions! After merging, I'll cherrypick more changes to the 0.9.2 branch and push it to GitHub so that you guys can take a final look before I tag it.

PR #79, opened 2021-07-05
[build] No need for asserts in release mode

nehaljwani:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ TSAN ?= 0
 ifeq ($(DEBUG), 1)
   CPPFLAGS += -O0
 else
-  CPPFLAGS += -O2
+  CPPFLAGS += -O2 -DNDEBUG
 endif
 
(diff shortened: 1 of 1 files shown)


rui314:
I don't think that we need to disable `assert()` in the release build, too, as they are extremely cheap at runtime. I believe you can't observe any difference in performance between builds with/without assertions. So I'd like to keep them in the release build.

nehaljwani:
Oh, it was deliberate. Thank you for clarifying!

rui314:
FYI: https://blog.cloudflare.com/branch-predictor/ claims that on modern processors, branches that are always not-taken are essentially free, so I'm guessing that that's also the case for assertions which should always yield a true value. For assertions, we do some computations, so the conditions are not exactly the same as `if (debug) ...`, but for us, expressions in `assert(...)` is usually very cheap, so the ballpark number should be the same.

PR #82, opened 2021-07-07
Ensure correct visibility of symbols in .dynsym section

nehaljwani:

### DIFF
--- a/output_chunks.cc
+++ b/output_chunks.cc
@@ -1012,6 +1012,7 @@ void DynsymSection<E>::copy_buf(Context<E> &ctx) {
     } else {
       esym.st_shndx = sym.input_section->output_section->shndx;
       esym.st_value = sym.get_addr(ctx, false);
+      esym.st_visibility = sym.visibility;
     }
   }
 }


rui314:
We put only global symbols to .dynsym, so I think all symbols should have STV_DEFAULT. Am I missing something?

nehaljwani:
Take this for example:
```zsh
/t/nwani_1625632742 ❯❯❯ echo '__attribute__((visibility("protected"))) int foo;' | clang -fuse-ld=lld -xc - -shared -o libfoo.so
/t/nwani_1625632742 ❯❯❯ readelf -a libfoo.so | grep foo
     5: 000000000000379c     4 OBJECT  GLOBAL PROTECTED   23 foo
    23: 000000000000379c     4 OBJECT  GLOBAL PROTECTED   23 foo
    
/t/nwani_1625632742 ❯❯❯ echo '__attribute__((visibility("protected"))) int foo;' | clang -fuse-ld=$HOME/repos/mold/mold -xc - -shared -o libfoo.so
/t/nwani_1625632742 ❯❯❯ readelf -a libfoo.so | grep foo
     4: 0000000000003004     4 OBJECT  GLOBAL DEFAULT   21 foo
    19: 0000000000003004     4 OBJECT  GLOBAL PROTECTED   21 foo
```

output_chunks.cc
```diff
@@ -1012,6 +1012,7 @@ void DynsymSection<E>::copy_buf(Context<E> &ctx) {
     } else {
       esym.st_shndx = sym.input_section->output_section->shndx;
       esym.st_value = sym.get_addr(ctx, false);
+      esym.st_visibility = sym.esym().st_visibility;
```
rui314, line 1015: We merge symbol visibility when we see more than one symbol (this process is done in`ObjectFile::merge_visibility`). Merged visibility is accessible as `sym.visibility`. So please use `sym.visibility` instead of `sym.esym().st_visibility`.

rui314 (changes requested):

PR #91, opened 2021-07-13
Use `cmake --build` instead of `make` for mimalloc and TBB

IndianBoy42:
This will run whichever build tool was used to generate the cmake project

For example, I set `CMAKE_GENERATOR=Ninja` globally so that cmake will automatically use Ninja always. This would break with the current Makefile as it will try to run make in that directory when there is no Makefile

The only downside is I dont know which cmake version this was added, perhaps its too new, although the project already requires c++20 and some modern tooling

This could be backported to 0.9.1 as well potentially

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,12 +75,12 @@ $(OBJS): mold.h elf.h Makefile
 $(MIMALLOC_LIB):
 	mkdir -p mimalloc/out/release
 	(cd mimalloc/out/release; CFLAGS=-DMI_USE_ENVIRON=0 cmake ../..)
-	cmake --build mimalloc/out/release mimalloc-static
+	cmake --build mimalloc/out/release --target mimalloc-static
 
 $(TBB_LIB):
 	mkdir -p oneTBB/out
 	(cd oneTBB/out; cmake -DBUILD_SHARED_LIBS=OFF -DTBB_TEST=OFF -DCMAKE_CXX_FLAGS=-D__TBB_DYNAMIC_LOAD_ENABLED=0 ..)
-	cmake --build oneTBB/out tbb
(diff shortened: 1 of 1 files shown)


rui314:
If the feature might be too new to use, how about adding `-G'Unix Makefiles'` to `cmake` invocation and continue using make? It looks like a simpler solution to me.

Alcaro:
--build is present in [2.8.0](https://cmake.org/cmake/help/v2.8.0/cmake.html#opt%3a--builddir), released November 2009. I think that's old enough.

rui314:
Does `cmake` respect `-j` option passed to its parent `make` process? As long as it works, I don't care too much about how subprojects are built, but it looks like sticking with `make` is simpler. In future, we might want to migrate to `cmake` altogether, and if we do, recursively invoking `cmake --build` would be more appearing.

IndianBoy42:
I don't think by default it does, it just uses nproc internally to find the number of jobs it should use.

Get Outlook for Android<https://aka.ms/AAb9ysg>

________________________________

nehaljwani:
The standard way to pass `-j` to the generator is:
```
cmake --build /path/to/builddir -- -jX <target-name>
```
AFAIK, this would work with both ninja and make. The generator agnostic way to do that is to set `CMAKE_BUILD_PARALLEL_LEVEL` environment variable.

rui314:
So, as I said, I'd like to keep using make and fix the issue by appending `-G'Unix Makefiles'` to the cmake invocations.

pseyfert:
FTR cmake does support `-j` since version 3.12 (nov 2018) https://cmake.org/cmake/help/v3.12/release/3.12.html#id4

PR #109, opened 2021-08-13
Explicitly instruct CMake to use Unix Makefiles

mqudsi:
mold currently builds dependencies in its `Makefile` by explicitly
issuing `make -C ...` targeting the CMake build directory; this will
fail if a global `CMAKE_GENERATOR` environment variable is present that
specifies a different CMake build file generator.

Closes #91


rui314:
Could you read https://github.com/rui314/mold/blob/main/CONTRIBUTING.md and add a `signed-off-by` line to the commit message?

mqudsi:
No problem, done.

PR #134, opened 2021-12-06
allow overriding strip

Keruspe:
Some packaging systems want to handle the stripping by themselves.
Setting e.g. STRIP=true will result in true being used instead of strip and thus will skip the stripping part, leaving it to the package manager


rui314:
Nice. Thanks!

Issue #127, opened 2021-11-13
mold: mold-wrapper.so is missing


installgentoo:
Basically you look for mold-wrapper in hardcoded ../../lib
On gentoo we have lib and lib64(honestly i don't know how exactly these rules work, has smth to do with 32 abi), mold goes into lib64.
Maybe similar situation exists on other distros.

Perhaps allow providing path to it in an argument?

rui314:
If you run `make install`, mold-wrapper.so will be installed to `$PREFIX/lib`, so I guess Gentoo modifies the default Makefile so that it is installed to `lib64`. Are you using Gentoo's mold package?

installgentoo:
Yes, it does.
"	sed -i \
		-e '/	strip/d' \
		-e '/	gzip/d' \
		-e "s:\$(DEST)/lib:\$(DEST)/$(get_libdir):" \
		Makefile || die"

Yes i do.

One thing that i got impression off, the error is confusing, only when i searched it in the sources did i realise what was happening.

rui314:
Can you report a bug to Gentoo's package? The path to search `mold-wrapper.so` is hard-coded here  (https://github.com/rui314/mold/blob/v0.9.6/elf/subprocess.cc#L275). If you modify the Makefile, you also need that line too.

In the next release, I'll add LIBDIR variable to Makefile so that you don't have to apply a local patch to the C source file.

installgentoo:
Already did.

thesamesam:
Please do cross-link bugs in future if you can (the downstream bug is [here](https://bugs.gentoo.org/823653)). Thanks for the report both downstream and here. I'll fix the Gentoo package now.

thesamesam:
I think this can be closed now. It's fixed both in Gentoo (in the earlier release and in 1.0.0 and fixed upstream in 1.0.0 too.

PR #135, opened 2021-12-14
build: respect user *FLAGS, don't compress man pages

thesamesam:
Forwarding downstream patches I've applied in Gentoo:

- Respect user CXXFLAGS
- Rename CPPFLAGS (previously used in the sense of "flags for the C++ compiler") -> CXXFLAGS
- CPPFLAGS is generally used for "flags for the C(++) preprocessor.", so let's
  use it for that
- Respect user LDFLAGS
  (In one instance, we were respecting LDFLAGS, but doing it too late.
  We need to pass LDFLAGS _before_ any objects in order for -Wl,--as-needed
  to work correctly.)
- Don't compress man pages
   (Negligible saving and downstream, distributions	usually	recompress or
    compress with their own	specific options. Unconditionally compressing
    man pages, while well intended,	usually	creates	more hassle there.)

Signed-off-by: Sam James <sam@gentoo.org>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,10 +17,15 @@ STRIP ?= strip
 
 OS ?= $(shell uname -s)
 
-CPPFLAGS = -pthread -std=c++20 -fPIE -DMOLD_VERSION=\"0.9.6\" \
-	   -fno-exceptions -fno-unwind-tables -fno-asynchronous-unwind-tables \
-	   -DLIBDIR="\"$(LIBDIR)\"" $(EXTRA_CPPFLAGS)
-LDFLAGS += $(EXTRA_LDFLAGS)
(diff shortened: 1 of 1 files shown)


rui314:
Looks great! Thank you for doing this.

thesamesam:
Thanks a lot @rui314 - for both this and mold! This will make life a lot easier downstream 💯

rui314:
Did you have anything else? If so, please submit a PR. I'll be releasing mold 1.0 soon.

thesamesam:
Nothing else - I had the (hacky) patch for #127 but you've fixed that better in git (and I've confirmed it works).

I'm hitting a build failure on master but I'll file a bug for that now.

Issue #140, opened 2021-12-15
possible missing -lrt flag in Makefile invocation


miguelraz:
I'm trying to build a mold BinaryBuilder.jl recipe so that Julia users may install this software super easily.

This involves running the entire build process inside a sandboxed linux environment and only pulling in the necessary dependencies and compiler versions.

When running `make -j${nproc}`, this error shows up

```
sandbox:${WORKSPACE}/srcdir/mold # make
c++ -DMOLD_VERSION=\"1.0.0\" -DLIBDIR="\"/usr/local/lib\"" -DGIT_HASH=\"ed9924895d9b9584106791247596677db8113528\" -Ithird-party/mimalloc/include -Ithird-party/tbb/include -O2 -pthread -fPIE -fno-unwind-tables -fno-asynchronous-unwind-tables -std=c++20 -fno-exceptions  out/compress.o out/demangle.o out/filepath.o out/hyperloglog.o out/main.o out/perf.o out/tar.o out/elf/arch-arm64.o out/elf/arch-i386.o out/elf/arch-x86-64.o out/elf/cmdline.o out/elf/gc-sections.o out/elf/icf.o out/elf/input-sections.o out/elf/linker-script.o out/elf/main.o out/elf/mapfile.o out/elf/object-file.o out/elf/output-chunks.o out/elf/output-file.o out/elf/passes.o out/elf/relocatable.o out/elf/subprocess.o out/macho/arch-arm64.o out/macho/arch-x86-64.o out/macho/cmdline.o out/macho/dead-strip.o out/macho/dumper.o out/macho/input-sections.o out/macho/main.o out/macho/mapfile.o out/macho/object-file.o out/macho/output-chunks.o out/macho/output-file.o out/macho/tapi.o out/macho/yaml.o -o mold -pthread -lz -lxxhash -ldl -lm -Wl,-whole-archive out/mimalloc/libmimalloc.a -Wl,-no-whole-archive out/tbb/libs/libtbb.a -lcrypto
/opt/x86_64-linux-gnu/bin/../lib/gcc/x86_64-linux-gnu/11.1.0/../../../../x86_64-linux-gnu/bin/ld: out/perf.o: in function `mold::TimerRecord::stop() [clone .part.0]':
perf.cc:(.text+0x27): undefined reference to `clock_gettime'
[...]

PR #141, opened 2021-12-15
add missing -lrt flag to Makefile

miguelraz:
vis #140

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ CFLAGS += $(COMMON_FLAGS)
 CXXFLAGS ?= -O2
 CXXFLAGS += $(COMMON_FLAGS) -std=c++20 -fno-exceptions
 CPPFLAGS += -DMOLD_VERSION=\"1.0.0\" -DLIBDIR="\"$(LIBDIR)\""
-LIBS = -pthread -lz -lxxhash -ldl -lm
+LIBS = -pthread -lz -lxxhash -ldl -lm -lrt
 
 SRCS=$(wildcard *.cc elf/*.cc macho/*.cc)
 HEADERS=$(wildcard *.h elf/*.h macho/*.h)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -66,7 +66,7 @@ else ifneq ($(OS), Darwin)
   else
     MIMALLOC_LIB = out/mimalloc/libmimalloc.a
     CPPFLAGS += -Ithird-party/mimalloc/include
-    LIBS += -Wl,-whole-archive $(MIMALLOC_LIB) -Wl,-no-whole-archive
+    LIBS += -Wl,-whole-archive $(MIMALLOC_LIB) -Wl,-no-whole-archive -lrt
```
rui314, line 69: Don't you want to move this to line 29 instead of this line?

rui314 (changes requested):
Please read CONTRIBUTING.md and add a signed-off-by line.
Thanks!

Makefile
miguelraz: Fair, wasn't sure where precisely. Thanks!

miguelraz:
Newest commit is done with `git commit -s`, feel free to squash and merge.

thesamesam (approved):
lgtm

rui314:
It looks like it breaks macOS build, but I'll fix that after merging this PR. Thank you for doing this!

thesamesam:
@rui314 As time goes on, I anticipate having to add/remove more of these hacks based on platform (e.g. explicit/implicit pthread and so on). How do you feel about me looking at meson? no promises, but just had an idea I've had the last few days. It'd handle stuff like this for us.

rui314:
@thesamesam At this moment, I'd like to stick with the good ol' makefile. Everybody knows how to use the make command, and the current state of our makefile isn't messy IMO. In the future, especially when porting it to non-Unix platform (i.e. Windows), we'll probably have to start using a more sophisticated, cross-platform build system such as cmake. But for now, I think makefile works fine for us.

thesamesam:
@rui314 No problem. I agree it's not really messy right now. Let me know if/when that changes. We'll see what happens if any more bug reports come in and such too. Thank you!

miguelraz:
Thank you for working on this!
I'll wanna ship this a recipe for this with BinaryBuilder.jl so that Julia users don't have to worry about installing it.

PR #147, opened 2021-12-16
fix clang pkg name in fedora instructions

Frederik-Baetens:
clang-c++ doesn't exist (at least not in fedora35)

```
~ : sudo dnf install clang-c++    
No match for argument: clang-c++
Error: Unable to find a match: clang-c++
```

I tried the installation process with just the clang package, and that works great.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ sudo apt-get install -y build-essential git clang cmake libstdc++-10-dev libssl-
 #### Fedora 34 and later
 
 ```shell
-sudo dnf install -y git clang-c++ cmake openssl-devel xxhash-devel zlib-devel
+sudo dnf install -y git clang cmake openssl-devel xxhash-devel zlib-devel
 ```
 
 ### Compile mold


rui314:
I'm not using Fedora so I can't verify if this is a correct fix.

@Calinou Do you know if clang++ is installed just by installing clang?

markusgod:
Fedora's clang package provides clang, clang++ and clang-cl.
https://fedora.pkgs.org/35/fedora-x86_64/clang-13.0.0~rc1-1.fc35.x86_64.rpm.html

rui314:
Did you know what was the case in Fedora 34?

markusgod:
Same goes for 34. https://fedora.pkgs.org/34/fedora-x86_64/clang-12.0.0-0.3.rc1.fc34.x86_64.rpm.html

rui314:
Then looks like it is ok to merge this change. Thank you for doing this and thank you for double-checking this!

PR #153, opened 2021-12-16
Force the execution of tests in English

sylvestre:
If the tests are executed in a different locale like French, some logs  (ex: ```out/test/elf/auxiliary/log```) can contain output like:

```La section dynamique à l'offset 0x2000 contient 10 entrées :```

which of course doesn't work as the grep is
```
fgrep -q 'Auxiliary library: [foo]' $t/log
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -123,10 +123,10 @@ $(TBB_LIB):
 
 ifeq ($(OS), Darwin)
 test tests check: all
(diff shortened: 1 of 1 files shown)


rui314:
Can you read https://github.com/rui314/mold/blob/main/CONTRIBUTING.md and add a `signed-off-by` line to your commit message

sylvestre:
Sure, signed-off added :)

PR #166, opened 2021-12-19
musl binaries

felixguendling:
We're using mold to build [Motis](https://github.com/motis-project) because we're hitting RAM limit of the GitHub runners with default `ld`.

I created pre-built zero-dependency binaries for this:
https://github.com/motis-project/mold/releases/tag/v1.0.0-musl-strip

Is this something you're interested in to have for the main repo? Then I would be happy to tidy this PR up. If not, it's also fine and you can just close this PR.

Thank you for providing this awesome tool - saves a lot of memory and time waiting for builds :smile:


rui314:
I'm interested in creating and distributing a statically-linked mold. There's another PR that ports my `build-static.sh` script to Alpine Linux, so that the resulting executable is statically-linked against musl instead of glibc. https://github.com/rui314/mold/pull/193 So I guess this and that PR try to solve the same problem. Is this correct?

felixguendling:
This PR also switches the build system to CMake and the dependency management to [this](https://github.com/motis-project/pkg) which is a structured, reproducible way that works on all systems and does not depend on external dependencies. However, I know that if you would have wanted to use CMake you probably would already have used it in the first place and switching to a dependency management that someone built on their own is maybe also not what you want. And since the main benefit (dependency free musl binaries in the GitHub release) can be achieved with the other PR, too, I'm closing this PR. If you think otherwise, you can always reopen and respond here. I'm happy if I can help.

felixguendling:
A .. maybe another question: the PR you mentioned seems to deliver musl binaries only for x86-64.

This PR also delivers binaries for 32bit ARM, 64 bit ARM as well as 32bit x86. The toolchain I'm using is based on crosstool-NG and is able to produce statically linked musl binaries for many more target platforms (MIPS, PowerPC, etc.) if you like.

rui314:
As to CMake, I'd like to stick with the good ol' makefile at this moment. It's working fine for us, and everybody knows how to use it. And if you take a look at the mold's code, you may notice that I'm trying to not add too many external dependencies to the codebase. If we can live without CMake, I'd choose that route. Eventually, I think we need to start using CMake, especially when we start working on Windows port of the mold linker. But I think it's not the right time to make the switch.

On the second point, I think it's great to release binary packages for multiple architectures. I'm not sure if there's a need for ARM32 and i386, but ARM64 would be great. Can't we build an ARM64 binary just by installing a cross toolchain to an Alpine/Ubuntu docker environment and build inside it?

felixguendling:
> Can't we build an ARM64 binary just by installing a cross toolchain to an Alpine/Ubuntu docker environment and build inside it?

I think this could work. It would probably be slow due to being emulated via Qemu, but it could work.

But as you say, if you want to support many toolchains (including Windows), CMake is probably a solid option.

rui314:
I'd suggest using a cross toolchain, so it should be as fast as a native development.

felixguendling:
Seems like this is possible but not documented very good on alpine:
https://stackoverflow.com/questions/54558200/cross-compiling-alpine-linux-packages

But maybe this should be discussed in #193. I think if you're sure you want Alpine as a build env for the musl binaries, you can already merge #193 and make the cross-build another PR.

Tachi107:
I understand that this was made mainly to allow static linking, but what if I want to build mold with CMake and link dynamically to the dependencies, or link statically only when they are not found on the system? It seems that `pkg` can't currently handle this. Also, I'm not a big fan of using custom CMake scripts to download deps, I'd rather use [`FetchContent`](https://cmake.org/cmake/help/latest/module/FetchContent.html) or other first party modules. 

> I'd suggest using a cross toolchain, so it should be as fast as a native development.

Since you're not already using CMake, you may want to look into Meson, as it has pretty good [cross compilation](https://mesonbuild.com/Cross-compilation.html) support. It can also handle the "use system deps if available, link statically otherwise" situation that I mentioned above without requiring third party modules or complex logic in the build script; it's really nice, and while I often use CMake when working with other people's projects I always choose Meson for personal ones as it makes my life (and downstream's) easier.

felixguendling:
> I'd rather use FetchContent or other first party modules

From my perspective, CMake is a good build tool and a horrible dependency manager. In the sense of Unix "Make each program do one thing well" I would rather use CMake + a proper dependency manager.


> or link statically only when they are not found on the system?

To get predictable results and reproducible builds across Linux distributions/package managers, I would strongly advise to bring all dependencies required to build using a package manager. I don't see any reason to depend on system libraries except maybe for security reasons (e.g. fix heartbleed once in the dyn. library instead of fixing N libraries separately; however with "one Docker container per application, the benefits of this are vanishing).

----------

However, I made this primarily for internal use and just wanted to offer to include it upstream in case someone is interested. It seems like upstream is more directed to other solutions. This is perfectly fine for me. Closing.

PR #186, opened 2021-12-22
[ELF] Add missing calls to `va_end()`

sicherha:
Reported by cppcheck. Omitting `va_end()` causes undefined behaviour:
https://en.cppreference.com/w/c/variadic/va_end


rui314:
Thanks!

PR #192, opened 2021-12-23
Do not use custom `LDFLAGS` for linking `mold-wrapper.so`

### LINKED ISSUE #191: `build-static.sh` produces unusable `mold-wrapper.so`
The build script sets its own `LDFLAGS` for use inside the container. Since these `LDFLAGS` contain `-static`, the linker tries to produce `mold-wrapper.so` as a **statically** linked **dynamic** library - which is sort of an oxymoron. As a result, it is not possible to `LD_PRELOAD` the library.

How to reproduce:
```sh
make clean
./build-static.sh
make test
```

Relevant part of the output:
```
Testing mold-wrapper2 ...       not a dynamic executable
ERROR: ld.so: object '/home/christoph/Projects/mold/out/test/elf/mold-wrapper2/bin/mold-wrapper.so' from LD_PRELOAD cannot be preloaded (cannot dynamically load executable): ignored.
OK
Testing mold-wrapper ...        not a dynamic executable
ERROR: ld.so: object '/home/christoph/Projects/mold/test/elf/../../mold-wrapper.so' from LD_PRELOAD cannot be preloaded (cannot dynamically load executable): ignored.
ERROR: ld.so: object '/home/christoph/Projects/mold/test/elf/../../mold-wrapper.so' from LD_PRELOAD cannot be preloaded (cannot dynamically load executable): ignored.
/usr/bin/ld: cannot find execl: No such file or directory
make[1]: *** [Makefile.linux:6: elf/mold-wrapper.sh] Error 1
make: *** [Makefile:145: test] Error 2
```

Suggested solution: make sure that the custom `LDFLAGS` are applied only to `mold` itself, not to `mold-wrapper.so`.

sicherha:
Linking a dynamic library with `-static` would prevent it from being
loadable.

Fixes #191.

Signed-off-by: Christoph Erhardt <github@sicherha.de>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,7 +105,7 @@ endif
 all: mold mold-wrapper.so
 
 mold: $(OBJS) $(MIMALLOC_LIB) $(TBB_LIB) $(XXHASH_LIB)
-	$(CXX) $(CPPFLAGS) $(CXXFLAGS) $(LDFLAGS) $(OBJS) -o $@ $(LIBS)
+	$(CXX) $(CPPFLAGS) $(CXXFLAGS) $(LDFLAGS) $(EXTRA_LDFLAGS) $(OBJS) -o $@ $(LIBS)
 	ln -sf mold ld
 	ln -sf mold ld64.mold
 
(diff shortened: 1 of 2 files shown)


rui314:
Thanks!

Issue #187, opened 2021-12-22
Issues building mold statically on alpine


RichardoC:
If you build mold using Alpine as in https://github.com/RichardoC/mold/commit/97d31b8b74314cbace101afa23ed432c28d96083 it fails with 

```bash
# Ignored for brevity
In file included from /mold/third-party/tbb/src/tbb/concurrent_monitor_mutex.h:22,
                 from /mold/third-party/tbb/src/tbb/concurrent_monitor.h:23,
                 from /mold/third-party/tbb/src/tbb/address_waiter.cpp:19:
/mold/third-party/tbb/src/tbb/semaphore.h:61:10: fatal error: linux/futex.h: No such file or directory
   61 | #include <linux/futex.h>
      |          ^~~~~~~~~~~~~~~
```

This is either because of a missing dependancy in my dockerfile (included in that commit,) or an issue in that third party library. 

Seems similar to https://github.com/bitcoin/bitcoin/issues/16925 which might mean that an updated version of that tbb package might have the fix.

rui314:
Does installing `linux-headers` package solve your issue?

RichardoC:
Thanks, got a lot further. Now fails with

`ld.lld: error: out/hyperloglog.o: section header string table index 1 does not exist`

https://github.com/RichardoC/mold/commit/88d769bade79746a034a76d921546f4b40267063

I can provide the full logs if needed, but hope this is enough to help.


When I run it verbosely

```bash
$ cd /mold
/mold 
$ clang++ -DMOLD_VERSION=\"1.0.0\" -DLIBDIR="\"/usr/local/lib\"" -DGIT_HASH=\"97d31b8b74314cbace101afa23ed432c28d96083\" -Ithird-party/mimalloc/include -I/lib -Ithird-party/tbb/i
nclude -Ithird-party/xxhash -O2 -pthread -fPIE -fno-unwind-tables -fno-asynchronous-unwind-tables -std=c++20 -fno-exceptions -fuse-ld=lld -static -Wl,-u,pthread_rwlock_rdlock -Wl,-u,pt
hread_rwlock_unlock -Wl,-u,pthread_rwlock_wrlock out/compress.o out/demangle.o out/filepath.o out/hyperloglog.o out/main.o out/perf.o out/tar.o out/elf/arch-arm64.o out/elf/arch-i386.o
[...]

PR #193, opened 2021-12-23
Build for either ubuntu or alpine with Docker

RichardoC:
Also added some convenience for those using [nerdctl](https://github.com/containerd/nerdctl/releases) due to the docker desktop license changes.


RichardoC:
Should mean noone else needs to open the equivalent of https://github.com/rui314/mold/issues/187

rui314:
This may be a dumb question, but what is the point of allowing Alpine to create a mold executable? Is that better than Ubuntu in some way?

Please add a `signed-off-by` line to your commit as explained in https://github.com/rui314/mold/blob/main/CONTRIBUTING.md.

RichardoC:
Ubuntu uses glibc and Alpine uses MUSL. It is possible to compile for MUSL on Ubuntu but it's easier to just build it on Alpine.
I've fixed my commits now, sorry about that! For anyone who does the same thing, here's a simple script to fix them
```bash
export SIGNOFF="signed-off-by: $(git config --get user.name) <$(git config --get user.email)>"
git filter-branch -f --msg-filter     "git interpret-trailers --trailer \"$SIGNOFF\""      HEAD~10..HEAD
git push --force
```

test/elf/bug179.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+export LANG=
+set -e
+cd $(dirname $0)
+mold=`pwd`/../../mold
```
RichardoC: If you're always running this in a git repo, you could use 
```bash
export ROOT=$(git rev-parse --show-toplevel | tr -d '\n')
mold="$ROOT/mold"
```

rui314:
So I guess your intention is to get rid of glibc from the statically-linked mold executable and use musl instead. Is this correct? If so, how about also removing the dependency to GNU's libstdc++ and switch to LLVM's libc++? If it's doable, and if it doesn't negatively impact mold's performance, I'd do that.

After that, we can simply replace the current `build-static.sh` with your Alpine's one.

rui314:
It looks like your pull request contains my commits. Please rebase so that this PR consists only of your changes.

sicherha (commented):
Could you explain what these additions to `CPPFLAGS` and `LIBS` are supposed to achieve? The comment says something about stuff being installed to `/lib` instead of `/usr/local/lib`, but that doesn't seem to have any relation to what is actually being done:
* Add `/lib` to the preprocessor's include path
* Add the root directory to the linker's library path

Unless Alpine has an incredibly weird filesystem layout that I'm unaware of, I don't think these changes are actually needed.

RichardoC:
Abandoning in favour of https://github.com/rui314/mold/pull/166
Thanks for the reviews everyone, that's a much better solution than this PR

PR #194, opened 2021-12-23
Container improvements

sicherha:
This patch series brings the following enhancements for `build-static.sh`:
* Add compatibility for Podman + SELinux (a common combination on Fedora).
* Allow running the testsuite inside the container. Note: requires including a few extra packages in the container image.
* Allow passing extra arguments to `make` inside the container.

Let me know if you find these additions useful, and feel free to reject some or all of them if you don't. 🙂


rui314:
Thanks!

PR #200, opened 2021-12-26
[ELF] Fix '?' for globbing + version scripts

jmglogow:
With all the buzz about mold, I tried to link LibreOffice with it. After implementing the setup with gcc -B, this resulted in missing symbols. Turned out that '?' for globbing was not handled correctly.


rui314:
Great, thanks!

PR #206, opened 2021-12-29
Using pathmax

SvenJo:
reduce magic numbers (4096 & 8192) for the max path length with PATH_MAX.
should make the code also a tiny bit more portable

(I forgot to use -s in the first commit, 2 commit are the same changes, but with "-s")

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,16 +105,13 @@ If you can specify an additional command line option to your compiler
 driver by modifying build system's config files, add one of the
 following flags to use `mold` instead of `/usr/bin/ld`:
 
-- Clang before 12: pass `-fuse-ld=<absolute-path-to-mold-executable>`;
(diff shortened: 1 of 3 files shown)


rui314:
Thanks! It makes sense.

I think it is better to directly use `PATH_MAX` instead of defining a constexpr variable `pathmax` as an alias to that constant value. Also, can you squash the changes into one commit?

SvenJo:
There is an additional change: using a std::string instead of a char buffer (buf)I will keep this change (only to double check)RegardsSvenAm 29.12.2021 14:08 schrieb Rui Ueyama ***@***.***>:
Thanks! It makes sense.
I think it is better to directly use PATH_MAX instead of defining a constexpr variable pathmax as an alias to that constant value. Also, can you squash the changes into one commit?

PR #207, opened 2021-12-29
using PATH_MAX

SvenJo:
Signed-off-by: Sven Johannsen <s.johannsen@bretgeld-engineering.de>

### DIFF
--- a/filepath.cc
+++ b/filepath.cc
@@ -5,15 +5,15 @@
 namespace mold {
 
 std::string get_current_dir() {
-  std::string buf(8192, '\0');
+  std::string buf(PATH_MAX, '\0');
   getcwd(buf.data(), buf.size());
   buf.resize(buf.find_first_of('\0'));
   return buf;
 }
 
 std::string get_realpath(std::string_view path) {
-  char buf[8192];
-  if (!realpath(std::string(path).c_str(), buf))
+  std::string buf(PATH_MAX, '\0');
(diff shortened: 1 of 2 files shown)


filepath.cc
```diff
@@ -5,15 +5,15 @@
...
-  std::string buf(8192, '\0');
+  std::string buf(PATH_MAX, '\0');
   getcwd(buf.data(), buf.size());
   buf.resize(buf.find_first_of('\0'));
   return buf;
 }
 
 std::string get_realpath(std::string_view path) {
-  char buf[8192];
-  if (!realpath(std::string(path).c_str(), buf))
```
rui314, line 16: It looks like the new code always returns a string of PATH_MAX length filled with NUL bytes. I think just replacing 8192 with PATH_MAX is better.

sicherha:
Caution! `PATH_MAX` is a bit of a false friend; [relying on it is trickier than one might expect](https://eklitzke.org/path-max-is-tricky) as an actual path might be more than `PATH_MAX` characters long. `getcwd()` and `realpath()` in particular require some special handling. Let me see if I can whip up an alternative patch.

Alcaro:
In POSIX 2008, realpath(path, NULL) calls malloc, so that one's easy.

But yes, getcwd is nontrivial. glibc getcwd has the same NULL-malloc feature, but that's nonstandard.

sicherha:
I guess we can close this pull request, right?

PR #209, opened 2021-12-29
Add lots of quotes to shell scripts

sicherha:
This makes it possible to build and test mold in a path that contains
whitespace characters - with the notable exception of the tests where
`LD_PRELOAD` is used. That's because `LD_PRELOAD` unconditionally treats
any whitespace as separator, regardless of quoting.

The following ShellCheck warnings are eliminated by this commit:
* SC2046: Quote this to prevent word splitting.
* SC2086: Double quote to prevent globbing and word splitting.

Signed-off-by: Christoph Erhardt <github@sicherha.de>

### DIFF
--- a/build-static.sh
+++ b/build-static.sh
@@ -41,4 +41,4 @@ fi
 
 docker run -it --rm "${docker_args[@]}" \
   mold-build-ubuntu20 \
-  make -C /mold -j$(nproc) EXTRA_LDFLAGS="$EXTRA_LDFLAGS" "$@"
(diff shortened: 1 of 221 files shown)


rui314:
Thanks! I honestly don't recommend working in a directory that contains whitespace characters in its name, as it causes troubles like this. But still it is better to support such environment.

sicherha:
Neither do I, but I believe it's worthwhile being cautious. 🙂 After all, a lack of quoting might also lead to very weird behaviour for paths containing `*`, `?` or other wildcard characters.

PR #210, opened 2021-12-29
Fix handling of path buffer sizes

sicherha:
This is a draft pull request that tries to address the concerns raised in #207. I have retained the part of @SvenJo's commit that remains unchanged, and added my suggested overhaul of ` get_current_dir()` and `get_realpath()` as additional commits on top.

@rui314, @SvenJo, please let me know what you think.

### DIFF
--- a/elf/subprocess.cc
+++ b/elf/subprocess.cc
@@ -1,5 +1,6 @@
 #include "mold.h"
 
+#include <filesystem>
 #include <signal.h>
 #include <sys/socket.h>
 #include <sys/stat.h>
@@ -249,15 +250,8 @@ void daemonize(Context<E> &ctx, std::function<void()> *wait_for_client,
   };
 }
 
-template <typename E>
-static std::string get_self_path(Context<E> &ctx) {
-  char buf[4096];
-  size_t n = readlink("/proc/self/exe", buf, sizeof(buf));
-  if (n == -1)
-    Fatal(ctx) << "readlink(\"/proc/self/exe\") failed: " << errno_string();
-  if (n == sizeof(buf))
-    Fatal(ctx) << "readlink: path too long";
-  return {buf, n};
+static std::string get_self_path() {
+  return std::filesystem::read_symlink("/proc/self/exe");
 }
 
 static bool is_regular_file(const std::string &path) {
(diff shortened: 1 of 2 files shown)


rui314:
It's not your fault, but the APIs of realpath, getcwd, readlink and the like are a bit messy, and that's why I just give a large buffer to the functions so that we can forget about the details (unless a user is trying to use mold with an insanely long pathname). Can we just a buffer with 8192 bytes (which is I believe more than enough, but if not, we can even pass a 16 KiB buffer) so that these functions won't fail?

sicherha:
Wait a second - why don't we simply use the [C++17 filesystem API](https://en.cppreference.com/w/cpp/filesystem)? 💡

rui314:
Indeed this looks better. Thanks!

SvenJo:
I also like the idea of using std::filesystem and concentrating to change  get_current_dir and get_realpath. In my own tests I had trouble to implement the rest of the functions inside  filepath.cc using std::filesystem equivalents. (`path_is_dir` is maybe an easy to convert)

Regards,
Sven

rui314:
I landed a patch to use C++17 filesystem API instead of POSIX API inhttps://github.com/rui314/mold/commit/f6e91df440b387c69a33df71d2a44bac58ad14d0, but had to revert it because it broke many builds. I'll find the cause of the issue, fix it and re-land the patch soon.

sicherha:
Here's the relevant compiler error:
```
In file included from demangle.cc:1:
./mold.h:437:40: error: no member named 'generic_format' in 'std::filesystem::path'
  return {path, std::filesystem::path::generic_format};
                ~~~~~~~~~~~~~~~~~~~~~~~^
1 error generated.
```
You may need to replace `std::filesystem::path::generic_format` with `std::filesystem::path::format::generic_format`.

sicherha:
After a bit of code archaeology, here's the explanation: the macOS build image comes with a version of libcxx that was released prior to commit https://github.com/llvm/llvm-project/commit/26005c788b06755ccafe1205de181bbd8b2ed664, which converts the mistaken (overeager) declaration of `enum class format { ...} ;` to the standard-conforming `enum format { ... };`.

rui314:
The problem I was talking about is not about the compilation error on macOS. It started mislinking lots of programs for some reason. I'll investigate, fix it and land it again hopefully soon.

PR #212, opened 2021-12-30
Avoid infinite recursion in do_main<X86_64>

tklauser:
Running mold on x86_64 macOS 11 currently leads to a segfault due to
inifinite recursion in do_main<X86_64>. Protect against this case by
only calling into do_main<X86_64> once if cputype doesn't match.

### DIFF
--- a/macho/main.cc
+++ b/macho/main.cc
@@ -369,8 +369,10 @@ static int do_main(int argc, char **argv) {
   std::vector<std::string> file_args;
   parse_nonpositional_args(ctx, file_args);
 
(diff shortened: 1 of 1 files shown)


rui314:
Can you read https://github.com/rui314/mold/blob/main/CONTRIBUTING.md and add a `signed-off-by` line to your commit? Other than that, LGTM

tklauser:
> Can you read https://github.com/rui314/mold/blob/main/CONTRIBUTING.md and add a `signed-off-by` line to your commit? Other than that, LGTM

Done, thanks.

rui314:
Thanks!

PR #213, opened 2021-12-30
Fix skip check in i386-mergable-strings.sh

tklauser:
Check for the ability to build 32-bit binaries like in the other i386
test cases.


rui314:
Thanks!

PR #214, opened 2021-12-30
[Mach-O] Fatal if no input files are given

tklauser:
Follow the same behavior in the ELF version of read_input_files and
fatal in case no input files are given. This avoids a segfault in
successive functions when running `./ld64.mold` without any arguments.


rui314:
Thanks!

PR #217, opened 2021-12-31
Fix building on Android again

joshcangit:
The commit 946e0c3 undid the fix from pull request #204.
There's now segmentation fault when using mold.
Since macOS's uname doesn't support the `-o` option, the better option is `-a` and not `-r`.
I'm using Termux and this change works.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,7 @@ endif
 USE_MIMALLOC = 1
 ifeq ($(OS), Darwin)
   USE_MIMALLOC = 0
-else ifneq (, $(findstring android, $(shell uname -r)))
+else ifneq (, $(findstring Android, $(shell uname -a)))
   USE_MIMALLOC = 0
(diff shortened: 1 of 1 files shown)


rui314:
Apologies for breaking your Android build.

One question: isn't `-r` work for you? `uname -a` should certainly work for most cases, but if someone who names their machine `Android`, `findstring` matches falsely.

rui314:
I submitted d743acb6cddc0403c8f3c022287938513be0693f instead to restore `uname -o`. Unlike the original code, that command is run only when it is running on Linux, so it's compatible with macOS.

joshcangit:
Thanks, that was better than this pull request.

leleliu008:
may be `ifneq ($(findstring -android,$(shell $(CC) -dumpmachine 2>/dev/null)),)` is better.

rui314:
@leleliu008 Thanks! I'll use that technique.

PR #219, opened 2021-12-31
Compatibility for libc++ < 14, plus two additional fixes 

sicherha:
* Improve Xcode 13.1 workaround (caused by a libc++ bug)
* Fix `errno_string()` returning an empty string on GNU
* Terminate `argv` array with `nullptr`

### DIFF
--- a/main.cc
+++ b/main.cc
@@ -8,8 +8,15 @@ namespace mold {
 
 std::string_view errno_string() {
   static thread_local char buf[200];
-  strerror_r(errno, buf, sizeof(buf));
-  return buf;
+#if _GNU_SOURCE
+  // The GNU version of strerror_r() returns a char * and may completely ignore
+  // buf
+  return strerror_r(errno, buf, sizeof(buf));
+#else
+  // The POSIX.1-2001-compliant version of strerror_r() returns an int and
+  // writes the string into buf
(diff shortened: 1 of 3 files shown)


main.cc
```diff
@@ -8,7 +8,15 @@ namespace mold {
 
 std::string_view errno_string() {
   static thread_local char buf[200];
-  strerror_r(errno, buf, sizeof(buf));
+
+  // Special handling for the GNU version of strerror_r(): the function returns
+  // a char * and may completely ignore buf.
+  // The POSIX.1-2001-compliant version of strerror_r() returns an int and
+  // always writes to buf.
+  auto res = strerror_r(errno, buf, sizeof(buf));
```
rui314, line 16: We don't define `_GNU_SOURCE`, so is this really needed?

Alcaro: GNU_SOURCE [is on by default](https://godbolt.org/z/P1zfexTTj) in C++ mode. [It's a known issue.](https://gcc.gnu.org/onlinedocs/libstdc++/faq.html#faq.predefined)

sicherha: Changed the implementation to preprocessor fiddling, because the compiler does perform type checking even in the body of an `if constexpr` that evaluates to `false`. I thought I was being clever...

rui314: I actually like your new code which uses the C preprocessor directive!

rui314:
Thanks!

PR #221, opened 2021-12-31
Remove duplicate code

sicherha:
The exact same four lines of code exist a few lines below this hunk.


rui314:
Thanks. Can you add a `signed-off-by` line?

rui314:
Thanks!

PR #230, opened 2022-01-02
Rewrite manual page in `mdoc(7)`

Alhadis:
This PR refactors `docs/mold.1` to use [`mdoc(7)`](https://man.openbsd.org/mdoc.7) instead of [`man(7)`](https://man.openbsd.org/man.7). Mdoc is a modern alternative to the historic `man` macro-set, which abstracts presentational details from higher-level markup.

[Here's a before-and-after](https://codepen.io/alhadis/full/KKXoBOV) comparison to show you how the page will look when formatted for the terminal.

### DIFF
--- a/docs/mold.1
+++ b/docs/mold.1
@@ -1,734 +1,949 @@
-.TH MOLD 1
-.SH NAME
-mold \- A Modern Linker
-
-.SH SYNOPSIS
-mold [\fBoptions\fR] \fIobjfile\fR \fB...\fR
-
-.SH DESCRIPTION
-\fBmold\fR is a multi-threaded, high-performance linker that is
-several times faster than the industry-standard ones, namely, GNU ld,
-GNU gold or LLVM lld. It is developed as a drop-in replacement for
-these linkers and command-line compatible with them with a few
-exceptions.
-.PP
-Note that even though \fBmold\fR can build many userland programs,
-including large ones such as Chrome and Firefox, it is still very
-experimental. Do not use \fBmold\fR for production unless you know
-what you are doing. \fBmold\fR supports Linux/x86-64 and Linux/i386.
-
-.SS "How to use mold"
-On Unix, \fBcc\fR (or \fBgcc\fR or \fBclang\fR, depending on what
-compiler you use) is not just a C compiler but is actually a compiler
-driver for various kinds of input files. \fBcc\fR dispatches based
-on input file extensions. If you pass object files (\fB.o\fR files) to
-a compiler driver, the linker command, which is usually installed as
-\fB/usr/bin/ld\fR, is invoked as a backend.
-.PP
+.\"
+.\" This manpage is written in mdoc(7).
+.\"
+.\" * Language reference:
+.\"   https://man.openbsd.org/mdoc.7
+.\"
+.\" * Atom editor support:
+.\"   https://atom.io/packages/language-roff
+.\"
+.\" * Linting changes:
+.\"   mandoc -Wall -Tlint /path/to/this.file  # BSD
+.\"   groff -w all -z /path/to/this.file      # GNU/Linux, macOS
+.\"
+.\"
+.\" When making changes, please keep the following in mind:
+.\"
+.\" * In Roff, each new sentence should begin on a new line. This gives
+.\"   the Roff formatter better control over text-spacing, line-wrapping,
+.\"   and paragraph justification.
+.\"
+.\" * If a line exceeds the maximum length enforced by a project's \
+.\"   coding style, prefer line-continuation instead of hard-wrapping; \
+.\"   that is, end each incomplete (physical) line with a backslash, \
+.\"   like in this paragraph.
+.\"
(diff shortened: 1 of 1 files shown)


rui314:
I'm pretty much noob when it comes to man pages. I just wrote this man page by copy-pasting macros from other manuals. So, I"m sure this is a dumb question, but what is the point of writing a man page with the mandoc style instead of the current style?

Semantic markup may be a good thing, but does this make a visible difference?

Is this compatible on all Unix-like platforms, i.e. all Linux distros, BSDs and macOS?

What about other projects? Do other project write their man pages in mandoc rather than the GNU-style these days?

Alhadis:
> Is this compatible on all Unix-like platforms, i.e. all Linux distros, BSDs and macOS?

Yes, and it has been for the past 3 decades. I know I said `mdoc` was "modern", but in the context of a 50-year old typesetting system, practically anything written post-1990 is "modern" in comparison (the original `man(7)` macros were written in the mid-70s).

> So, I"m sure this is a dumb question, but what is the point of writing a man page with the mandoc style instead of the current style?

BSD's [`man(1)`](https://man.openbsd.org/man.1) and [`apropos(1)`](https://man.openbsd.org/apropos.1#EXAMPLES) implementations use semantically-tagged data to perform advanced lookup and indexing queries based on page content; e.g., searching for man pages that contain pathnames ending in `.so` that also reference an environment variable named `LD_PRELOAD`:

~~~console
$ apropos 'Pa~\.so$' -a Ev=LD_PRELOAD
~~~

Unlike a conventional search, `.so` is guaranteed only to match strings marked up with [`.Pa`](https://man.openbsd.org/mdoc.7#Pa). There's also the added benefits of improved hyperlinking: I can link directly to the definition of [`.Ev`](https://man.openbsd.org/mdoc.7#Ev) inside `mdoc(7)`, for example.

None of these features are possible with purely presentational markup, which only concerns itself with the _appearance_ of text (bold, underlined, indent, etc).

> Semantic markup may be a good thing, but does this make a visible difference?

In the terminal, the improvements to formatting are marginal, but they spring to life when rendered in any other format (HTML, PDF, PostScript, etc). Compare:

**Unstyled HTML output:**
* [Using `mdoc(7)`](https://codepen.io/alhadis/full/LYzdXej)
* [Using `man(7)`](https://codepen.io/alhadis/full/jOGzQaj)

> What about other projects? Do other project write their man pages in mandoc rather than the GNU-style these days?

BSD-derived systems (OpenBSD, FreeBSD, NetBSD, DragonflyBSD, etc) all use `mdoc(7)` extensively, whereas GNU/Linux gravitate toward the older `man(7)` package. The reason for this is primarily historical: at the time `mdoc` was first written, the `man` macros were already firmly-entrenched in the Unix ecosystem. It was around this time that Linux appeared on the scene and really took off… authors continued to write man pages in the format they were most used to.

Today, the majority of manual pages that use `man(7)` are either
1. generated from sloppy conversion software (DocBook is an infamous example),
2. copy+pasted piecemeal from other man pages (as you did),
3. written by maintainers of [official GNU projects](https://www.gnu.org/philosophy/categories.html#GNUsoftware)[^1]
4. written by casual man page authors who learned through osmosis and/or points 1. or 2. above; these authors often feel little pressure or incentive to learn more about Roff than is necessary to maintain a basic manual page for a project.

[Node.js](https://github.com/nodejs/node) used to be an example of case \#&zwnj;4 until I modernised it for them (see nodejs/node#18559).

[^1]: "For some reason, the FSF *insists* on using the legacy `man(7)` macros. I have no idea why"

rui314:
It looks like only 66 man pages among 3917 in my /usr/share/man/man1/ are written in the mdoc format, so it's definitely a minority. But I'm fine with mdoc because ld.lld's man page is also written in mdoc. I wrote part of it, but I wasn't aware of what that man page was written in.

I'll merge this commit.

One question: why is line continuation preferred over hard line break?

Alhadis:
> It looks like only 66 man pages among 3917 in my /usr/share/man/man1/ are written in the mdoc format, so it's definitely a minority.

Perhaps on your system, which I'm assuming is some flavour of GNU/Linux? On BSD-derived systems, `mdoc` is much more prevalent.

Moreover, you only scanned the contents of your `/usr/share/man/man1/` directory. 😉 Of the 9 different `/usr/share/man/man?` directories, `man1` — the section dedicated to userland commands — is the one most likely to contain a mixture of GNU and BSD-licensed programs (meaning the ratio between `man`/`mdoc` can vary widely).

> One question: why is line continuation preferred over hard line break?

When a sentence ends in an appropriate punctuation character like `. , ) ? !`, the formatter can optimise the spacing of surrounding text, improving the readability of justified or filled paragraphs. A hard line-break is interpreted as the end of a sentence which is missing the terminating full-stop (or comma, question mark, etc).

This tends to be much more noticeable in PDF and PostScript output, but it's also relevant in terminal output as well (particularly when densely-packaged blocks of text are involved).

Issue #222, opened 2022-01-01
Build error on RISC-V 64bits system


OctopusET:
## Error
When I was trying to build with `make -j$(nproc)` on riscv64(QEMU) Debian unstable, I get this error.

>  error: '__TBB_machine_fetchadd4' was not declared in this scope
>   348 |     return __TBB_machine_fetchadd4(ptr, 1) + 1L;

I found similar error https://bugs.funtoo.org/browse/FL-8615?attachmentSortBy=fileName

## Fix
- And I fixed with this [patch](https://code.funtoo.org/bitbucket/users/calrama/repos/kit-fixups/commits/f2b857a6d5c0cffc68afba3e8a273233708d59eb)

- Or using this [patch](https://github.com/oneapi-src/oneTBB/issues/186#issuecomment-871649431) would work too.
```diff
-    return __TBB_machine_fetchadd4(ptr, 1) + 1L;
+    return __atomic_fetch_add(ptr, 1L, __ATOMIC_SEQ_CST) + 1L;
```

And I got another error after patch.
> undefined reference to `__atomic_exchange_1'

I fixed it with this.
```diff
-LIBS = -pthread -lz -ldl -lm
+LIBS = -pthread -lz -ldl -lm -latomic
```
\
\
Although those patches fix build error, but upstream onetbb does not require any patch when I was building on same system, it seems like they [removed](https://github.com/oneapi-src/oneTBB/blob/d1667d514df697f05d771602b268e92560c434c4/src/tbbmalloc/frontend.cpp#L802) Arch specific code. I think updating onetbb to upstream will be better option.

#### References:
- https://github.com/advancedtelematic/aktualizr/issues/1427
- https://stackoverflow.com/questions/35339514/undefined-reference-in-static-library-to-c11-atomic

OctopusET:
Please let me know what would you think the best solution, I can make PR. Thanks

rui314:
Thank you for testing this on RISC-V!

It looks like upgrading onetbb is the best option if it solves the problem for you and doesn't have any regression. Let me take a look.

OctopusET:
Thank you for response!

Please let me know if there is anything I can do.

rui314:
I updated our oneTBB subtree to v2021.5.0, the latest stable version of the library. Can you try again to see if it builds on RISC-V?

OctopusET:
@rui314 Thank you for work.

[...]

PR #232, opened 2022-01-02
Fix build error on riscv64 (#222)

OctopusET:
Patch from this PR: https://github.com/oneapi-src/oneTBB/pull/550
Thank you, @felixonmars

Build works with this patch, but does not support elf64lriscv. So, test fails.

Fix: #222

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@ endif
 STRIP ?= strip
 
 OS ?= $(shell uname -s)
+ARCH ?= $(shell uname -m)
 
 # Used for both C and C++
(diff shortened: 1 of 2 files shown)


OctopusET:
Tested with `make test` on x86_64 system(with docker), and passed every test.

OctopusET:
Changed Makefile to use `-latomic` flag only for RV64

rui314:
Thanks! Can you read https://github.com/rui314/mold/blob/main/CONTRIBUTING.md and add a `signed-off-by` line to your commit message? Other than that, LGTM.

OctopusET:
@rui314 Oh my bad, I added `signed-off-by`

Thanks!

rui314:
Thanks!

PR #237, opened 2022-01-03
add GitHubActions to build for android

leleliu008:


rui314:
Could you read https://github.com/rui314/mold/blob/main/CONTRIBUTING.md and add a `signed-off-by` line to your commit message?

rui314:
This looks a bit too complicated -- it builds various targets on various hosts and downloads some tar file from other repo using `curl`. I think that just build mold by running `make` on a single host (Android on ARM64) would suffice. Do you mind if I ask you to strip this patch down to minimum? I'm not familiar with GitHub actions, but if you don't have time to do that, I'll do that later.

leleliu008:
> and downloads some tar file from other repo using `curl`.

I use the external prebuild runtime dependencies. I don't use the included `/third-party/tbb` and `third-party/xxhash`, beacause the build script should be changed :
https://github.com/rui314/mold/blob/main/Makefile#L165
https://github.com/rui314/mold/blob/main/Makefile#L170

this project need  external prebuild openssl.

PR #241, opened 2022-01-04
Use pkg-config to know where libcrypto resides

sina-ht:
Let Makefile know how to link libcrypto in a directory other than the default
directories by using pkg-config. If you don't have pkg-config, then $(shell)
returns empty.

Signed-off-by: Hiroshi Takekawa <sian.ht@gmail.com>


rui314:
Thanks!

PR #249, opened 2022-01-05
Allow overriding the default target architecture

sicherha:
I'm packaging mold for Fedora.

I tripped over the fact that mold currently hardcodes the default ELF emulation to `EM_X86_64` (and the default Mach-O architecture to `CPU_TYPE_ARM64`). As a consequence, an aarch64 build of mold would default to producing x86_64 binaries unless overridden by `-m`.

This patch allows distributions to select the desired defaults when building mold. It is controlled through the make options `DEFAULT_ELF_EMULATION` and `DEFAULT_MACHO_ARCH`.
In the process, I've also added a patch to fix two unit tests that fail on aarch64.

I have seen that lld is really clever: it automatically selects the target architecture based on the input files it receives. Is such a feature planned for mold as well?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,14 @@ ifeq ($(LTO), 1)
   LDFLAGS  += -flto
 endif
 
+# Allow overriding the default ELF emulation and the default Mach-O architecture
+ifdef DEFAULT_ELF_EMULATION
+  CPPFLAGS += -DDEFAULT_ELF_EMULATION=$(DEFAULT_ELF_EMULATION)
+endif
+ifdef DEFAULT_MACHO_ARCH
+  CPPFLAGS += -DDEFAULT_MACHO_ARCH=$(DEFAULT_MACHO_ARCH)
+endif
+
 # By default, we want to use mimalloc as a memory allocator. mimalloc
 # is disabled when ASAN or TSAN is on, as they are not compatible.
 # It's also disabled on macOS and Android because it didn't work on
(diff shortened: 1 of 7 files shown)


rui314:
I thought in practice `-m` option is always given. Isn't that the case?

If not, we need to do the same as lld does. mold should not have any host-dependent configs for build reproducibility. I.e. if you give the exact same command line options and input files to mold, mold should behave exactly the same on any host machine.

sicherha:
If the linker is invoked by the compiler driver, then yes - but there may be valid use cases for calling the linker directly. At the very least, mold's own test suite is one such scenario. 😉

sicherha:
The change I propose here would interact with the packaging in the following way: https://github.com/sicherha/fedora-mold/blob/master/mold.spec#L29
```spec
# Force the default ELF emulation / Mach-O arch to the package's own arch.
# Otherwise, mold would default to EM_X86_64 (ELF) and CPU_TYPE_ARM64 (Mach-O).
%ifarch x86_64
  %define arch_flags DEFAULT_ELF_EMULATION=EM_X86_64 DEFAULT_MACHO_ARCH=CPU_TYPE_X86_64
%endif
%ifarch %{ix86}
  %define arch_flags DEFAULT_ELF_EMULATION=EM_386 DEFAULT_MACHO_ARCH=CPU_TYPE_X86_64
%endif
%ifarch aarch64
  %define arch_flags DEFAULT_ELF_EMULATION=EM_AARCH64 DEFAULT_MACHO_ARCH=CPU_TYPE_ARM64
%endif
```

rui314:
In order to not depend on host-specific configuration, I submitted be49d673a562e4d813b570b70d3eebfa2ea9481d instead.

sicherha:
Nice, thanks a lot!

PR #252, opened 2022-01-05
Fix test failures on aarch64

sicherha:
Follow-up to #249. I can confirm that be49d67 works as expected on aarch64, but there are still two test cases that need fixing.


rui314:
Thanks!

PR #263, opened 2022-01-07
buildsys: Improve readability of the Makefile

leleliu008:


rui314:
It's a matter of personal preference, but I'm happy with the current indentation style. I don't think we need to vertically align variable names.

As to `OS`, Android is Linux, so I'm not in favor of overwriting `OS` with `Android`.

gruenich:
So, can we close this pull request?

PR #278, opened 2022-01-19
Improve compatibility with NetBSD

depressed-pho:
NetBSD still uses BFD ld. It is freaking slow and eats up all the memory. So I was looking for a good alternative and found mold. Mold looked great. **It looked really great.** Except it didn't work on NetBSD so here are my patches to make it compatible with the OS.

### DIFF
--- a/elf/input-files.cc
+++ b/elf/input-files.cc
@@ -346,9 +346,8 @@ void ObjectFile<E>::read_ehframe(Context<E> &ctx, InputSection<E> &isec) {
   for (std::string_view data = contents; !data.empty();) {
     i64 size = *(u32 *)data.data();
     if (size == 0) {
-      if (data.size() != 4)
-        Fatal(ctx) << isec << ": garbage at end of section";
-      break;
+      data = data.substr(4);
+      continue;
     }
 
     i64 begin_offset = data.data() - contents.data();
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ CFLAGS += $(COMMON_FLAGS) $(EXTRA_CFLAGS)
 CXXFLAGS ?= -O2
 CXXFLAGS += $(COMMON_FLAGS) $(EXTRA_CXXFLAGS) -std=c++20 -fno-exceptions
 CXXFLAGS += -DMOLD_VERSION=\"1.0.1\" -DLIBDIR="\"$(LIBDIR)\""
-LIBS = -pthread -lz -ldl -lm
+LIBS = -pthread -lz $(DL_LIBS) -lm
 
 SRCS=$(wildcard *.cc elf/*.cc macho/*.cc)
 HEADERS=$(wildcard *.h elf/*.h macho/*.h)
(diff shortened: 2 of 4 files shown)


elf/input-files.cc
```diff
@@ -346,9 +346,8 @@ void ObjectFile<E>::read_ehframe(Context<E> &ctx, InputSection<E> &isec) {
   for (std::string_view data = contents; !data.empty();) {
     i64 size = *(u32 *)data.data();
     if (size == 0) {
-      if (data.size() != 4)
-        Fatal(ctx) << isec << ": garbage at end of section";
-      break;
+      data = data.substr(4);
+      continue;
```
rui314, line 350: What object file did you see this error for? I'm worried about this change a bit because it could hide bugs.

rui314 (commented):
Thank you for doing this! Generally looking good.

elf/input-files.cc
rui314: Oh, sorry, I didn't notice that the reason was explained in the commit message.

depressed-pho: This one has an `.eh_frame` which has 8 octets of 0x00:
https://github.com/NetBSD/src/blob/trunk/lib/csu/arch/x86_64/crtend.S#L46

If I've correctly interpreted the specification, this should represent a pair of an empty CIE and an end-of-records marker for FDEs. See also http://dandylife.net/blog/archives/686

rui314: Here is an excerpt from https://refspecs.linuxbase.org/LSB_3.1.1/LSB-Core-generic/LSB-Core-generic/ehframechpt.html:

> Length: A 4 byte unsigned value indicating the length in bytes of the CIE structure, not including the Length field itself. If Length contains the value 0xffffffff, then the length is contained in the Extended Length field. If Length contains the value 0, then this CIE shall be considered a terminator and processing shall end.

This spec is a bit vague as to what "processing shall end", but at least in lld we ignore data after a terminator. mold should probably do the same (i.e. ignore a garbage after a terminator instead of reporting it as an error).

BenBE: Might issue a diagnostic (warning) though …

rui314: I don't think we want to show a diagnostic message for a trailing garbage after a terminator, because it would warn every time mold runs on NetBSD.

depressed-pho: > I don't think we want to show a diagnostic message for a trailing garbage after a terminator, because it would warn every time mold runs on NetBSD.

True. That's not desirable. A simple `break` might be better than my change then.

rui314: Yup. Do you want to make such change and push it again to github? Or I can just merge this commit as-is and make a change immediately after that.

depressed-pho: Shall I update my PR? Doing a `git push --force` would nuke this review IIRC though.

rui314: I don't know. I'm not very familiar with github review system. Let me just merge this patch to avoid possible mess.

rui314:
Do tests now pass with these changes? I want to double-check because I'd like to note about that in the release notes for the next release.

PR #279, opened 2022-01-19
Use ld.gold as fallback linker when possible

akawashiro:
Currently, mold fallback to ld.bfd when it finds GCC LTO
related-symbols. And ld.bfd is very slower than ld.gold. So, if we use
ld.gold instead of ld.bfd, we make it faster.

Because old binutils doesn't contains ld.gold, I kept the existing
fallback code for ld.bfd also.

### DIFF
--- a/elf/main.cc
+++ b/elf/main.cc
@@ -490,10 +490,16 @@ static int elf_main(int argc, char **argv) {
 
   // Do the same for GCC LTO.
   if (Symbol<E> *sym = get_symbol(ctx, "__gnu_lto_slim"); sym->file) {
+    std::string gnu_ld;
+    if (access("/bin/ld.gold", F_OK) == 0){
+      gnu_ld = "ld.gold";
+    } else{
+      gnu_ld = "ld.bfd";
+    }
     Warn(ctx) << *sym->file
-              << "GCC LTO is detected, so falling back to ld.bfd";
-    execvp("ld.bfd", argv);
-    Fatal(ctx) << "execvp failed: ld.bfd: " << errno_string();
+              << "GCC LTO is detected, so falling back to " << gnu_ld;
+    execvp(gnu_ld.c_str(), argv);
+    Fatal(ctx) << "execvp failed: " << gnu_ld << ": " << errno_string();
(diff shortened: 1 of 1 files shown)


rui314:
I wonder how common it is to use gold instead of bfd when you are doing an LTO build.

Also, does this actually make a difference? If you are doing LTO for a big program, LTO is very slow anyway.

If using gold for LTO is common, I'm fine with this change.

akawashiro:
I took a small benchmark to compare `ld.bfd` and `ld.gold` and noticed that their speed are not so different. I close this PR. Sorry.
```> cat t.sh 
#! /bin/bash

for i in $(seq 1 5); do
    rm lto-enabled.so
    time ninja
done

# ld.gold
> ./t.sh
[1/1] Linking CXX shared module lto-enabled.so

real    0m17.017s
user    0m14.006s
sys     0m2.984s
[1/1] Linking CXX shared module lto-enabled.so

real    0m16.161s
user    0m13.072s
sys     0m3.048s
[1/1] Linking CXX shared module lto-enabled.so

real    0m13.687s
user    0m10.990s
sys     0m2.697s
[1/1] Linking CXX shared module lto-enabled.so

real    0m14.523s
user    0m11.824s
sys     0m2.697s
[1/1] Linking CXX shared module lto-enabled.so

real    0m15.227s
user    0m11.920s
sys     0m3.300s
77298ms

# ld.bfd
> ./t.sh
[1/1] Linking CXX shared module lto-enabled.so

real    0m14.796s
user    0m11.937s
sys     0m2.856s
[1/1] Linking CXX shared module lto-enabled.so

real    0m14.803s
user    0m12.046s
sys     0m2.751s
[1/1] Linking CXX shared module lto-enabled.so

real    0m14.590s
user    0m11.811s
sys     0m2.776s
[1/1] Linking CXX shared module lto-enabled.so

real    0m15.003s
user    0m12.140s
sys     0m2.858s
[1/1] Linking CXX shared module lto-enabled.so

real    0m13.478s
user    0m10.829s
sys     0m2.647s
```

PR #291, opened 2022-01-22
[ELF]: Fix issues with relr relocations

BertalanD:
Thank you for creating this cool linker!

I found two small issues while [attempting to build](https://github.com/SerenityOS/serenity/pull/12049) the SerenityOS userland with mold + `relr` relocations. This PR fixes those.

Let me know if something doesn't look right or I should add tests.


rui314:
Thanks! I'll write a test for them.

PR #292, opened 2022-01-22
Explicitly include sys/select and sys/time

ADKaster:
The SerenityOS LibC doesn't leak the definitions of select(2) or struct
timeval into other POSIX headers. Manually include these headers in
elf/subprocess.cc where they are used for better compatibility.

This is the [equivalent patch](https://github.com/SerenityOS/serenity/blob/194cad0fc18fbb170c9a6a935e5832e970d9742b/Ports/mold/patches/0004-Add-POSIX-headers-for-timeval-and-select.patch) for what we currently have for version 1.0.1.

As a side note, I tested our port with latest master and the only issues were easily update-able by shuffling a couple more 2-3 line patches around!

### DIFF
--- a/elf/subprocess.cc
+++ b/elf/subprocess.cc
@@ -2,8 +2,10 @@
 
 #include <filesystem>
 #include <signal.h>
+#include <sys/select.h>
 #include <sys/socket.h>
 #include <sys/stat.h>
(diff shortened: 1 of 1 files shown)


rui314:
Great, thanks!

PR #299, opened 2022-01-23
Fix out-of-bounds assertion on aarch64 with `_GLIBCXX_ASSERTIONS` enabled

### LINKED ISSUE #298: `std::span` out-of-range failure when linking aarch64 program
I'm encountering this problem when running the mold 1.0.2 testsuite on Fedora aarch64. Fedora sets the preprocessor macro `_GLIBCXX_ASSERTIONS`, which enables assertions inside the libstdc++ code. This results in a failed bounds check becoming apparent.

# How to reproduce
```sh
make clean
./build-static.sh CXXFLAGS='-Og -g -D_GLIBCXX_ASSERTIONS'
docker run --rm -v "$PWD:$PWD:Z" -w "$PWD" -it mold-build-ubuntu20
# Inside the container:
apt update
apt -y install g++-aarch64-linux-gnu
cd test/elf
CC=aarch64-linux-gnu-gcc ./exclude-libs.sh
```

# Output
```
Testing exclude-libs ... collect2: fatal error: ld terminated with signal 6 [Aborted], core dumped
compilation terminated.
```

# Analysis
Stack trace:
```
/usr/bin/../lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/span:281: std::span::reference std::span<mold::elf::InputSection<mold::elf::ARM64> *, 18446744073709551615>::operator[](std::span::size_type) const [_Type = mold::elf::InputSection<mold::elf::ARM64> *, _Extent = 18446744073709551615]: Assertion '__idx < size()' failed.

Thread 2.1 "mold" received signal SIGABRT, Aborted.
[Switching to Thread 0x7ffff77690c0 (LWP 798697)]
__pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at pthread_kill.c:44
44            return INTERNAL_SYSCALL_ERROR_P (ret) ? INTERNAL_SYSCALL_ERRNO (ret) : 0;
(gdb) bt
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, 
    no_tid=no_tid@entry=0) at pthread_kill.c:44
#1  0x00007ffff77f98b3 in __pthread_kill_internal (signo=6, threadid=<optimized out>)
    at pthread_kill.c:78
#2  0x00007ffff77ac6a6 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007ffff77967d3 in __GI_abort () at abort.c:79
#4  0x0000000000410d9d in std::__replacement_assert (__file=<optimized out>, 
    __line=<optimized out>, __function=<optimized out>, __condition=<optimized out>)
    at /usr/bin/../lib/gcc/x86_64-linux-gnu/10/../../../../include/x86_64-linux-gnu/c++/10/bits/c++config.h:461
#5  0x000000000042972d in std::span<mold::elf::InputSection<mold::elf::ARM64>*, 18446744073709551615ul>::operator[] (this=0x7fffffffb3e8, __idx=3)
    at /usr/bin/../lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/span:281
#6  0x000000000041f4a2 in mold::elf::create_thunks (ctx=..., osec=...) at elf/arch-arm64.cc:577
#7  0x000000000041f033 in mold::elf::create_range_extension_thunks (ctx=...)
    at elf/arch-arm64.cc:738
#8  0x0000000000592d63 in mold::elf::elf_main<mold::elf::ARM64> (argc=<optimized out>, 
    argv=<optimized out>) at elf/main.cc:624
#9  0x000000000058f9b0 in mold::elf::elf_main<mold::elf::X86_64> (argc=6, argv=0x7fffffffdb78)
    at elf/main.cc:357
#10 0x000000000058f796 in mold::elf::main (argc=798697, argv=0xc2fe9) at elf/main.cc:721
#11 0x000000000041c1e7 in main (argc=6, argv=0x7fffffffdb78) at main.cc:42
```

Relevant code section:
```c++
    // Move C forward so that C is apart from B by GROUP_SIZE.
    while (c < members.size() &&
           members[c]->offset - members[b]->offset < GROUP_SIZE)
      c++;

    ...

    // Scan relocations between B and C to collect symbols that need thunks.
    tbb::parallel_for_each(&members[b], &members[c], [&](InputSection<E> *isec) {
      ...
    });
```
Even though the `tbb::parallel_for_each()` will never dereference the pointer to `members[c]`, calling `std::span::operator[]()` is already sufficient for triggering a failed assertion if `c == members.size()`.

The solution is probably to work with iterators instead of pointers. I'll try this out and submit a pull request if I succeed.

# Related issues
A similar issue (with a different stack trace) was previously reported in #69.

sicherha:
While an iterator may point beyond the last element of a `std::span`, a
pointer may not.

Fixes #298.

### DIFF
--- a/elf/arch-arm64.cc
+++ b/elf/arch-arm64.cc
@@ -574,7 +574,7 @@ static void create_thunks(Context<E> &ctx, OutputSection<E> &osec) {
     thunk.offset = offset;
 
     // Scan relocations between B and C to collect symbols that need thunks.
-    tbb::parallel_for_each(&members[b], &members[c], [&](InputSection<E> *isec) {
+    tbb::parallel_for_each(members.begin() + b, members.begin() + c, [&](InputSection<E> *isec) {
       std::span<ElfRel<E>> rels = isec->get_rels(ctx);
       isec->range_extn.resize(rels.size());
 
@@ -619,7 +619,7 @@ static void create_thunks(Context<E> &ctx, OutputSection<E> &osec) {
     }
 
     // Scan relocations again to fix symbol offsets in the last thunk.
-    tbb::parallel_for_each(&members[b], &members[c], [&](InputSection<E> *isec) {
+    tbb::parallel_for_each(members.begin() + b, members.begin() + c, [&](InputSection<E> *isec) {
       std::span<ElfRel<E>> rels = isec->get_rels(ctx);
 
       for (i64 i = 0; i < rels.size(); i++) {


rui314:
Thank!

PR #300, opened 2022-01-24
build-sys: allow overriding pkg-config

Keruspe:
On Exherbo, pkg-config is arch-prefixed like the rest of the toolchain and raw pkg-config is forbidden by the package manager

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,9 @@ ifeq ($(origin CXX), default)
   CXX = c++
 endif
 
+# Allow overriding pkg-config binary
+PKG_CONFIG = pkg-config
+
 # If you want to keep symbols in the installed binary, run make with
 # `STRIP=true` to run /bin/true instead of the strip command.
 STRIP = strip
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -16,6 +16,9 @@ ifeq ($(origin CXX), default)
   CXX = c++
 endif
 
+# Allow overriding pkg-config binary
+PKG_CONFIG ?= pkg-config
```
rui314, line 20: Can this be just `=` instead of `=?` just like other assignments?

Keruspe: Is there any benefit from this? `=?` takes the value from env if present, `=` doesn't and we then need to pass `PKG_CONFIG=${PKG_CONFIG}` to make which feels... redundant

rui314: Well, even though I saw both `=` and `?=` in the wild, it looks like `=` is more conventional than `?=` (e.g. https://www.gnu.org/prep/standards/html_node/Command-Variables.html), and since we use `=` for all the other assignment, it is also better for the sake of consistency.

In addition to that, I think I personally prefer being more explicit even if it's a bit redundant here. As long as you are using `=`, your makefile works the same if you pass the same arguments. If it implicitly takes arguments from environments, it becomes a bit more tricky. As a person who has to handle bug reports, I prefer the former.

Keruspe: Got it, updated!

rui314:
Thanks!

PR #301, opened 2022-01-24
fix build with libc++

Keruspe:
Otherwise fails with:

/usr/include/c++/v1/__threading_support:370:43: error: use of undeclared identifier 'PTHREAD_MUTEX_RECURSIVE'

200809L triggers __USE_XOPEN2K8 to be defined which enables pthread.h to define the missing identifier.


rui314:
Thanks!

PR #302, opened 2022-01-24
Fix Makefile for Termux on Android

joshcangit:
Makefile was installing to nowhere on Termux, Android.
Changed first line to fix this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,6 @@
-PREFIX = /usr/local
+ifndef PREFIX
+  PREFIX = /usr/local
+endif
 BINDIR = $(PREFIX)/bin
 LIBDIR = $(PREFIX)/lib
 LIBEXECDIR = $(PREFIX)/libexec


rui314:
Can't you just run `make PREFIX=$PREFIX install` to pass PREFIX?

joshcangit:
I can but I wonder when and why that line was edited from `?=` to `=`?

rui314:
Please see my comment in https://github.com/rui314/mold/pull/300

joshcangit:
I was wondering.
Maybe not just that line but you replaced all the `?=`.
So should I change that to use `ifdef`?
I can do that right now.

rui314:
I made that change so that our Makefile does not inherit variables such as PREFIX, CC, CXX, etc. from the environment. It's intentional. From 1.0.2, please pass PREFIX as `make PREFIX=$PREFIX`.

joshcangit:
Was thinking maybe I want to add checks for directories but then nevermind.
I'm closing this request.

Was also thinking to editing the README.md for this but then I also decided to leave that.
I think it to be a little optional FYI stuff.

PR #310, opened 2022-01-27
Makefile: avoid calling '$(shell)' on each $(CXX) execution

trofi:
Before the change calls to $(shell git rev-parse) and $(shell pkg-config)
happened on every mold's .o file (as make's "A = $(B)" performs late
expansion).

After the change $(shell) calls are evaluated once with "A := $(B)"
assignment.

Noticed by chance when 'make' was complaining about missing pkg-config
tool on each build step.


rui314:
Can you read CONTRIBUTION.md and add a `signed-off-by` line to your commit message? Other than that, LGTM.

trofi:
Sounds good. Added a `Signed-off-by:`

rui314:
Thanks!

PR #311, opened 2022-01-27
Ignore unknown --warn-* options

sina-ht:
Building llvm-project-13.0.0 we encounter an error:

[4131/6160] Linking C shared library lib/libomp.so
FAILED: lib/libomp.so
(snip)
mold: unknown command line option: --warn-shared-textrel

There're several --warn-* options implemented in ld.bfd from newer binutils.
When configuration stage we use 'ld' not necessarily mold, but ld.bfd, to
figure out what options 'ld' understands. Then unknown options are passed to
mold. Just ignore --warn-* that mold doesn't understand. --warn-* would be safe
to ignore.

Signed-off-by: Hiroshi Takekawa <sian.ht@gmail.com>

### DIFF
--- a/elf/cmdline.cc
+++ b/elf/cmdline.cc
@@ -873,6 +873,9 @@ void parse_nonpositional_args(Context<E> &ctx,
       remaining.push_back("-push-state");
     } else if (read_flag(args, "pop-state")) {
       remaining.push_back("-pop-state");
+    } else if (args[0].starts_with("--warn-") && args[0].size() > 7) {
+      Warn(ctx) << "unknown command line option: --warn-* " << args[0];
+      args = args.subspan(1);
     } else if (args[0].starts_with("-z") && args[0].size() > 2) {
       Warn(ctx) << "unknown command line option: " << args[0];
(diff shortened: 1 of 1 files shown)


rui314:
Thanks, but I think ignoring all --warn- options including future ones is a bit too much to fix your problem. We can instead just implement them. Let me try to do that.

sina-ht:
Of course, implementing them is way much better, actually, the best!
Thanks!

Below is an explanation for my change:
I found you ignored unknown -z xxx options, so I thought we could just ignore unknown --warn-* options until they're implemented, assuming we could safely ignore warnings, we wouldn't encounter errors even if new --warn-* options are implemented by other linkers.
As you felt, I admit it'd be overkill.

(Ideally we should also use mold when configuring build system (such as ./configure.))

rui314:
I implemented the --warn-shared-textrel flag in https://github.com/rui314/mold/commit/03e0cef726571fa0f6435ae0ef9d4146a285b1f8.

sina-ht:
Fantastic!
Thanks!

PR #313, opened 2022-01-27
Use nettle for SHA256

### LINKED ISSUE #246: Uses deprecated OpenSSL APIs
OpenSSL 3 has deprecated a fair chunk of the EVF APIs.  For example:

```
elf/icf.cc:234:18: warning: ‘int SHA256_Update(SHA256_CTX*, const void*, size_t)’ is deprecated: Since OpenSSL 3.0 [-Wdeprecated-declarations]
  234 |     SHA256_Update(&sha, &val, sizeof(val));
      |     ~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
```

The non-deprecated APIs existed in OpenSSL 1.1, so you can migrate to the new API without breaking the build.

cryptomilk:
As mold currently uses deprecated SHA function from OpenSSL, lets just move to libnettle. This is widely available as it is used by GnuTLS as the low level crypto library.

This fixes #246

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,8 +102,8 @@ endif
 
 # Use pkg-config to know where libcrypto resides.
 ifneq ($(OS), Darwin)
-  MOLD_CXXFLAGS += $(shell $(PKG_CONFIG) --cflags-only-I openssl)
-  MOLD_LDFLAGS += $(shell $(PKG_CONFIG) --libs-only-L openssl) -lcrypto
+  MOLD_CXXFLAGS += $(shell $(PKG_CONFIG) --cflags nettle)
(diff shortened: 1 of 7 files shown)


rui314:
Thanks! However, I think that I don't want to remove the dependency to OpenSSL and add a new one to GNU nettle, as the OpenSSL's license is more liberal than GNU's. As stated in https://github.com/rui314/mold/issues/246#issuecomment-1011662433, I'm fine with OpenSSL except that newly introduced deprecation warnings. We should fix that warnings instead. Let me try to do that.

rui314:
Feature-wise, what nettle provides is exactly what we needed, though.

PR #314, opened 2022-01-28
mold-wrapper: Execute mold if ld.bfd is executed

kofuk:
Current `is_ld` function doesn't regard `ld.bfd` as ld.
By this change, mold will be used even if compiler is about to execute `ld.bfd`.


rui314:
Can you read https://github.com/rui314/mold/blob/main/CONTRIBUTING.md and add a `signed-off-by` line to your commit message? Other than that, LGTM.

kofuk:
Fixed.
Sorry that I forgot to do so.

rui314:
Thanks!

Issue #317, opened 2022-01-30
Docker build fail for a lot of `undefined reference`


t0ny-peng:
Hi Rui314. First thanks for the great work on mold. The performance test is really amazing. 

Btw, I'm trying to use the script you provide to build mold within a container. However it failed for the following message. An example error would be
```
/usr/bin/ld: out/macho/main.o: in function `mold::MappedFile<mold::macho::Context<mold::macho::ARM64> >* mold::macho::find_framework<mold::macho::ARM64>(mold::macho::Context<mold::macho::ARM64>&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >)':
```
[output.log](https://github.com/rui314/mold/files/7965539/output.log)
Since I only plan to use `mold` on x86_64, not sure why it still wants to find `ARM64`. Or does `mold` supports cross-platform linking natively(that'd be a nice feature).

Here's my docker and OS version
```
<username>@my-desktop:~/git/mold$ docker --version
Docker version 20.10.11, build dea9396
<username>@my-desktop:~//git/mold$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.3 LTS
Release:        20.04
Codename:       focal
```

rui314:
If you check out `57f996f80693b3061bb9914bf70285a15cfffecd` (by `git checkout 57f996f80693b3061bb9914bf70285a15cfffecd`) and try again, will it fix your problem?

mold always supports cross-linking, so even if you are building mold on x86-64, it'll build code for ARM64. So that part is not an error.

t0ny-peng:
@rui314 Thanks. I realized that `libc++-dev` needs to be installed in the Docker image. A PR was created to fix that. Please review. Thx

rui314:
It looks like your `ubuntu:20.04` image is different from my `ubuntu:20.04` image. I pulled that image 7 months ago. Since then, I think they made a change to `ubuntu:20.04` image which caused this issue. If that's the case, even if we fix the problem now, it could happen again in the future. We should specify the exact image ID instead of just `20.04`.

If you change `FROM ubuntu:20.04` in `build-static.sh` to `FROM ubuntu@sha256:cf31af331f38d1d7158470e095b132acd126a7180a54f263d386da88eb681d93`, does mold compile?

t0ny-peng:
[...]

PR #318, opened 2022-01-30
Install libc++dev in the Docker image to build mold

t0ny-peng:
This PR installs `libc++-dev` in the Docker to build `mold`. I've verified that it worked on both MacOS and Ubuntu 20.04.

Screen Shot 2022-01-30 at 11 11 13 AM

close https://github.com/rui314/mold/issues/317


rui314:
Instead of pulling libc++, I hard-coded the SHA256 ID of the Ubuntu image in 3232c1f8840885e5656c8cee7fab5f8e52926bd8 so that the result is reproducible.

PR #321, opened 2022-01-31
Fix -Wmissing-field-initializers

dmantipov:
Add `mold::elf::GlobPattern::Element` and `mold::macho::SplitInfo`
trivial constructors to fix all `-Wmissing-field-initializers` as
reported both by g++ and clang++.


rui314:
Thanks!

PR #322, opened 2022-01-31
Use libsodium for SHA256

### LINKED ISSUE #246: Uses deprecated OpenSSL APIs

joshcangit:
Currently, mold uses deprecated SHA functions from OpenSSL.
If or when they are no longer usable, mold should migrate to libsodium instead.
Not only is it widely available, it also has blake2b functionality.
Furthermore, libsodium has blake3 on its 1.0.x roadmap.

This fixes #246

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: apt-get
-      run: sudo dpkg --add-architecture i386 && sudo apt-get update && sudo apt-get install build-essential libstdc++-10-dev cmake clang libssl-dev zlib1g-dev libxxhash-dev libtbb-dev git bsdmainutils dwarfdump libc6-dev:i386 lib32gcc-10-dev libstdc++-10-dev-arm64-cross gcc-aarch64-linux-gnu g++-aarch64-linux-gnu qemu-user
+      run: sudo dpkg --add-architecture i386 && sudo apt-get update && sudo apt-get install build-essential libstdc++-10-dev cmake clang libsodium-dev zlib1g-dev libxxhash-dev libtbb-dev git bsdmainutils dwarfdump libc6-dev:i386 lib32gcc-10-dev libstdc++-10-dev-arm64-cross gcc-aarch64-linux-gnu g++-aarch64-linux-gnu qemu-user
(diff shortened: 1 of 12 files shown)


rui314:
Thanks, but I want to be careful about adding a new library dependency (that's one reason why I include libraries in `third-party` di). libsodium may be popular, but OpenSSL should be much more popular and the de facto choice on most platforms. The only problem with OpenSSL is that they started deprecating the functions we are using, but I expect that they'll be available for many more years, so I think that's not a problem in practice.

If libsodium's SHA256 or its Blake3 is faster than OpenSSL's SHA256, I think we should consider switching to libsodium, but if all the conditions are equal, I think we still want to stick with OpenSSL.

memoryruins:
If it helps, repology lists the availability of packages among distros:

* openssl: https://repology.org/project/openssl/versions
* libsodium: https://repology.org/project/libsodium/versions

joshcangit:
Was wondering the same thing.
That's why I still left as a draft.

I think in terms of availability, libsodium is kind of close to OpenSSL.
However, it's lacking in hardware acceleration.
[libsodium and hardware acceleration](https://esp32.com/viewtopic.php?t=9243)

Nevermind, I'm closing this request.
I thought I had it figured out but looks like the search continues.
As @rui314 said, it'll be years before OpenSSL removes `SHA256_CTX` so there's plenty of time until then.

rui314:
One thing we can do is to just copy source code of SHA256 implementations to our repo. I think importing hardware-accelerated implementations of x86-64 and ARM64 and unaccelerated C version for other archs would suffice. SHA256 impls are very small, so this is a viable option. That being said, I don't think we want to do that now. OpenSSL works great for us at the moment.

PR #328, opened 2022-02-05
[ELF] add method to check whether an input file is shared object

dmantipov:
Prefer pure virtual `mold::elf::InputFile::is_dso()` (actually
implemented in `mold::elf::ObjectFile` and `mold::elf::SharedFile`)
over the field of the same name, adjust related code.

### DIFF
--- a/elf/input-files.cc
+++ b/elf/input-files.cc
@@ -15,8 +15,6 @@ InputFile<E>::InputFile(Context<E> &ctx, MappedFile<Context<E>> *mf)
     Fatal(ctx) << *this << ": not an ELF file";
 
   ElfEhdr<E> &ehdr = *(ElfEhdr<E> *)mf->data;
-  is_dso = (ehdr.e_type == ET_DYN);
-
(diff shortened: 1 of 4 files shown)


rui314:
Overall looking good. Originally, I didn't use any virtual functions in this class, so it made sense to keep it a variable. But now we have virtual functions, it feels more natural to have this predicate as a virtual function.

rui314:
Is it OK to merge this commit? Because I'm making lots of changes to mold, any outstanding PRs quickly become stale and fail to compile. I want to merge this one now (even if it doesn't compile) and fix any issues just after that.

dmantipov:
Hopefully I've resolved the conflicts just now.

rui314:
Thanks!

rui314:
I think we should revert this change as it adds the cost of virtual function dispatch to very hot functions such as `Symbol::get_addr`. `is_dso` just returns constant true or false, but the function is too expensive now.

PR #331, opened 2022-02-07
Restore `SYSTEM_XXHASH` option

sicherha:
Even though we no longer link against `libxxhash.so`, we still want to provide the option to build mold against the system-provided xxHash headers. This is particularly useful for downstream distributions.

This reverts parts of commit e5f4b967ed9339fc215dadcef62d656c6d98228a.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,8 +40,8 @@ CFLAGS = -O2
 CXXFLAGS = -O2
 
 MOLD_CXXFLAGS := -std=c++20 -fno-exceptions -fno-unwind-tables \
-                 -fno-asynchronous-unwind-tables -Ithird-party/xxhash \
-                 -DMOLD_VERSION=\"1.0.3\" -DLIBDIR="\"$(LIBDIR)\""
+                 -fno-asynchronous-unwind-tables -DMOLD_VERSION=\"1.0.3\" \
+                 -DLIBDIR="\"$(LIBDIR)\""
 
 MOLD_LDFLAGS := -pthread -lz -lm
 
@@ -81,6 +81,10 @@ else
   TBB_LIB = out/tbb/libs/libtbb.a
   MOLD_LDFLAGS += $(TBB_LIB)
   MOLD_CXXFLAGS += -Ithird-party/tbb/include
+endif
+
+ifndef SYSTEM_XXHASH
+  MOLD_CXXFLAGS += -Ithird-party/xxhash
 endif
 
 ifeq ($(OS), Linux)


rui314:
In what situation do you want to #include against system-installed xxhash.h instead of the bundled one?

sicherha:
Essentially, in **all** situations. 🙂

From a distro maintainer's point of view, bundling should only be used a last resort. Most major distributions explicitly recommend against it, e.g. [Debian](https://www.debian.org/doc/debian-policy/ch-source.html#s-embeddedfiles), [Fedora](https://docs.fedoraproject.org/en-US/fesco/Bundled_Software_policy/) and [Gentoo](https://wiki.gentoo.org/wiki/Why_not_bundle_dependencies).

For example, let's assume a severe bug were discovered and fixed in xxHash. Downstream distributions would then patch or update their xxHash package.
* If the mold package is built against the system headers, distro maintainers only need to trigger a rebuild of mold.
* If the mold package is built against its bundled copy of xxHash, distro maintainers *additionally need to backport the patch to that bundled copy*, then rebuild mold.

rui314:
I understand that unbundling is generally preferred by package maintainers, but this case is a bit different. First of all, it's extremely unlikely that you would find a bug in xxHash. It's not a cryptographic hash. Even if there's a bug in xxhash, we already know that it works fine at least for us, so we don't need to upgrade. We do not depend on hash value uniqueness nor hash value reproducibility. As long as the hash function produces the same hash value within the process, that works for us. We do not store hash values in any external storage or anything.

You can think of xxhash as a very internal implementation detail of the mold linker. We already have a few hash functions embedded to our code base (e.g. https://github.com/rui314/mold/blob/main/elf/output-chunks.cc#L12). Other linkers also embeds similar functions in their linkers, so there's no easy way to "upgrade" or "fix" these hash functions. But that's OK because they are not a moving part, unlike OpenSSL or something like that.

From the viewpoint of this linker's developer, I don't really want to make it configurable. We know that the xxhash function in our repo compiles and works fine. But what if the upstream xxhash makes an API-breaking change? Or what if their new release has a bug? It looks like allowing an external xxhash.h just makes the situation a bit more complicated than necessary without any benefit.

There are bunch of C++ "header-only" libraries out there, and many projects just copies a single header file to their repo and use it. Our situation with xxhash is exactly like that. We use just one inline function by including a header file.

BenBE:
While I agree that, from a maintainer's point of view, having all the moving parts like external libraries under control, it's a bit different for those people packaging things for a distribution. Where possible, the distribution people try to minimize duplicate code across packages. While these may seem diametral they can – and most times do – align quite well, as the distribution people try to keep up-to-date with what the upstream maintainer requires. So if your embedded copy of xxHash needs to be a certain version, the distribution will likely go and match that version as closely as permitted; and only resort to the bundled version if there are major complications with the distribution version of such a dependency. But: issues related to these version differences are for the distribution to resolve usually, so apart from some – usually minor – changes to help the distribution people package some software (nearly) no ongoing effort is required off the upstream author.

So, IMHO all asked for is to allow the distributions to specify some `./configure --with-xxhash=/path/for/distro/version` which takes the library from the distribution instead of the vendored copy which can be kept the default if nothing else is specified.

rui314:
While I understand that using un-vendored library is usually considered a best practice for package managers, I disagree that that is beneficial for this specific case. The more I think about it, the more it feels that there's no strong reason to use system-installed xxhash.h instead of the one in our repo.

We are now using xxhash as a header-only library. If the system's header works as expected, the resulting mold executable will exactly be the same as the one built with xxhash.h in our repo. It cannot fix anything but can break. It also adds an extra build dependency to the xxhash library although the resulting mold executable will not be different from the one built with xxhash.h in our repo. It looks like it just adds one moving part that can break without a real benefit. If it were a big inline function, the situation might have been different, but what we are talking about is a small inline function that computes a noncryptographic hash value, so it shouldn't be a big deal. I think we can just include the same header.

sicherha:
I understand your point of view and your line of argument, and I freely admit that my personal stance on the topic of code bundling is a wee bit on the extremist side. 😉
Still, does it hurt to provide an opt-in, off-by-default choice for people who know what they're doing? We're talking about three lines of code in the `Makefile`.

rui314:
I'm sorry, but I think the amount of code doesn't matter much in this case, as it is more like whether we want this as a public interface or not. If we provide SYSTEM_XXHASH, its our commitment to support that usage, but it is honestly hard for me to make that commitment for the reasons I explained above.

sicherha:
Thanks Rui for the open discussion!

I respect your reasoning and your decision, so let's close this pull request. For those who wish to build against the system-provided xxHash headers anyway - *at their own responsibility!* - there exists a [trivial workaround](https://src.fedoraproject.org/rpms/mold/blob/rawhide/f/mold.spec#_72).

rui314:
I'm sorry that I couldn't satisfy your wish this time, and thank you very much for proposing something new to the project and having an open discussion!

PR #337, opened 2022-02-09
[ELF][MACHO] Remove unused data structures

dmantipov:
Drop unused (older leftovers?) `mold::elf::Context::obj_pool`,
`mold::elf::Context::dso_pool`, `mold::macho::Context::obj_pool`
and `mold::macho::Context::dylib_pool`, adjust related code.

### DIFF
--- a/elf/input-files.cc
+++ b/elf/input-files.cc
@@ -76,9 +76,7 @@ template <typename E>
 ObjectFile<E> *
 ObjectFile<E>::create(Context<E> &ctx, MappedFile<Context<E>> *mf,
                       std::string archive_name, bool is_in_lib) {
-  ObjectFile<E> *obj = new ObjectFile<E>(ctx, mf, archive_name, is_in_lib);
-  ctx.obj_pool.push_back(std::unique_ptr<ObjectFile<E>>(obj));
-  return obj;
+  return new ObjectFile<E>(ctx, mf, archive_name, is_in_lib);
 }
 
 template <typename E>
@@ -1178,9 +1176,7 @@ std::ostream &operator<<(std::ostream &out, const InputFile<E> &file) {
 template <typename E>
 SharedFile<E> *
 SharedFile<E>::create(Context<E> &ctx, MappedFile<Context<E>> *mf) {
-  SharedFile<E> *obj = new SharedFile(ctx, mf);
-  ctx.dso_pool.push_back(std::unique_ptr<SharedFile<E>>(obj));
-  return obj;
+  return new SharedFile(ctx, mf);
 }
 
 template <typename E>
--- a/elf/mold.h
+++ b/elf/mold.h
@@ -1530,8 +1530,6 @@ struct Context {
   tbb::concurrent_vector<std::unique_ptr<TimerRecord>> timer_records;
   tbb::concurrent_vector<std::function<void()>> on_exit;
 
-  tbb::concurrent_vector<std::unique_ptr<ObjectFile<E>>> obj_pool;
-  tbb::concurrent_vector<std::unique_ptr<SharedFile<E>>> dso_pool;
   tbb::concurrent_vector<std::unique_ptr<u8[]>> string_pool;
   tbb::concurrent_vector<std::unique_ptr<MappedFile<Context<E>>>> mf_pool;
   tbb::concurrent_vector<std::vector<ElfRel<E>>> rel_pool;
--- a/elf/passes.cc
+++ b/elf/passes.cc
@@ -273,7 +273,6 @@ static std::optional<u64> parse_defsym_addr(std::string_view s) {
 template <typename E>
 ObjectFile<E> *create_internal_file(Context<E> &ctx) {
   ObjectFile<E> *obj = new ObjectFile<E>;
-  ctx.obj_pool.push_back(std::unique_ptr<ObjectFile<E>>(obj));
 
   // Create linker-synthesized symbols.
   auto *esyms = new std::vector<ElfSym<E>>(1);
--- a/macho/main.cc
+++ b/macho/main.cc
@@ -23,7 +23,6 @@ split_string(std::string_view str, char sep) {
 template <typename E>
 static void create_internal_file(Context<E> &ctx) {
   ObjectFile<E> *obj = new ObjectFile<E>;
-  ctx.obj_pool.push_back(std::unique_ptr<ObjectFile<E>>(obj));
   ctx.objs.push_back(obj);
 
   auto add = [&](std::string_view name) {
--- a/macho/mold.h
+++ b/macho/mold.h
@@ -806,8 +806,6 @@ struct Context {
   std::unique_ptr<OutputFile<E>> output_file;
   u8 *buf;
 
-  tbb::concurrent_vector<std::unique_ptr<ObjectFile<E>>> obj_pool;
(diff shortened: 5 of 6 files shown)


sicherha:
The pools (vectors of `unique_ptr`s) you want to remove here are not redundant; they ensure that the lifetimes of all these dynamically allocated objects are tied to the lifetime of the surrounding `Context` object. Once the big `Context` object goes out of scope, all the little objects tied to the context are automatically deallocated along with it.

dmantipov:
> The pools (vectors of `unique_ptr`s) you want to remove here are not redundant; they ensure that the lifetimes of all these dynamically allocated objects are tied to the lifetime of the surrounding `Context` object. Once the big `Context` object goes out of scope, all the little objects tied to the context are automatically deallocated along with it.

In case of ELF/X86_64, could you hit the breakpoint installed to `mold::elf::ObjectFile<mold::elf::X86_64>::~ObjectFile()` and/or `mold::elf::SharedFile<mold::elf::X86_64>::~SharedFile()` during normal `mold` invocation?

sicherha:
Sure, these breakpoints would get hit when `elf_main()` returns. Note that in order to observe this, you need to `set follow-fork-mode child` in GDB since mold carries out all its actual work within a dedicated child process.

rui314:
As @dmantipov pointed out, they are for lifetime management. The point is that I want a use case in which you call mold's main function from other program. So we can't remove these members.

rui314:
@sicherha I usually pass `-no-fork` and `-no-threads` developer flags to mold when running under gdb. FYI.

dmantipov:
> Sure, these breakpoints would get hit when `elf_main()` returns. Note that in order to observe this, you need to `set follow-fork-mode child` in GDB since mold carries out all its actual work within a dedicated child process.

Hmm...am I doing wrong something?

```
$ cat t.c
void t (void) {
}
$ gcc -fPIC -c t.c
$ gdb -q ld.mold
Reading symbols from ld.mold...
(gdb) b mold::elf::ObjectFile<mold::elf::X86_64>::~ObjectFile() 
Breakpoint 1 at 0x5ab498: mold::elf::ObjectFile<mold::elf::X86_64>::~ObjectFile(). (2 locations)
(gdb) b mold::elf::SharedFile<mold::elf::X86_64>::~SharedFile() 
Breakpoint 2 at 0x5adab0: mold::elf::SharedFile<mold::elf::X86_64>::~SharedFile(). (2 locations)
(gdb) i b
Num     Type           Disp Enb Address            What
1       breakpoint     keep y   <MULTIPLE>         
1.1                         y   0x00000000005ab498 in mold::elf::ObjectFile<mold::elf::X86_64>::~ObjectFile() at elf/mold.h:1011
1.2                         y   0x00000000005ab5aa in mold::elf::ObjectFile<mold::elf::X86_64>::~ObjectFile() at elf/mold.h:1011
2       breakpoint     keep y   <MULTIPLE>         
2.1                         y   0x00000000005adab0 in mold::elf::SharedFile<mold::elf::X86_64>::~SharedFile() at elf/mold.h:1094
2.2                         y   0x00000000005adb20 in mold::elf::SharedFile<mold::elf::X86_64>::~SharedFile() at elf/mold.h:1094
(gdb) r --no-fork --no-threads -shared -o libt.so t.o
Starting program: /usr/bin/ld.mold --no-fork --no-threads -shared -o libt.so t.o
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
[Inferior 1 (process 8865) exited normally]
(gdb) 
```

dmantipov:
One more observation: `mold::elf::ObjectFile` is backed by `mold::MappedFile` to abstract over memory-mapped I/O. Since `mold::elf::MappedFile::~MappedFile()` has an explicit `munmap()` call, it should be displayed when running under `strace`. But what I'm seeing is that object file gets `mmap()`ed, e.g.

```
openat(AT_FDCWD, "t.o", O_RDONLY)       = 3
newfstatat(3, "", {st_mode=S_IFREG|0644, st_size=1224, ...}, AT_EMPTY_PATH) = 0
mmap(NULL, 1224, PROT_READ, MAP_PRIVATE, 3, 0) = 0x7f1fcc16c000
close(3)                                = 0
```

but there is no matching `munmap()` call for `0x7f1fcc16c000`.

rui314:
We call `std::quick_exit()` on exit to quickly exit from the program. No constructors or functions registered with `atexit(3)` are called on quick_exit (that's why it's quicker). If you want to see the destructor to work, pass `-no-quick-exit` to mold.

dmantipov:
Oops. By default, there is no normal return from `elf_main()` because `ctx.arg.quick_exit` is `true` and so `Context` and all of its members are not destroyed in regular way by calling the corresponding destructors.

Anyway, having `std::vector<ObjectFile<E> *> objs` for normal processing and `tbb::concurrent_vector<std::unique_ptr<ObjectFile<E>>> obj_pool` (with the same contents) just for resource management looks redundant. Shouldn't this be managed with reference-counting scheme instead, probably via `std::shared_ptr<ObjectFile<E>>`?

rui314:
`std::shared_ptr` is slow. In general, managing object with reference counting is slow because every time new reference is created, it has to mutate the shared pointer. I don't see a need to change the scheme how we manage objects here. We have pools to manage object lifetime and use pool-managed objects as if they were permanent objects throughout its corresponding Context object lifetime.

dmantipov:
> `std::shared_ptr` is slow. In general, managing object with reference counting is slow because every time new reference is created, it has to mutate the shared pointer. I don't see a need to change the scheme how we manage objects here. We have pools to manage object lifetime and use pool-managed objects as if they were permanent objects throughout its corresponding Context object lifetime.

Even if so, anyway, why manage an extra vector of `smart` pointers just to arrange destructor calls? Why not just use `std::vector<std::unique_ptr<ObjectFile<E>>> objs`? This should not cause an issues passing normal (read: fast) C-style `ObjectFile<E> *` pointers during all of the `Context` object lifetime.

rui314:
Maybe we can change it to `std::vector<std::unique_ptr<ObjectFile<E>>>`, but there are a lot of loops that iterates over input objects, and I don't want to rewrite them from `for (ObjectFile *file : ctx.objs)` to `for (std::unique_ptr<E> &file : ctx.objs)`. The latter is cognitively too verbose.

dmantipov:
> Maybe we can change it to `std::vector<std::unique_ptr<ObjectFile<E>>>`, but there are a lot of loops that iterates over input objects, and I don't want to rewrite them from `for (ObjectFile *file : ctx.objs)` to `for (std::unique_ptr<E> &file : ctx.objs)`. The latter is cognitively too verbose.

If you don't like `auto` in this context, this may be solved by `template <typename E> using ObjPtr = std::unique_ptr<ObjectFile<E>>&`  and so using `for (ObjPtr<E> file : ctx.objs)` instead.

rui314:
It should work, but it's more like a personal preference than fixing a bug or something. So there's no right or wrong. Since the entire code base is written in my style, I'd like to keep it as-is.

rui314:
Note that I prefer writing an underlying type such as `std::unique_ptr<ObjectFile<E>>` instead of typedef'ing that type and use that as a type identifier. I prefer the former for the same reason as I prefer to write an explicit type instead of `auto`.

dmantipov:
Closed. Nevertheless I think there is a room for an improvements in this area.

PR #343, opened 2022-02-14
[ELF] Fix -Wunused-variable warnings

dmantipov:
Fix the following `-Wunused-variable` warnings as reported by `gcc`:
```
elf/input-sections.cc:116:8: warning: unused variable ‘is_code’
[-Wunused-variable]
  116 |   bool is_code = (shdr().sh_flags & SHF_EXECINSTR);
```
```
elf/lto.cc:514:16: warning: unused variable ‘st’ [-Wunused-variable]
  514 |   PluginStatus st = claim_file_hook(file, &claimed);
```

### DIFF
--- a/elf/input-sections.cc
+++ b/elf/input-sections.cc
@@ -113,7 +113,6 @@ template <typename E>
 void InputSection<E>::dispatch(Context<E> &ctx, Action table[3][4], i64 i,
                                const ElfRel<E> &rel, Symbol<E> &sym) {
   Action action = table[get_output_type(ctx)][get_sym_type(sym)];
-  bool is_code = (shdr().sh_flags & SHF_EXECINSTR);
   bool is_writable = (shdr().sh_flags & SHF_WRITE);
 
   auto error = [&] {


elf/lto.cc
```diff
@@ -511,7 +511,8 @@ ObjectFile<E> *read_lto_object(Context<E> &ctx, MappedFile<Context<E>> *mf) {
 
   // claim_file_hook() calls add_symbols() which initializes `plugin_symbols`
   int claimed = false;
-  PluginStatus st = claim_file_hook(file, &claimed);
+  // Temporarily unused.
+  [[maybe_unused]] PluginStatus st = claim_file_hook(file, &claimed);
```
rui314, line 515: We should change this to `Fatal` rather than ignoring an error if assert is not enabled. I'll make such change.

rui314:
Please rebase.

elf/lto.cc
dmantipov: OK

PR #344, opened 2022-02-14
[PERF] Use std::call_once to arrange timer record stop

dmantipov:
Prefer `std::once_flag` and `std::call_once()` over plain
C-style flag to arrange (the body of) `TimerRecord::stop()`
to execute just once, adjust related code.

### DIFF
--- a/mold.h
+++ b/mold.h
@@ -543,12 +543,12 @@ struct TimerRecord {
 
   std::string name;
   TimerRecord *parent;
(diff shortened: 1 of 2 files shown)


rui314:
`std::call_once` is used to make the thing that is protected by `call_once` thread-safe. In this case, `stop` is called by a single thread, so we don't need to use `call_once`.

PR #350, opened 2022-02-17
[ELF] Use std::string references when processing options

dmantipov:
Prefer `std::string &` when calling and processing the
result of `add_dashes()`, thus saving a few calls to
`std::string` copy constructor.

### DIFF
--- a/elf/cmdline.cc
+++ b/elf/cmdline.cc
@@ -172,7 +172,7 @@ Options:
 mold: supported targets: elf32-i386 elf64-x86-64 elf64-littleaarch64
 mold: supported emulations: elf_i386 elf_x86_64 aarch64linux aarch64elf)";
 
-static std::vector<std::string> add_dashes(std::string name) {
+static std::vector<std::string> add_dashes(std::string &name) {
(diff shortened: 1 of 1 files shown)


rui314:
We don't need to optimize the code in this file, so conservatively copy strings is fine. Generally, we should avoid optimizing code until it is proven to be a bottleneck.

LowLevelMahn:
but they can be made const std::string& - that would makes it also clear that they do not change inside and they would not copy, or? does fall really fall under "premature optimization"?

rui314:
In terms of performance, that doesn't make any measurable difference in this case. So it's more like a personal preference, and I feel conservatively copying strings is at least just as clear than passing a const reference.

jonesmz:
Bottlenecks not withstanding, wouldn't you want to use std::string_view, over a non-const lvalue-reference anyway?

rui314:
In general, we prefer `std::string_view`, but I guess there was a reason to use `std::string` here. For example, you can't concatenate `std::string_view` with `operator+`.

PR #355, opened 2022-02-21
Set default CFLAGS and CXXFLAGS values only if they are not already set.

fcambus:
Signed-off-by: Frederic Cambus <fred@statdns.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,8 +42,8 @@ endif
 
 # If you want to compile mold for debugging, invoke make as
 # `make CXXFLAGS=-g`.
-CFLAGS = -O2
-CXXFLAGS = -O2
+CFLAGS ?= -O2
+CXXFLAGS ?= -O2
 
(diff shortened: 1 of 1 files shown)


rui314:
You can set `CFLAGS` and `CXXFLAGS` by passing them as arguments rather than environment variable. E.g. `make CFLAGS=-g`.

fcambus:
I'm aware of this, but why not also allow to pass them as environment variables?

Do you see any drawbacks in doing so?

rui314:
From the point of maintainer's view, making it explicit makes user support easier. Currently, if a user do not explicitly pass CFLAGS or CXXFLAGS, I'm sure no flags are passed. If we accept both environment variables and command line arguments, I have to ensure that the user didn't set a bogus CFLAGS/CXXFLAGS environment variable. So, yes, I saw a drawback in doing this.

fcambus:
Thanks for explaining your rationale. Fine then, closing the issue.

PR #369, opened 2022-02-27
platform: Add basic support for Haiku

kallisti5:


kallisti5:
Still a WIP given the last error below:

* Mold changes were pretty simple
* tbb fixes might be misplaced.. anywhere to upstream?
* getting "Could not resolve symbol 'scalable_malloc'" from mold on execution. Seems tbb related?

rui314:
Here is the upstream repo of TBB. Your changes to third-party/tbb should be made here. https://github.com/oneapi-src/oneTBB

`scalable_malloc` seems to be related to TBB. TBB provides its own implementation of `malloc`, and that is `scalable_malloc`. I don't know why mold tries to resolve that function at runtime on Haiku, though.

kallisti5:
> scalable_malloc seems to be related to TBB. TBB provides its own implementation of malloc, and that is scalable_malloc. I don't know why mold tries to resolve that function at runtime on Haiku, though.

scalable_malloc is a weak symbol by TBB.  It seems odd they chose that

ADKaster:
I got those scalable_malloc errors on SerenityOS untill I added this patch:

https://github.com/SerenityOS/serenity/blob/master/Ports/mold/patches/0003-Tell-TBB-that-SerenityOS-does-not-support-weak-symbo.patch

Defining away weak symbol usage in tbb made the linker stop putting unresolved weak symbol references to "scalable_malloc" into the binary. But I would also expect it to use "malloc" as the actual final symbol name 🤷‍♂️

rui314:
I will update the in-tree TBB copy at some point in the future, and I can't maintain a local patch to TBB for Haiku, so please upstream that patch. We can only accept a local patch as a temporary measure.

kallisti5:
@rui314 sounds good. I can definitely upstream.    I can confirm the patch @ADKaster mentioned (adjusted for Haiku) fixed the final issue.

mold

kallisti5:
Ok. I dropped the tbb changes from this patchset.

It should be applicable on mold without any impact to other platforms.  I'll work on up-streaming the tbb changes to the repo mentioned.

rui314:
This patch LGTM, but let me wait until an upstream patch is accepted by the TBB folks. It shouldn't take too long. (I sent a patch to TBB to support the musl libc, and I didn't take too long.)

X547:
My current Mold patches for Haiku: https://github.com/X547/mold/commit/ccbadabd1d1596da406a81396426ee58ddfab42d.

PR #385, opened 2022-03-08
tests: Improve the version detection of ldd

sylvestre:
When running with a different locale, the copyright symbol can be different:
```
$ LANG=fr_FR.UTF-8 ldd --version | grep -P 'Copyright'
Copyright © 2021 Free Software Foundation, Inc.
```

### DIFF
--- a/test/elf/hello-static-pie.sh
+++ b/test/elf/hello-static-pie.sh
@@ -18,7 +18,7 @@ echo 'int main() {}' | $CC -o $t/exe -xc -
 ldd $t/exe | grep -q ld-musl && { echo OK; exit; }
 
 # -static-pie works only with a newer version of glibc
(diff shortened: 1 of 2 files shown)


BenBE:
Why not force `LC_ALL=C` when querying the copyright message?

rui314:
Let me unconditionally set LC_ALL=C in all scripts to fix potential locale-related issues once and for all.

joshcangit:
Using `ldd --version` works on my PC but not in Termux.
It outputs this.
```
objdump: '--version': No such file
```

rui314:
Now the error message shouldn't be displayed to the terminal.

rui314:
Fixed in a different way, so closing this PR.

PR #400, opened 2022-03-18
image-base i64 to u64 parse_unsigned_number

nick-rawlings:
Linking a limine test using stivale2 protocol required me to use `--image-base=0xffffffff80000000` which caused an error with `stol` in `parse_number`, so I created `parse_unsigned_number` to be able to use the unsigned long value in image-base. the `ctx.arg.image_base` is already `u64`

use stoul for unsigned long
image-base i64 to u64 parse_unsigned_number
Signed-off-by: nick-rawlings <100548346+nick-rawlings@users.noreply.github.com>

### DIFF
--- a/elf/cmdline.cc
+++ b/elf/cmdline.cc
@@ -298,6 +298,16 @@ static i64 parse_number(Context<E> &ctx, std::string opt,
 }
 
 template <typename E>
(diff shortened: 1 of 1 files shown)


rui314:
Thanks!

PR #408, opened 2022-03-28
[ELF] Fix issues caught by UBSan

yurai007:
While running tests on UBSan instrumented mold binary 2 separated problems are reported:

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior elf/main.cc:592:15 in
elf/output-chunks.cc:1866:24: runtime error: member call on null pointer of type 'mold::elf::VerdefSection<mold::elf::X86_64> *'

SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior elf/output-chunks.cc:1866:24 in
elf/input-files.cc:1000:8: runtime error: load of value 190, which is not a valid value for type 'bool'

This change fixes both of them.

### DIFF
--- a/elf/main.cc
+++ b/elf/main.cc
@@ -589,7 +589,8 @@ static int elf_main(int argc, char **argv) {
   ctx.dynsym->finalize(ctx);
 
   // Fill .gnu.version_d section contents.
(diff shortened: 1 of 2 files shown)


rui314:
Thanks!

PR #425, opened 2022-04-11
Make libexec symlink relocatable

peniblec:
Hi!

This is a slight tweak to the symlink in libexec to allow packagers to archive build artifacts, and untar the archive under different paths.

As shown in the commit message, the current link is absolute, so it will not be resolved properly (1) while mold sits in DESTDIR (2) if the build artifacts are deployed under a different PREFIX.

Thank you for your time, and let me know if you need more context about the use-case!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -178,7 +178,7 @@ install: all
 	$(STRIP) $D$(LIBDIR)/mold/mold-wrapper.so
 
 	$(INSTALL) -d $D$(LIBEXECDIR)/mold
-	ln -sf $(BINDIR)/mold $D$(LIBEXECDIR)/mold/ld
+	ln -rsf $D$(BINDIR)/mold $D$(LIBEXECDIR)/mold/ld
 
 	$(INSTALL) -d $D$(MANDIR)/man1
 	$(INSTALL_DATA) docs/mold.1 $D$(MANDIR)/man1


rui314:
I wanted to do this, but it looks like `ln` commands on some Unixes (at least macOS) don't support the `-r` option, so we can't use that flag.

peniblec:
Good catch. I've thought about using `realpath --relative-to`, but the BSD manpages I could find[[1]](https://www.freebsd.org/cgi/man.cgi?query=realpath&sektion=1)[[2]](https://man.openbsd.org/man1/realpath.1) suggest that this flag is also GNU-only.

So far I haven't found a shell solution to produce relative paths without GNU coreutils, which sounds mildly surprising to me; I'll mull over this some more.

rui314:
I did this using Python in the above commit.

peniblec:
Thanks! Sorry we couldn't find a more "lightweight" solution (IIUC pulling in GNU coreutils is doable on MacOS; not sure about the BSDs?).

I've updated my branch with a further tweak: adding `$D` before the link target (this was in my original PR, but not easy to spot since I didn't really advertize it). The change is here:

https://github.com/rui314/mold/compare/main...peniblec:relative-symlink

Should I submit a new PR?

rui314:
I removed `$D` from both arguments instead.

peniblec:
Thanks again, should work just as well!

PR #434, opened 2022-04-15
Skip test if `dwarfdump` is unavailable

### LINKED ISSUE #430: testsuite: please skip dead-debug-sections if dwarfdump is not installed
I miss the tool and so I face the following error:

```
Testing dead-debug-sections ... ./elf/dead-debug-sections.sh: line 40: dwarfdump: command not found
make[1]: *** [Makefile.linux:6: elf/dead-debug-sections.sh] Error 127
make[1]: *** Waiting for unfinished jobs....
```

sicherha:
Fixes #430.

### DIFF
--- a/test/elf/dead-debug-sections.sh
+++ b/test/elf/dead-debug-sections.sh
@@ -14,6 +14,8 @@ mold="$(pwd)/mold"
 t=out/test/elf/$testname
 mkdir -p $t
 
+which dwarfdump >& /dev/null || { echo skipped; exit; }
+
 cat <<EOF | $CXX -c -o $t/a.o -g -xc++ -
 #include <iostream>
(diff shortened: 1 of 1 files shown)


owl4ce:
Using the built-in shell `command -v` (POSIX) instead of `which`?

sicherha:
I used `which` for consistency with `compressed-debug-info.sh`, `i386-compressed-debug-info.sh` and `compress-debug-sections.sh`.

If we want to replace `which` with `command -v`, I would favour doing so consistently in all scripts.

rui314:
Thanks!

`which` is fine, as I believe all Unix-like systems have this command. I didn't actually know that that wasn't in POSIX.

Issue #444, opened 2022-04-19
Backtraces in gdb do not work with --gdb-index


llunak:
Gdb cannot print backtraces properly for symbols in shared libraries if --gdb-index is used. It seems backtrace printing requires address ranges in .gdb_index, but this problem actually exists even with a build before d4cc0ed3fb5f1d595cdc2e4747f46aa7944791d0 , suggesting that the address range generation wasn't correct.

$ cat a2.cpp
#include <stdlib.h>
static int func2(int var) { abort(); }
int func(int var) { return func2(var); }
$ clang++ -g -ggnu-pubnames -Wl,--gdb-index a2.cpp -fPIC -shared -o liba.so -fuse-ld=mold

$ cat a1.cpp
extern int func(int);
int main() { return func(3); }
$ clang++ -g -ggnu-pubnames -Wl,--gdb-index a1.cpp -la -L. -fuse-ld=mold

$ LD_LIBRARY_PATH=. gdb ./a.out
(gdb) r
Starting program: /home/seli/tmp/a.out 

Program received signal SIGABRT, Aborted.
0x00007ffff70b0cdb in raise () from /lib64/libc.so.6
(gdb) bt
#0  0x00007ffff70b0cdb in raise () from /lib64/libc.so.6
#1  0x00007ffff70b2375 in abort () from /lib64/libc.so.6
#2  0x00007ffff7ff0140 in func2(int) ()
#3  0x00007ffff7ff0123 in func(int) () from ./liba.so
#4  0x0000000000201149 in main () at a1.cpp:2

The backtrace should show source and argument value for func() and func2() (and it does if e.g. lld is used).

rui314:
I'm thinking of creating DWARF5 `.debug_names` and `.debug_str` sections instead of `.gdb_index`, as they are essentially a gdb-index minus addresses ranges. I'm not sure how good/bad it is, but I really want to avoid parsing lots of DWARF records just for creating address ranges.

rui314:
It looks like GNU gold doesn't fill the address area for the shared object at all, but we still get a correct stack trace for that shared object. So there must be something wrong in other parts of our output.

```
ruiu@blue:~/mold$ llvm-dwarfdump-14 --gdb-index liba.so
liba.so:      file format elf64-x86-64

.gdb_index contents:
  Version = 7

  CU list offset = 0x18, has 1 entries:
    0: Offset = 0x0, Length = 0x654

  Types CU list offset = 0x28, has 0 entries:

[...]

PR #448, opened 2022-04-21
Gdb index fixes

llunak:
Various fixes for gdb index support (#444). I can re-link entire LibreOffice and gdb seems to work fine. The commits here mostly depend on previous ones, I can't easily submit them separately.

### DIFF
--- a/elf/output-chunks.cc
+++ b/elf/output-chunks.cc
@@ -820,6 +820,8 @@ static std::string_view get_output_name(Context<E> &ctx, std::string_view name) 
       return ".debug_str";
     if (name == ".zdebug_types")
       return ".debug_types";
+    if (name == ".zdebug_addr")
+      return ".debug_addr";
     return save_string(ctx, "."s + std::string(name.substr(2)));
   }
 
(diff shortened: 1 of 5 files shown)


elf/output-chunks.cc
```diff
@@ -2392,9 +2404,11 @@ void GdbIndexSection<E>::write_address_areas(Context<E> &ctx) {
 
     // Fill trailing null entries with dummy values because gdb
     // crashes if there are entries with address 0.
-    u64 filler = (e == begin) ? ctx.etext->get_addr(ctx) - 1 : e[-1].start;
-    for (; e < begin + file->num_areas; e++)
+    u64 filler = (e == begin) ? ctx.etext->get_addr(ctx) - 1 : e[-1].end;
+    for (; e < begin + file->num_areas; e++) {
       e->start = e->end = filler;
+      e->attr = file->compunits_idx;
```
rui314, line 2410: Did this actually matter?

rui314:
Thank you very much! Let's see how it works.

elf/output-chunks.cc
llunak: I don't know if it matters in practice, as the commit log message says, I wasn't sure why GDB was crashing. I added this more for safety, and also to avoid uninitialized values when dumping .gdb_index using llvm-dwarfdump.

llunak:
BTW, note that the un-reverted dwarf5 test lacks changes that you've meanwhile done to the other gdb-index tests.

rui314:
Thank you for pointing that out. I updated the test.

PR #450, opened 2022-04-21
Use -batch mode for gdb tests.

marxin:
Use the mode, otherwise one can be asked things like:
Quit anyway? (y or n)

which make the test stuck.


rui314:
Thanks!

PR #455, opened 2022-04-21
[ELF] do not add 0 addresses to .gdb_index

llunak:
GDB crashes in that case in dw2_find_pc_sect_compunit_symtab() for a reason I do not understand. This is presumably what I was trying to avoid already in e0dd53b4ad3ccf39568cc50d8553ec561938463e but didn't quite manage (so reverting that part).

I don't understand the problem, but simply dropping 0 address entries seems to be consistent with the "filler" code below, and LLD also does not include the entries despite .debug_rnglists including those entries that start at 0 address.


rui314:
I don't understand the exact behavior too, but since this stuff is gdb-specific and not documented well, I think the actual gdb behavior is the gold standard. We just need to make it happy.

Issue #453, opened 2022-04-21
testsuite: ./test/elf/gdb-index-dwarf3.sh fails with fatal: /lib64/crtn.o: --gdb-index: missing debug_ranges


marxin:
I see the following failures:

```
$ ./test/elf/gdb-index-dwarf3.sh
Testing gdb-index-dwarf3 ... mold: fatal: /usr/lib64/crt1.o: --gdb-index: unknown debug info form: 0xf
mold: fatal: /usr/lib64/crti.o: --gdb-index: missing debug_rangescollect2: error: ld returned 1 exit status

$ readelf -SW /usr/lib64/crti.o
There are 23 section headers, starting at offset 0x608:

Section Headers:
  [Nr] Name              Type            Address          Off    Size   ES Flg Lk Inf Al
  [ 0]                   NULL            0000000000000000 000000 000000 00      0   0  0
  [ 1] .text             PROGBITS        0000000000000000 000040 000000 00  AX  0   0  1
  [ 2] .data             PROGBITS        0000000000000000 000040 000000 00  WA  0   0  1
  [ 3] .bss              NOBITS          0000000000000000 000040 000000 00  WA  0   0  1
  [ 4] .note.gnu.property NOTE            0000000000000000 000040 000050 00   A  0   0  8
  [ 5] .init             PROGBITS        0000000000000000 000090 000016 00  AX  0   0  4
  [ 6] .rela.init        RELA            0000000000000000 000380 000018 18   I 20   5  8
  [ 7] .fini             PROGBITS        0000000000000000 0000a8 000008 00  AX  0   0  4
  [ 8] .note.GNU-stack   PROGBITS        0000000000000000 0000b0 000000 00      0   0  1
  [ 9] .debug_line       PROGBITS        0000000000000000 0000b0 00005b 00   C  0   0  8
  [10] .rela.debug_line  RELA            0000000000000000 000398 000090 18   I 20   9  8
  [11] .debug_line_str   PROGBITS        0000000000000000 00010b 000032 01  MS  0   0  1
  [12] .debug_info       PROGBITS        0000000000000000 00013d 000023 00      0   0  1
  [13] .rela.debug_info  RELA            0000000000000000 000428 000090 18   I 20  12  8
  [14] .debug_abbrev     PROGBITS        0000000000000000 000160 000012 00      0   0  1
  [15] .debug_aranges    PROGBITS        0000000000000000 000178 000032 00   C  0   0  8
[...]

PR #457, opened 2022-04-22
[ELF] handle DW_FORM_udata as an offset (#453)

llunak:


rui314:
Sorry, I already had the same change.

Issue #467, opened 2022-04-25
testsuite: ./test/elf/gdb-index-dwarf5.sh fails


marxin:
Since 6baef60a48bb72ba996ea0fde58801bffbd50270, all `gdb-index-dwarf*.sh` tests work except `gdb-index-dwarf5.sh`:

```
$ bash -x ./test/elf/gdb-index-dwarf5.sh
...
+ grep -Pq 'fn3 \(\) at .*/b.c:8' out/test/elf/gdb-index-dwarf5/log
+ grep -Pq 'fn2 \(\) at .*/a.c:4' out/test/elf/gdb-index-dwarf5/log
```

Where `log` looks like:
```
$ cat out/test/elf/gdb-index-dwarf5/log
Breakpoint 1 at 0x20114a: file <stdin>, line 4.
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".

Breakpoint 1, main () at <stdin>:4
4	<stdin>: No such file or directory.
Breakpoint 2 at 0x7ffff7fb7030 (2 locations)

Breakpoint 2, 0x00007ffff7fb7030 in trap@plt () from out/test/elf/gdb-index-dwarf5/e.so
#0  0x00007ffff7fb7030 in trap@plt () from out/test/elf/gdb-index-dwarf5/e.so
#1  0x00007ffff7fb71e3 in fn8 () at out/test/elf/gdb-index-dwarf5/d.c:6
#2  0x00007ffff7fb71f4 in fn7 () at out/test/elf/gdb-index-dwarf5/d.c:10
#3  0x00007ffff7fb71ab in fn6 () at out/test/elf/gdb-index-dwarf5/c.c:4
#4  0x00007ffff7fb71bc in fn5 () at out/test/elf/gdb-index-dwarf5/c.c:8
#5  0x00007ffff7fb7189 in fn4 () at out/test/elf/gdb-index-dwarf5/b.c:4
#6  0x00007ffff7fb719a in fn3 () at out/test/elf/gdb-index-dwarf5/b.c:8
#7  0x00007ffff7fb7167 in fn2 () from out/test/elf/gdb-index-dwarf5/e.so
#8  0x00007ffff7fb7178 in fn1 () from out/test/elf/gdb-index-dwarf5/e.so
#9  0x0000000000201154 in main () at <stdin>:4
A debugging session is active.

	Inferior 1 [process 1974] will be killed.

[...]

PR #470, opened 2022-04-26
Gdb index address size

llunak:
I see you pushed a fix for #467 just at the same time, but I think the other two commits are still useful.


rui314:
Could you rebase this so that my commit is not included in your commit list?

rui314:
Maybe it would have worked if I just hit the merge button. Thanks anyways!

PR #474, opened 2022-04-26
[ELF] handle also DW_FORM_addrx3 and DW_FORM_strx3

llunak:
Apparently these also may get used with large enough binaries, despite being somewhat tricky with regard to endianess.


rui314:
Thanks!

PR #476, opened 2022-04-27
[ELF] the --gdb-index code does not handle .debug_types

llunak:
And there is probably little need to do so, since it's DWARF4-only and requires explicit -fdebug-types-section, but at least detect and abort in this case.

I think that with this the .gdb_index support can be considered practically complete (unless somebody decides that support for either .debug_types or DWARF64 is worth the effort).

### DIFF
--- a/elf/input-files.cc
+++ b/elf/input-files.cc
@@ -234,6 +234,9 @@ void ObjectFile<E>::initialize_sections(Context<E> &ctx) {
           debug_pubtypes = isec;
           isec->is_alive = false;
         }
+
+        if (name == ".debug_types")
+          has_debug_types = true;
       }
 
       static Counter counter("regular_sections");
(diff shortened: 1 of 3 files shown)


rui314:
Handling this as a fatal error is probably too severe. I'd just warn on a presence of `.debug_types` if `--gdb-index` is given.

llunak:
GDB assumes that .debug_types will be included in the index if present, otherwise it'll say things like "'uniform' has unknown type; cast it to its declared type". IMO it's better to abort rather than create a partially broken index. The usual developer does not know what .debug_types actually is and may not realize GDB misbehaves because of the previous warning (and that's assuming the warning would even be always noticed).

rui314:
OK, let me merge this. I may make a change to turn this into warning and disable `--gdb-index` (instead of creating a broken `.gdb_index`) though.

llunak:
Fair enough.

PR #496, opened 2022-05-05
tests: print skipped if valgrind is missing

marxin:


rui314:
Is this necessary? Even if valgrind is not available, most tests in this file are not skipped.

marxin:
Yes, otherwise I get:

```
$ ./test/elf/mold-wrapper.sh
Testing mold-wrapper ... abuild@marxinbox:~/rpmbuild/BUILD/mold-1.999.999+git.20220505.ca25f1fd> echo $?
1
```

Which means the tests will fail.

rui314:
Ah, you are right. I'll submit a follow-up patch to suppress the "skipped" message after merging.

marxin:
Thanks, I can confirm it works fine for me now.

PR #498, opened 2022-05-09
system mimalloc: link statically

haampie:
`-lmimalloc` results in dynamic linking by default, but static linking is what we want.

Maybe it would be better to make `mold` a CMake project, given that it's already using `cmake` from `make` anyways, and two of its dependencies have cmake config files?

Edit, just found https://github.com/rui314/mold/pull/49#issuecomment-882649529, sounds good.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,7 @@ endif
 ifeq ($(USE_MIMALLOC), 1)
   ifdef SYSTEM_MIMALLOC
     MOLD_CXXFLAGS += -DUSE_SYSTEM_MIMALLOC
-    MOLD_LDFLAGS += -lmimalloc
+    MOLD_LDFLAGS += -Wl,-whole-archive -l:libmimalloc.a -Wl,-no-whole-archive
   else
     MIMALLOC_LIB = out/mimalloc/libmimalloc.a
     MOLD_CXXFLAGS += -Ithird-party/mimalloc/include


rui314:
Can you elaborate a bit on the motivation to statically link to the system-wide mimalloc library?

haampie:
Hm, maybe I got the docs wrong, but https://github.com/microsoft/mimalloc seems to suggest you can either LD_PRELOAD the shared library, or link statically:

> Static override
> 
> On Unix-like systems, you can also statically link with mimalloc to override the standard malloc interface. The recommended way is to link the final program with the mimalloc single object file (mimalloc-override.o). We use an object file instead of a library file as linkers give preference to that over archives to resolve symbols. To ensure that the standard malloc interface resolves to the mimalloc library, link it as the first object file. For example:
> 
> gcc -o myprogram mimalloc-override.o  myfile1.c ...
> 
> Another way to override statically that works on all platforms, is to link statically to mimalloc (as shown in the introduction) and include a header file in each source file that re-defines malloc etc. to mi_malloc. This is provided by [mimalloc-override.h](https://github.com/microsoft/mimalloc/blob/master/include/mimalloc-override.h). This only works reliably though if all sources are under your control or otherwise mixing of pointers from different heaps may occur!

haampie:
Basically my worry was that both `libmimalloc.so` and `libc.so` provide a `malloc` symbol, and I'm not sure if the runtime linker is guaranteed to use `libmimalloc.so`'s `malloc`. Static linking forces mimalloc's malloc.

But after some testing, it seems that the dynamic linker picks the `malloc` symbol from the first library in DT_NEEDED, so probably this PR can be closed. Sorry for the noise.

Issue #485, opened 2022-04-30
Do ICF to merge data if `--ignore-data-address-equality` is given


rui314:
Currently, mold's `--icf=all` does not merge data sections at all. This is because many programs depend on the pointer uniqueness for global variables.

LLVM lld has an option `--ignore-data-address-equality` to force the linker to merge data sections. We should implement it too.

Issue #484, opened 2022-04-30
Implement `--icf=safe`


rui314:
Identical Code Folding (ICF) is a powerful optimization to reduce the size of a linker output. It merges functions that happen to be compiled to the identical machine code that behave exactly the same.

The downside of doing this is the optimization per se violates the specification of the C/C++ language specs. In these languages, taking pointers of two distinctive functions must result in two non-equivalent pointer values. However, if we optimize two distinctive functions into a single function, that resulting two pointers will have the same value.

Currently, mold supports only `--icf=all`. That instructs the linker to do the optimization anyway.

Other linkers, notably LLVM lld, supports `--icf=safe`. It uses `.llvm_addrsig` sections to identify functions that are safe to merge. Functions that are not mentioned in the section are not address-taken (no one takes its pointer), so they are safe to merge.

We should support `--icf=safe` in the same manner as LLVM lld.

rui314:
`.llvm_addrsig` is currently an LLVM-specific section. If we implement the support of the section to mold, we should propose the same feature to gcc and possibly send a patch to gcc to implement it.

ishitatsuyuki:
I'll try to implement this next week.

rui314:
I filed a feature request to GCC to ask if GCC can support `.llvm_addrsig`. Let's see how it goes.

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105625

rui314:
@ishitatsuyuki Are you interested in adding a `.addrsig` directive support to the GNU assembler?

ishitatsuyuki:
> Are you interested in adding a .addrsig directive support to the GNU assembler?

Sure. Not really familiar with the GNU codebase, but I could certainly try that.

rui314:
Nice! Then please go ahead and send a patch to GNU binutils. I'm totally unfamiliar with binutils too. Let me reopen this issue to track this feature request on our side.

PR #516, opened 2022-05-15
ICF: explain, implement icf=safe and ignore-data-address-equality.

ishitatsuyuki:
Issue #484 and #485.

Statistics on llvm-project libclang.so.15 (@master):

<table>
<thead>
<tr>
	<td> Flags
	<td> Saved bytes
<tbody>
<tr>
	<td> --icf=all
	<td> 3530320
<tr>
	<td> --icf=safe
	<td> 3443898
<tr>
	<td> --icf=safe --ignore-data-address-equality
	<td> 4190222
</table>

LLVM test suite appears to pass.

```
Testing Time: 111.82s
  Skipped          :    39
  Unsupported      :  1637
  Passed           : 77112
  Expectedly Failed:   178
```

### DIFF
--- a/elf/icf.cc
+++ b/elf/icf.cc
@@ -40,17 +40,29 @@
 // regular trees are equal. Given this formulation, we want to find as
 // many identical vertices as possible.
 //
-// Solving such problem is computationally intensive, but mold is quite fast.
-// For Chromium, mold's ICF finishes in less than 1 second with 20 threads.
-// This is contrary to lld and gold, which take about 5 and 50 seconds to
-// run ICF under the same condition, respectively.
+// Just like a lot of problems with graph, this problem doesn't have a
+// straightforward "optimal" solution, and we need to resort to heuristics.
 //
-// mold's ICF is faster because we are using a better algorithm.
-// It's actually me who developed and implemented the lld's ICF algorithm,
-// and I can say that mold's algorithm is better than that in all aspects.
-// It scales better for number of available cores, require less overall
-// computation, and has a smaller working set. So, it's better with a single
-// thread and even better with multiple threads.
+// mold approaches this problem by hashing program trees with increasing depth
+// on each iteration.
+// For example, when we start, we only hash individual functions with
+// their call into other functions omitted. From the second iteration, we
+// put the function they call into the hash by appending the hash of those
+// functions from the previous iteration. This means that the nth iteration
+// hashes call chain up to (n-1) levels deep.
+// We use a cryptographic hash function, so the unique number of hashes will
+// only monotonically increase as we take into account of deeper trees with
+// iterations (otherwise, that means we have found a hash collision). We stop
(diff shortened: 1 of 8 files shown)


elf/input-files.cc
```diff
@@ -740,6 +746,35 @@ void ObjectFile<E>::register_section_pieces(Context<E> &ctx) {
...
 
+template <typename E>
+void ObjectFile<E>::fill_addrsig(Context<E> &ctx) {
+  if (llvm_addrsig) {
+    const u8 *start = reinterpret_cast<const u8 *>(llvm_addrsig->contents.data());
+    const u8 *end = start + llvm_addrsig->contents.size();
+    const u8 *cur = start;
+    while (cur != end) {
+      u64 symIndex = read_uleb(cur);
+      if (this->symbols[symIndex]->file != this) continue;
```
ishitatsuyuki, line 757: I added checks so it will prevent any potential race condition, but I didn't manage to understand why a symbol could be from another file's section. What are the cases where this happen?

elf/icf.cc
```diff
@@ -106,12 +118,12 @@ static void uniquify_cies(Context<E> &ctx) {
 }
 
 template <typename E>
-static bool is_eligible(InputSection<E> &isec) {
+static bool is_eligible(Context<E> &ctx, InputSection<E> &isec) {
```
ishitatsuyuki, line 121: For now I've adopted the LLD behavior where addr_sig sections does not participate in ICF at all -- though I wonder if it's legal to merge other non-addr_sig sections into one of the addr_sig section.

rui314: Are you asking if it is a correct thing to consider all sections in an object file as if they were address-taken if the object file does not have a `.llvm_addrsig` section? If so, I think the answer is yes. That's a conservative assumption required for correctness.

rui314 (approved):
Overall looking good, but it needs a testcase.

elf/icf.cc
ishitatsuyuki: That’s not what I meant — sorry it wasn’t clear enough.

What I tried to say is to merge a section that doesn’t get its address taken into another section that has its address taken. It sounds safe because only the latter section can observe the “address”. But this merging currently doesn’t happen right now because address significant regions are completely removed from the ICF graph.

rui314: I think we cannot merge two address-taken functions, but we can merge a non-address-taken function with an address-taken function. So the lld's behavior seems too conservative.

ishitatsuyuki: I see. I'll try to measure if there's a significant difference. If there is, we could implement the more aggressive behavior.

rui314: Thanks. If it will make a difference, please report it to lld (https://github.com/llvm/llvm-project/issues) with a reference to your patch so that they can fix the issue too.

PR #522, opened 2022-05-20
-no-keep-memory breaks mold, passthrough

jacob-abraham:
existing build systems using -no-keep-memory break when run with mold
PR just adds a pass-through to the cmdline for elf for this


rui314:
Thanks!

Issue #335, opened 2022-02-09
Uncaught exception running mold on macos


andrewdavidmackenzie:
I have installed mold on macos (Monterey 12.0.1) using homebrew:

`uname -a`
`Darwin MacBook-Pro 21.1.0 Darwin Kernel Version 21.1.0: Wed Oct 13 17:33:23 PDT 2021; root:xnu-8019.41.5~1/RELEASE_X86_64 x86_64
`

`which mold`
`/usr/local/bin/mold`

`mold -v`
`mold 1.0.3 (compatible with GNU ld)`

When I try to use it, it fails with a broken symlink error, and there seems to be a reference to a linux /proc folder.

```
mold -run cargo build              
libc++abi: terminating with uncaught exception of type std::__1::__fs::filesystem::filesystem_error: filesystem error: in read_symlink: No such file or directory [/proc/self/exe]
zsh: abort      mold -run cargo build
````

**Comment by LoganDark** [comment]
Fairly sure `mold` does not support true Unix systems, only Linux. Sorry :(

**Comment by andrewdavidmackenzie** [comment]
Ok, thanks
I didn't see anything either way about macos, and when I saw homebrew I assumed macos.

**Comment by rui314** [comment]
@Alcaro That's a good suggestion. Let me do that.

**Comment by ShawnZhong** [comment]
For `mold -run` intercepting `exec` on `ld`, `ld.gold` and `ld.lld`, macOS does not support `LD_PRELOAD` and the alternative `DYLD_INSERT_LIBRARIES` does not work with System Integrity Protection enabled: 

```
> DYLD_INSERT_LIBRARIES=./mold-wrapper.so env | grep DYLD  # empty
```

I guess we need an alternative way for `mold -run` to work on macOS. Say creating symbolic link `ld -> mold` under a directory and add it to `PATH`. But that won't work when the absolute path of a linker is used.

andrewdavidmackenzie:
This caught my eye this week:

https://www.reddit.com/r/rust/comments/wd8xuc/mold_on_macos_works_kinda_and_i_confirmed_with_a/?utm_medium=android_app&utm_source=share

PR #537, opened 2022-06-09
fix `mold -run ...` on macos

zyctree:
it will fix https://github.com/rui314/mold/issues/335


rui314:
You may be able to suppress the error message on macOS with this patch, but this patch doesn't make it work. Subcommands will still use /usr/bin/ld instead of mold. It's because LD_PRELOAD doesn't work on macOS in the first place. So this is not a correct fix.

rui314:
I submitted 8f02509ba6528354cfe6660bb56512af38a80f87 instead.

PR #538, opened 2022-06-09
Better reporting of undefined symbols

llunak:
This makes mold's reporting of undefined symbols mostly match lld output (i.e. grouped by symbol, listing the function that referenced it, limiting the size of the output, etc.). One thing that is notably missing is using debuginfo to report the exact source location, which would require interpreting .debug_line .


llunak:
Never mind that libdwarf part, I find its API and documentation so bad that it's simpler for me to understand the lowlevel details from the dwarf spec itself and it'd be simpler to implement it directly.

rui314:
I've just merged your commits. Thanks!

PR #540, opened 2022-06-12
Implement `--(no-)allow-shlib-undefined.

### LINKED ISSUE #526: Support for -Wl,--allow-shlib-undefined ?
Hello,

Is support for this option planned ? 
It is necessary in my case to build a plug-in against a proprietary executable software which exports some symbols dynamically, to give one use case.

Thanks !

rbartlensky:
I believe this fixes #526. 

I hope this covers all cases. If not, let me know and I can look into it some more!

### DIFF
--- a/elf/cmdline.cc
+++ b/elf/cmdline.cc
@@ -54,6 +54,8 @@ Options:
   --Tdata                     Set address to .data
   --Ttext                     Set address to .text
   --allow-multiple-definition Allow multiple definitions
+  --allow-shlib-undefined     Allow undefined symbols in shared libraries (default for --shared)
+    --no-allow-shlib-undefined
   --as-needed                 Only set DT_NEEDED if used
     --no-as-needed
   --build-id [none,md5,sha1,sha256,uuid,HEXSTRING]
@@ -932,7 +934,9 @@ std::vector<std::string> parse_nonpositional_args(Context<E> &ctx) {
     } else if (read_flag("disable-new-dtags")) {
     } else if (read_flag("nostdlib")) {
     } else if (read_flag("allow-shlib-undefined")) {
+      ctx.arg.allow_shlib_undefined = true;
     } else if (read_flag("no-allow-shlib-undefined")) {
+      ctx.arg.allow_shlib_undefined = false;
     } else if (read_flag("no-add-needed")) {
     } else if (read_flag("no-call-graph-profile-sort")) {
     } else if (read_flag("no-copy-dt-needed-entries")) {
@@ -1068,6 +1072,10 @@ std::vector<std::string> parse_nonpositional_args(Context<E> &ctx) {
   // object file.
   if (ctx.arg.shared)
     ctx.overwrite_output_file = false;
+
+  // By default, this is set to true when building shared objects
+  if (!ctx.arg.allow_shlib_undefined)
+    ctx.arg.allow_shlib_undefined = ctx.arg.shared;
 
   if (version_shown && remaining.empty())
     exit(0);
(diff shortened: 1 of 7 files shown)


rbartlensky:
Force pushed an amend to my commit which should fix CI.

rbartlensky:
Hmm, running `make CC=clang CXX=clang++ -j$(nproc) test` works for me locally. Will probably need to set up a docker image or install the same version of clang to see exactly what's going on :thinking:

rui314:
The failing tests are AddressSanitizer and ThreadSanitizer tests, so it implies that your patch has a memory safety and thread safety issue, like dereferencing an invalid pointer or accessing a free'd memory. Please try running tests with `make test-asan` or `make test-tsan`.

rbartlensky:
@rui314 it looks like `test` is failing though. In any case I ran the two commands you mentioned and they both pass for me. Could it be a ccache issue by any chance?

rui314:
I don't think it is ccache-related. Can't you run the GitHub CI on your fork of this repo?

ishitatsuyuki:
Looks like 792773be20b558a07295bb33875baded1a92b9ca broke merge compatibility for this, btw.

Tested this locally on refs/pull/540/head (before merge) and I can't reproduce the failure either. @rbartlensky Perhaps we should get this rebased first and see if there's any issue due to changes in master.

rbartlensky:
Rebased + confirmed that the tests pass locally

rui314:
I cannot merge this until tests are all green on GitHub Actions. This situation happens occasionally to me too, as my local development environment isn't exactly the same as the one on GitHub Actions. Feel free to continue pushing new commits until tests are all green.

llunak:
The added test is missing -fPIC for some of the -shared compilations, which might have something to do with the test failing. Locally I even get "Testing shlib-undefined ... /usr/lib64/gcc/x86_64-suse-linux/7/../../../../x86_64-suse-linux/bin/ld: error: /tmp/ccH1v1Ip.o: requires dynamic R_X86_64_PC32 reloc against 'fn1' which may overflow at runtime; recompile with -fPIC".

rbartlensky:
@rui314 thank you for your patience! Forks unfortunately don't allow you to run CI yourself, I guess I would need to "manually" fork the project.

@llunak Great catch! I also found a small problem with one of the tests (I wasn't handling 1 of the 4 cases). 

Let's see if CI is happy now :)

rui314:
@rbartlensky I wonder if you can clone this repo to your account and run the CI there instead of here. It looks like just pushing the "Fork" button didn't allow you to do that, but if you manually copy this repo to your account, doesn't it work?

rbartlensky:
@rui314 ok I did just that. I have a "mirror" instead of a fork, so I'm just pushing stuff here: https://github.com/rbartlensky/mold-mirror/pull/1. Once I get CI to pass, I will update this branch and ping you. Too bad forks don't allow you to run CI :/

rui314:
@rbartlensky `$CC` in tests is always called with `-B.`, so `$CC` tries to find `ld` from the current directory. Since the current directory is the top directory of mold, and there should be a symlink from `ld` to `mold` there, so mold will be invoked instead of `/usr/bin/ld`.

rbartlensky:
@rui314 I'm back again, this time with good news. First of all, thank you so much for your help -- I was silly for missing `-B.`, but now I learned! CI is passing in my [mirror](https://github.com/rbartlensky/mold-mirror/pull/1), so I hope it will pass here as well :)

rui314:
Merged. Thanks!

PR #543, opened 2022-06-17
(WIP) use debuginfo for undefined references

llunak:
This is an unfinished work that tries to use debuginfo to report source locations of places that cause undefined references, rather than just the object file. The reason I'm submitting this unfinished version is that debuginfo requires relocating to be usable, otherwise references between sections do not work (relocations include also offsets within the section, and without relocations those values are 0, pointing to incorrect places in the debuginfo data in the .o file).

The debuginfo relocations need to be done somewhere, and currently it's done while writing the entire output file, so for now the patch delays reporting the undefined symbols until after that, since currently molds always processes the entire file even if there are errors. But you reverted the delayed processing of undefined references, and I don't know if it's possible to process debuginfo and its relocations before non-delayed undefined references handling. This changeset reverts the revert for simplicity, but I don't want to change the code back to delayed if you'd change it back yet again, so I wanted to ask for feedback on this.

### DIFF
--- a/elf/arch-arm32.cc
+++ b/elf/arch-arm32.cc
@@ -316,7 +316,7 @@ void InputSection<E>::apply_reloc_nonalloc(Context<E> &ctx, u8 *base) {
     u8 *loc = base + rel.r_offset;
 
     if (!sym.file) {
-      record_undef_error(ctx, rel);
+      add_undef(ctx, file, sym, shndx, &rel);
       continue;
     }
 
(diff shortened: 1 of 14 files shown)


rui314:
I honestly feel that this is a bit too much code and complexity just to show the line number for an undefined symbol reference. After all, we just want to know the line number, but unfortunately DWARF is too complicated that that doesn't allow us to do that easily. It is usually easy for me to identify a location of an undefined reference without a line number, so I didn't personally find it that useful.

PR #544, opened 2022-06-17
Also copy mold-wrapper.so from docker build.

lancethepants:


rui314:
Can you read https://github.com/rui314/mold/blob/main/CONTRIBUTING.md and add a `Signed-off-by` line to your commit message?

lancethepants:
Done. Thanks for looking at this.

PR #557, opened 2022-06-26
Don't use grep -P in tests

wangp:
The version of grep being used may not support the -P flag
(Perl-compatible regular expressions).


rui314:
Can we replace `-P` with `-E` instead?

wangp:
If you prefer. But `grep -E` won't recognise `\d` so `\d` still needs to be replaced by either `[0-9]` or `[[:digit:]]`.

rui314:
Thanks!

PR #569, opened 2022-07-03
Fix more compiler warnings

sicherha:
In the long run, I would like to enable `-Wall -Wextra` by default to catch such issues in advance. Unfortunately, we're not there yet - among other things, GCC gives a bunch of warnings about calling `memset()` to initialise a non-trivial object type.

### DIFF
--- a/cmdline.h
+++ b/cmdline.h
@@ -61,7 +61,7 @@ std::vector<std::string_view> expand_response_files(C &ctx, char **argv) {
   return vec;
 }
 
-static std::string_view string_trim(std::string_view str) {
+static inline std::string_view string_trim(std::string_view str) {
   size_t pos = str.find_first_not_of(" \t");
   if (pos == str.npos)
     return "";
--- a/mold.h
+++ b/mold.h
@@ -311,7 +311,7 @@ inline u64 read_uleb(u8 const*&buf) {
 }
 
 inline i64 uleb_size(u64 val) {
-#pragma unroll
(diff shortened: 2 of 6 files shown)


mold.h
```diff
@@ -311,7 +311,7 @@ inline u64 read_uleb(u8 const*&buf) {
 }
 
 inline i64 uleb_size(u64 val) {
-#pragma unroll
+#pragma GCC unroll 8
```
rui314, line 314: This is a valid pragma for Clang.

https://clang.llvm.org/docs/AttributeReference.html#pragma-unroll-pragma-nounroll

cmdline.h
```diff
@@ -61,7 +61,7 @@ std::vector<std::string_view> expand_response_files(C &ctx, char **argv) {
   return vec;
 }
 
-static std::string_view string_trim(std::string_view str) {
+[[maybe_unused]] static std::string_view string_trim(std::string_view str) {
```
rui314, line 64: Instead of adding an attribute, can you change this to `static inline`?

sicherha: Done.

mold.h
sicherha: `#pragma GCC unroll <n>` is supported by both GCC and Clang, whereas `#pragma unroll` is only supported by Clang. Using the GCC-compatible notation fixes an *unknown pragma* warning when compiling with GCC.

rui314:
Thanks!

Issue #190, opened 2021-12-23
Support Windows


rui314:
We want to support the Windows target by mold 3.0.

tearosccebe:
I want this!!!

r0nsha:
Can't wait for this!

a97077088:
+1

vrubleg:
Guys, stop spamming. All your useless comments are sent to all subscribed people. Just use reactions on the first message.

a97077088:
ok ,i'm sorry

3rav:
In the case of the msys2/mingw environment currently there is this error: 

```
In file included from filepath.cc:1:
mold.h:17:10: fatal error: sys/mman.h: No such file or directory
   17 | #include <sys/mman.h>
```

yuyoyuppe:
Is it possible to help on this issue somehow? E.g. port to a Cmake, make some logic platform-independent etc.

aadog:
Then, please sponsor with donations

aadog:
Please initiate a sponsorship link

rui314:
I think a bulleted list of features or an issue tracker-based task management work after I set the foundation of mold/Windows. The core part has to be written by me (or by someone who is good at writing optimized programs and has extremely deep knowledge of the Windows linker) to achieve our performance goal. So what I can do right now is to try to finish up mold/macOS as quickly as possible so that I can start working on mold/Windows.

aadog:
If the windows version of gold is implemented, can go/cgo achieve a link speed of %100++?

rui314:
I don't want to speculate about speed because mold/Windows is vaporware at the moment. I'll create mold/Windows because I see some opportunity there, and I can only say that I'll do my best.

X547:
PE executables are used not only in Windows, but also in EFI (so many operating systems include UNIX-like can use it for boot loader), [SanOS](http://www.jbox.dk/sanos/), BeOS R4 x86 and others.

justanotheranonymoususer:
Any news about this? Currently, that's the only mold I have on my windows;

![image](https://user-images.githubusercontent.com/5781692/192221977-5341624b-0ec5-4d2c-8b24-ca50da1c0bb5.png)

rui314:
I will update this issue when I have an update, so do not ping us.

369BG963:
Do we have any news about MOLD Coming to Windows?

rui314:
[...]

PR #571, opened 2022-07-05
CMake Port

clemenswasser:
For supporting Windows(#190), we definitely need something more portable than Makefiles. There are currently many Build-systems promising Cross-platform support and CMake is one of them. I would argue, that with mold's `third-party` modules TBB and mimalloc already using CMake, it is a natural choice. This PR ports mold to CMake. I have implemented every functionality(build, install, testing) from the current Makefiles.


rui314:
This generally seems good. I agree with you that CMake is probably the build system we should migrate to, as we have already been using it for the third-party packages. I don't want to integrate this until mold 2.0 though, as our Makefile is pretty stable right now and we don't have a need to change it immediately.

clemenswasser:
Agree with both of you. Thanks for your quick feedback 👍

CMakeLists.txt
```diff
@@ -0,0 +1,225 @@
...
+    find_package(ZLIB REQUIRED)
+    find_library(M_LIBRARIES m REQUIRED)
+    target_link_libraries(mold PRIVATE
+        ZLIB::ZLIB
+        ${M_LIBRARIES})
+endif()
+
+target_link_libraries(mold PRIVATE
+    TBB::tbb
+    $<$<BOOL:MOLD_USE_MIMALLOC>:mimalloc::mimalloc>)
```
friendlyanon, line 96: This doesn't need to be a genex.

CMakeLists.txt
```diff
@@ -0,0 +1,225 @@
...
+
+if(CMAKE_SYSTEM_NAME MATCHES Linux)
```
friendlyanon, line 98: OpenSSL is used for SHA256 everywhere but Apple.

CMakeLists.txt
```diff
@@ -0,0 +1,225 @@
...
+
+file(STRINGS .git/HEAD GIT_HASH)
+
+if(GIT_HASH MATCHES "ref: ")
+    string(REPLACE "ref: " "" GIT_HASH "${GIT_HASH}")
+    file(STRINGS .git/${GIT_HASH} GIT_HASH)
+endif()
+
+target_compile_definitions(mold PRIVATE
+    GIT_HASH="${GIT_HASH}"
```
friendlyanon, line 117: This will cause nodes in the build graph to be stale whenever the git hash changes. This leads to unnecessary compiles when building incrementally.

@rui314 This hash thing should really be done in a way that makes the binary deterministic and not dependent on git. I played with this locally and you can add a placeholder in the binary like so:

```cpp
char const git_hash_buffer[] = "\x00" "patch git hash here patch git hash here";
auto const git_hash = std::string_view(git_hash_buffer);
```

After that, the resulting binary is reproducible and can be patched before packaging/distribution:

```sh
printf %s "$git_hash" | dd of=mold bs=1 count=40 conv=notrunc "seek=$(grep -aboEe '.patch git hash here patch git hash here' | cut -f1 -d:)"
```

And because the first byte is a null byte, you can even adjust accordingly at runtime.

CMakeLists.txt
```diff
@@ -0,0 +1,225 @@
...
+    COMMAND ${CMAKE_COMMAND} -E create_symlink mold ld
+    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+    COMMENT "Creating ld symlink"
+    VERBATIM)
+add_custom_command(
+    TARGET mold POST_BUILD
+    COMMAND ${CMAKE_COMMAND} -E create_symlink mold ld64
+    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+    COMMENT "Creating ld64 symlink"
+    VERBATIM)
```
friendlyanon, line 177: Symlinks aren't quite portable. They also need admin privs on Windows.

CMakeLists.txt
```diff
@@ -0,0 +1,225 @@
...
+set(MOLD_TESTS_PATH ${CMAKE_CURRENT_LIST_DIR}/test/)
+
+if(CMAKE_SYSTEM_NAME MATCHES Darwin)
+    string(APPEND MOLD_TESTS_PATH "macho")
+else()
+    string(APPEND MOLD_TESTS_PATH "elf")
+endif()
+
+string(APPEND MOLD_TESTS_PATH "/*")
+FILE(GLOB MOLD_TESTS ${MOLD_TESTS_PATH})
```
friendlyanon, line 190: Do not glob source files for targets.

CMakeLists.txt
```diff
@@ -0,0 +1,225 @@
...
+foreach(MOLD_TEST IN LISTS MOLD_TESTS)
+    add_test(NAME ${MOLD_TEST} COMMAND ${MOLD_TEST})
+endforeach()
+
+# FIXME(cwasser): This is just a convenience ctest target,
+# which ensures that mold is build. This should ideally be fixed in CMake.
+include(ProcessorCount)
+ProcessorCount(MOLD_NPROC)
+add_custom_target(check COMMAND ${CMAKE_CTEST_COMMAND} -j${MOLD_NPROC}
+    DEPENDS mold USES_TERMINAL)
```
friendlyanon, line 201: "fixed in CMake"? CMake needs no fixing in this regard, you just need to let go of Linux-isms. You can use presets to automate testing to the point of needing to only run `ctest --preset=dev`.

The `DEPENDS` argument is also not how things work https://cmake.org/cmake/help/latest/command/add_custom_target.html#:~:text=at%20build%20time.-,DEPENDS,-Reference%20files%20and

CMakeLists.txt
```diff
@@ -0,0 +1,225 @@
...
+
+install(TARGETS mold DESTINATION ${CMAKE_INSTALL_BINDIR})
```
friendlyanon, line 203:
```suggestion
install(TARGETS mold)
```

CMake >= 3.14 will fill the destinations properly if omitted for a few groups, `RUNTIME` being one of those.

CMakeLists.txt
```diff
@@ -0,0 +1,225 @@
...
+install(TARGETS mold-wrapper DESTINATION lib/mold)
```
friendlyanon, line 204: Don't hardcode the destination, use the standard and cross-platform compliant locations from `GNUInstallDirs`.

CMakeLists.txt
```diff
@@ -0,0 +1,225 @@
...
+        \$\{REL_SYMLINK\}\
+        \$\{INSTALL_PREFIX\}/${CMAKE_INSTALL_LIBEXECDIR}/mold/ld)\n\
+    execute_process(\
+        COMMAND ${CMAKE_COMMAND} -E create_symlink\
+        mold\
+        \$\{INSTALL_PREFIX\}/${CMAKE_INSTALL_BINDIR}/ld.mold)\n\
+    execute_process(\
+        COMMAND ${CMAKE_COMMAND} -E create_symlink\
+        mold\
+        \$\{INSTALL_PREFIX\}/${CMAKE_INSTALL_BINDIR}/ld64.mold)")
```
friendlyanon, line 225: Only `$` needs escaping, but see above. Symlinks aren't portable. This is also a packaging concern that should be done using CPack.

friendlyanon (changes requested):
The CMake code needs many necessary changes to be easily usable. The main concerns are:

* Wrong defaults wrt vendored dependencies.
* Hardcoding compiler and platform specific flags.
* Symlinks.
* Breaking cross-platform conventions, despite quoting Windows support as the reason for this PR.

Please do ask away if you have any further questions.

CMakeLists.txt
```diff
@@ -0,0 +1,225 @@
...
+
+include(CMakeDependentOption)
+include(GNUInstallDirs)
+
+# Mold build options
+cmake_dependent_option(MOLD_USE_MIMALLOC "Use mimalloc" ON
+    "NOT CMAKE_SYSTEM_NAME MATCHES Darwin; NOT Android" OFF)
+option(MOLD_USE_SYSTEM_MIMALLOC "Use system mimalloc")
+option(MOLD_USE_SYSTEM_TBB "Use system TBB")
+option(MOLD_USE_LTO "Use Link Time Optimization")
```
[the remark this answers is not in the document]
clemenswasser: I tried to stick as much as possible to the existing Makefile and ported the variable over (as for the reasoning behind the flag). But your totally right, the user can just define `CMAKE_INTERPROCEDURAL_OPTIMIZATION` themselves, we should however document and highlight the variable in the build instructions.

CMakeLists.txt
```diff
@@ -0,0 +1,225 @@
...
+option(CMAKE_INSTALL_LOCAL_ONLY ON)
```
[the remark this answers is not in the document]
clemenswasser: I needed to set this at some point, before I knew about `add_subdirectory` with `EXCLUDE_FROM_ALL` existed. I can probably remove this, thanks!

CMakeLists.txt
```diff
@@ -0,0 +1,225 @@
...
+project(mold VERSION 1.3.0)
+
+include(CMakeDependentOption)
+include(GNUInstallDirs)
+
+# Mold build options
+cmake_dependent_option(MOLD_USE_MIMALLOC "Use mimalloc" ON
+    "NOT CMAKE_SYSTEM_NAME MATCHES Darwin; NOT Android" OFF)
+option(MOLD_USE_SYSTEM_MIMALLOC "Use system mimalloc")
+option(MOLD_USE_SYSTEM_TBB "Use system TBB")
```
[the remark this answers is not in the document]
clemenswasser: In the existing Makefile, using the system mimalloc or TBB is also opt-in. Why should I change this semantic?

CMakeLists.txt
```diff
@@ -0,0 +1,225 @@
...
+        function(mold_add_mimalloc)
+            string(APPEND CMAKE_C_FLAGS -DMI_USE_ENVIRON=0)
+            option(MI_BUILD_TESTS "Build test executables" OFF)
+            add_subdirectory(third-party/mimalloc EXCLUDE_FROM_ALL)
+        endfunction()
+
+        mold_add_mimalloc()
+    endif()
+
+    add_library(mimalloc::mimalloc ALIAS mimalloc)
```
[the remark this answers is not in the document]
clemenswasser: I know changing the `CMAKE_C_FLAGS` is bad. But I just tried to port the existing Makefile as close as possible.
Also, what this the advantage of this `Findmimalloc.cmake` script?

CMakeLists.txt
```diff
@@ -0,0 +1,225 @@
...
+    CXX_EXTENSIONS OFF
+    INTERPROCEDURAL_OPTIMIZATION ${MOLD_USE_LTO})
+target_include_directories(mold PRIVATE third-party)
+
+if(NOT MSVC)
+    target_compile_options(mold PRIVATE
+        $<$<PLATFORM_ID:Android>:-Wno-c++11-narrowing>
+        -fno-exceptions
+        -fno-unwind-tables
+        -fno-asynchronous-unwind-tables)
```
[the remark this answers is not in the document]
clemenswasser: But then nobody will do this, who wants to read the build guide anyway :wink:
On a more serious note, these flags are already passed by the existing Makefile and probably affect the performance.

clemenswasser:
> * Breaking cross-platform conventions, despite quoting Windows support as the reason for this PR.

@friendlyanon 
The intention of this PR is **not** to support Windows. It is just one building block to get there.
This PR is more of an initial port of the existing Makefiles.

CMakeLists.txt
clemenswasser: Thanks!

CMakeLists.txt
clemenswasser: Do you mean using `CMAKE_INSTALL_LIBDIR` instead of `lib/`?

CMakeLists.txt
clemenswasser: Yeah I'm not happy about this install code either. It's just the most direct port of the existing Makefile installation step I could think of. Is there any better option how I could do this?

CMakeLists.txt
clemenswasser: I wouldn't call this Linux-isms. When I run `cargo test` on one of my Rust projects, it ensures that the binaries, required for testing, are actually build first and not fail with an obscure error about missing files...

CMakeLists.txt
clemenswasser: Thanks, the Makefile does the same, good catch!

clemenswasser:
Rebased onto the latest main. @rui314 Sorry I think I accidentally reverted your rebase.

CMakeLists.txt
```diff
@@ -0,0 +1,225 @@
...
+option(MOLD_USE_SYSTEM_MIMALLOC "Use system mimalloc")
+option(MOLD_USE_SYSTEM_TBB "Use system TBB")
+option(MOLD_USE_LTO "Use Link Time Optimization")
+option(CMAKE_INSTALL_LOCAL_ONLY ON)
+
+# Default to `Release` build type,
+# to be consistent with the current Makefiles
+if(NOT CMAKE_BUILD_TYPE)
+    set(CMAKE_BUILD_TYPE Release CACHE STRING "" FORCE)
+endif()
```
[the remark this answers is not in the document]
friendlyanon: That still only handles single-config generators and after the `project()` call the variables are already set anyway. There isn't really a way to do this properly other than recommending things in the documentation.

CMakeLists.txt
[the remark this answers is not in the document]
friendlyanon: Using system dependencies is always preferable for package managers and they are the ones who distribute the software. Doing the right thing also happens to create less work for them.

CMakeLists.txt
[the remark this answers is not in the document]
friendlyanon: mimalloc already supports clients using CMake, because the project exports a CMake package. That should be used preferably. The approach with the find module makes vendoring opt-in, but the project code just has the idiomatic `find_package(mimalloc REQUIRED)` call that one would expect regardless. The linked example project illustrates this.

Interesting, I opened the github.dev page for mimalloc and project wide search didn't return anything at first for that macro. In this case you can just use `add_compile_definitions(MI_USE_ENVIRON=0)` instead.

CMakeLists.txt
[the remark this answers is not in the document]
friendlyanon: You are right, but project code is no place to hardcode flags regardless. Project code must only concern itself with build and usage requirements.

CMakeLists.txt
```diff
@@ -0,0 +1,225 @@
...
+
+if(NOT MSVC)
+    target_compile_options(mold PRIVATE
+        $<$<PLATFORM_ID:Android>:-Wno-c++11-narrowing>
+        -fno-exceptions
+        -fno-unwind-tables
+        -fno-asynchronous-unwind-tables)
+
+    find_package(ZLIB REQUIRED)
+    find_library(M_LIBRARIES m REQUIRED)
```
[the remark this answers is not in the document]
friendlyanon: Just because libm exists, doesn't mean it's necessary to be used. The check example determines exactly whether it's needed or not and you can act accordingly.

CMakeLists.txt
friendlyanon: Ninja and the Makefiles generators already have the `test` target, while other generators have `RUN_TESTS`. This target is created when `enable_testing()` or `include(CTest)` is used to enable testing. This target depends on "all" and makes sure everything is built.

Is this what you mean?

CMakeLists.txt
friendlyanon: `CMAKE_INSTALL_LIBDIR` uses the correct lib directory on multi-arch systems.

CMakeLists.txt
friendlyanon: I don't know the actual requirements this would satisfy, so I cannot say with certainty. Isn't this only needed for GCC?

friendlyanon:
> The intention of this PR is not to support Windows. It is just one building block to get there.

Sure, but why do extra work now that later has to be undone anyway, especially now that you know how to do that.  
The point of CMake is that you can write (mostly) declarative code that works on all platforms by default.

CMakeLists.txt
friendlyanon: Hmm okay, the target doesn't actually depend on the "all" target. The `check` target could still use improvement:

```suggestion
include(ProcessorCount)
ProcessorCount(MOLD_NPROC)
add_custom_target(
    check
    COMMAND "${CMAKE_COMMAND}" --build "${mold_BINARY_DIR}" --config "\$<CONFIG>" -j "${MOLD_NPROC}"
    COMMAND "${CMAKE_CTEST_COMMAND}" -C "\$<CONFIG>" -j "${MOLD_NPROC}"
    VERBATIM
    USES_TERMINAL
)
```

CMakeLists.txt
clemenswasser: I have listed all files manually now and ported the test-arch make target (and also extracted them into the test folder). Except for the uninstall target, this port should now be at least on feature parity.

CMakeLists.txt
clemenswasser: Thanks!

CMakeLists.txt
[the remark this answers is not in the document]
clemenswasser: mhm `add_compile_definitions` does the the define for the current directory and below, so it also sets this for the mold target...

CMakeLists.txt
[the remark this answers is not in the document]
clemenswasser: I've removed this. I am however still not sure if this is the best behavior. Having no feedback at all, when forgetting to set the `BUILD_TYPE`, just seems strange to me...

CMakeLists.txt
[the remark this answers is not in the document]
clemenswasser: I've changed this. For me it picks up the packages from my system :+1:

CMakeLists.txt
```diff
@@ -0,0 +1,225 @@
...
+        add_subdirectory(third-party/tbb EXCLUDE_FROM_ALL)
+    endfunction()
+
+    mold_add_tbb()
+endif()
+
+if(MOLD_USE_LTO)
+    include(CheckIPOSupported)
+    check_ipo_supported()
+endif()
```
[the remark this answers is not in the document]
clemenswasser: I was a bit confused by this page: https://cmake.org/cmake/help/latest/policy/CMP0069.html#examples.
Especially the last example:

> Apply IPO flags without any checks. This may lead to build errors if IPO is not supported by the compiler in the current environment. Produce an error if CMake does not know IPO flags for the current compiler

 I thought that setting `INTERPROCEDURAL_OPTIMIZATION` without `check_ipo_supported` would be silently ignored if the current compiler doesn't supports it. This is however not the case since CMake 3.9, as noted by the policy. Thanks, good catch!

CMakeLists.txt
```diff
@@ -0,0 +1,225 @@
...
+
+include(CheckPIESupported)
+check_pie_supported()
+
+add_library(mold-wrapper SHARED)
+set_target_properties(mold-wrapper PROPERTIES
+    POSITION_INDEPENDENT_CODE ON
+    PREFIX "" # Remove the default `lib` prefix
+    INTERPROCEDURAL_OPTIMIZATION ${MOLD_USE_LTO})
```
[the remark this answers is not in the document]
clemenswasser: Done

CMakeLists.txt
```diff
@@ -0,0 +1,225 @@
...
+target_link_libraries(mold-wrapper PRIVATE ${CMAKE_DL_LIBS})
+target_sources(mold-wrapper PRIVATE elf/mold-wrapper.c)
+
+add_executable(mold)
+target_compile_features(mold PRIVATE cxx_std_20)
+set_target_properties(mold PROPERTIES
+    CXX_EXTENSIONS OFF
```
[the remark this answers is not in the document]
clemenswasser: But shouldn't this always be set, otherwise someone might use a different preset and accidentally use extensions?

CMakeLists.txt
```diff
@@ -0,0 +1,225 @@
...
+    INTERPROCEDURAL_OPTIMIZATION ${MOLD_USE_LTO})
+target_include_directories(mold PRIVATE third-party)
```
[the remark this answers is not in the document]
clemenswasser: Done

CMakeLists.txt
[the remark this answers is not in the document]
clemenswasser: Could using something like `check_cxx_compiler_flag` be an alternative. Is this an alternative you could recommend?

CMakeLists.txt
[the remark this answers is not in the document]
clemenswasser: Yes, you are right. Another question, is there any better alternative to `target_link_libraries` when it comes to linking a library like libm, which we know is not a CMake target. I never know if `target_link_libraries` is really the best solution in that case, as it might prefer a CMake target with the same name instead...

CMakeLists.txt
clemenswasser: I've integrated this target and it works great, thanks! :smile:

CMakeLists.txt
clemenswasser: As I've said CMake and Make are not consistent (at least for me):
`CMAKE_INSTALL_LIBDIR`: `/lib64`
`LIBDIR` (Makefile): `/lib`
And I don't want, that the mold installation destinations change...

CMakeLists.txt
clemenswasser: > This is also a packaging concern that should be done using CPack.

How would I do this using CPack?

clemenswasser:
FYI @rui314
I have now also run the `./install-build-deps.sh` script and all tests, running on the host machine, pass :smile:.
Running `ctest -R .*_host$` yields: `100% tests passed, 0 tests failed out of 250`.
Now the CMake build system is on feature parity with the Makefiles (I still have to figure out how to run the arch test though)
Also, CTest is even more capable than the current Make testing, parallel testing, test selection, etc. :wink:.

rui314:
I made a few changes to the makefiles and test files. Can you rebase your patch?

CMakeLists.txt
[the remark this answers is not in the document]
friendlyanon: That just makes the configure process take longer. This should go in the documentation right next to the build type part. It's nice that mold can build with no exceptions, but it's never the end of the world if it happens to be built with exceptions on anyway.

CMakeLists.txt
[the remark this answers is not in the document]
friendlyanon: `target_link_libraries` **is** the way you link to targets or pass library names to the compiler. You don't have to worry about the names here, open source developers are generally pretty good with not stepping on others' toes with name conflicts, especially when it comes to such fundamental ones like libm.

CMakeLists.txt
friendlyanon: I see. In that case you can set things up like so:

```cmake
# FIXME: this is for parity with the makefiles that install directly to <prefix>/lib
# NOTE: defining this before including GNUInstallDirs makes its platform detection a noop
set(CMAKE_INSTALL_LIBDIR lib CACHE PATH "")

include(GNUInstallDirs)
```

I'd also be curious on what platform you are that still uses the `lib64` convention. You should talk to the mold package's maintainer about this for that platform. Maybe even @ them here if they are on GitHub.

CMakeLists.txt
friendlyanon: [CPACK_PRE_BUILD_SCRIPTS](https://cmake.org/cmake/help/latest/module/CPack.html#variable:CPACK_PRE_BUILD_SCRIPTS) allows you to run arbitrary code in the staging directory. I might just make an example for this later.

CMakeLists.txt
friendlyanon: Should this be made a separate issue? I think there is room to improve in this regard.

CMakeLists.txt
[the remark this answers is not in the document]
friendlyanon: Because this is something to be set from the command line. There are also vendored sources which this property does not affect anyway. Rarely is it a problem to build a standard conforming codebase with extensions either way.

CMakeLists.txt
clemenswasser: I'm on Fedora 36. Is this Platform/Distro dependent?

CMakeLists.txt
clemenswasser: Thanks! Will take a look.

CMakeLists.txt
friendlyanon: Yes, Ubuntu 11.04 and Debian 7 changed to `/usr/lib/<triplet>`. Some things to read on this topic:  
https://wiki.debian.org/Multiarch/  
https://wiki.ubuntu.com/MultiarchSpec

CMakeLists.txt
friendlyanon: Actually, this is not necessary on the mold side. Package maintainers can just do this in their install scripts. I tried looking at the [Debian packager script](https://salsa.debian.org/pkg-llvm-team/mold/-/blob/master/debian/rules) and it's just a makefile calling debhelper. I'm actually curious about a package maintainer's opinion on this. Since I was looking at Debian's package, hopefully @sylvestre can say something here.

The question here would be: is documenting the fact that these symlinks might be necessary enough for package maintainers? According to the [README](https://github.com/rui314/mold#how-to-use) these symlinks are only needed for GCC <12.1.0, if I understand things correctly.

CMakeLists.txt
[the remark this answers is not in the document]
clemenswasser: Done

CMakeLists.txt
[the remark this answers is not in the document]
clemenswasser: Okay, Thanks.

CMakeLists.txt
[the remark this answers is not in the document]
clemenswasser: Makes sense. Done

CMakeLists.txt
clemenswasser: Thanks for the info. I've adopted your change :+1:

rui314:
@friendlyanon 

I meant I would merge this but wouldn't recommend users to use CMake. In order to not confuse users, I'll move CMakeLists.txt under `contrib` directory to hide it from users.

rui314:
@clemenswasser Could you add a `signed-off-by` line to your commit messages as explained in https://github.com/rui314/mold/blob/main/CONTRIBUTING.md?

CMakeLists.txt
```diff
@@ -0,0 +1,204 @@
...
+include(CMakeDependentOption)
+
+# FIXME: this is for parity with the makefiles that install directly to <prefix>/lib
+# NOTE: defining this before including GNUInstallDirs makes its platform detection a noop
+set(CMAKE_INSTALL_LIBDIR lib CACHE PATH "")
+include(GNUInstallDirs)
+
+# Mold build options
+cmake_dependent_option(MOLD_USE_MIMALLOC "Use mimalloc" ON
+    "NOT CMAKE_SYSTEM_NAME MATCHES Darwin; NOT ANDROID" OFF)
```
friendlyanon, line 13:
```suggestion
    "NOT APPLE;NOT ANDROID" OFF)
```

https://cmake.org/cmake/help/latest/variable/APPLE.html

CMakeLists.txt
```diff
@@ -0,0 +1,204 @@
...
+option(MOLD_USE_SYSTEM_MIMALLOC "Use system or vendored mimalloc" ON)
```
friendlyanon, line 14:
```suggestion
cmake_dependent_option(MOLD_USE_SYSTEM_MIMALLOC "Use system or vendored mimalloc" ON
    MOLD_USE_MIMALLOC OFF)
```

CMakeLists.txt
```diff
@@ -0,0 +1,204 @@
...
+if(MOLD_USE_SYSTEM_TBB)
+    find_package(TBB REQUIRED)
+else()
+    function(mold_add_tbb)
+        set(CMAKE_BUILD_TYPE RelWithDebInfo)
+        set(BUILD_SHARED_LIBS OFF)
+        option(TBB_TEST "Enable testing" OFF)
+        string(APPEND CMAKE_CXX_FLAGS -D__TBB_DYNAMIC_LOAD_ENABLED=0)
+        option(TBB_STRICT "Treat compiler warnings as errors" OFF)
+        add_subdirectory(third-party/tbb EXCLUDE_FROM_ALL)
```
friendlyanon, line 44:
```suggestion
        set(BUILD_SHARED_LIBS OFF)
        set(TBB_TEST OFF CACHE INTERNAL "")
        set(TBB_STRICT OFF CACHE INTERNAL "")
        add_subdirectory(third-party/tbb EXCLUDE_FROM_ALL)
        target_compile_definitions(tbb PRIVATE __TBB_DYNAMIC_LOAD_ENABLED=0)
```

CMakeLists.txt
```diff
@@ -0,0 +1,204 @@
...
+        function(mold_add_mimalloc)
+            option(MI_BUILD_TESTS "Build test executables" OFF)
+            add_subdirectory(third-party/mimalloc EXCLUDE_FROM_ALL)
+            target_compile_definitions(mimalloc PRIVATE MI_USE_ENVIRON=0)
+        endfunction()
+
+        mold_add_mimalloc()
+    endif()
+
+    add_library(mimalloc::mimalloc ALIAS mimalloc)
```
friendlyanon, line 31: Why this alias?

CMakeLists.txt
```diff
@@ -0,0 +1,204 @@
...
+
+target_link_libraries(mold PRIVATE
+    TBB::tbb)
+
+if(MOLD_USE_MIMALLOC)
+    target_link_libraries(mold PRIVATE mimalloc::mimalloc)
+
+    if(MOLD_USE_SYSTEM_MIMALLOC)
+        target_compile_definitions(mimalloc INTERFACE USE_SYSTEM_MIMALLOC)
+    endif()
```
friendlyanon, line 84: Weird placement for this code. Should probably move it up to where mimalloc is imported in the first place.

CMakeLists.txt
```diff
@@ -0,0 +1,204 @@
...
+endif()
+
+if(NOT CMAKE_SYSTEM_NAME MATCHES Darwin)
```
friendlyanon, line 87:
```suggestion
if(NOT APPLE)
```

CMakeLists.txt
```diff
@@ -0,0 +1,204 @@
...
+    find_package(OpenSSL REQUIRED COMPONENTS Crypto)
+    target_link_libraries(mold PRIVATE OpenSSL::Crypto)
+endif()
+
+if(CMAKE_SYSTEM_PROCESSOR MATCHES riscv64
+    OR CMAKE_SYSTEM_PROCESSOR MATCHES armv6)
+    find_library(ATOMIC_LIBRARIES atomic REQUIRED)
+    target_link_libraries(mold PRIVATE ${ATOMIC_LIBRARIES})
```
friendlyanon, line 95:
```suggestion
    target_link_libraries(mold PRIVATE atomic)
```

`REQUIRED` was [added in CMake 3.18](https://cmake.org/cmake/help/latest/command/find_library.html#:~:text=repeated%20configure%20steps.-,REQUIRED,-New%20in%20version) to `find_library`. I also think that this is one of those libraries where you should just pass it as is to the compiler and let it figure it out.

CMakeLists.txt
```diff
@@ -0,0 +1,204 @@
...
+
+if(GIT_HASH MATCHES "ref: ")
+    string(REPLACE "ref: " "" GIT_HASH "${GIT_HASH}")
+    file(STRINGS .git/${GIT_HASH} GIT_HASH)
+endif()
+
+set_property(SOURCE main.cc macho/output-chunks.cc APPEND PROPERTY
+    COMPILE_DEFINITIONS MOLD_VERSION="${CMAKE_PROJECT_VERSION}")
+set_property(SOURCE main.cc APPEND PROPERTY
+    COMPILE_DEFINITIONS GIT_HASH="${GIT_HASH}")
```
friendlyanon, line 108:
```suggestion
    COMPILE_DEFINITIONS "MOLD_VERSION=\"${CMAKE_PROJECT_VERSION}\"")
set_property(SOURCE main.cc APPEND PROPERTY
    COMPILE_DEFINITIONS "GIT_HASH=\"${GIT_HASH}\"")
```

Please do not use make-ish [legacy quoting](https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#:~:text=Note%20To%20support-,legacy%20CMake%20code,-%2C%20unquoted%20arguments%20may).

CMakeLists.txt
```diff
@@ -0,0 +1,204 @@
...
+    COMMENT "Creating ld symlink"
+    VERBATIM)
+add_custom_command(
+    TARGET mold POST_BUILD
+    COMMAND ${CMAKE_COMMAND} -E create_symlink mold ld64
+    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+    COMMENT "Creating ld64 symlink"
+    VERBATIM)
+
+include(CTest)
```
friendlyanon, line 169: Please use the [module provided variable](https://cmake.org/cmake/help/latest/module/CTest.html#:~:text=module%20automatically%20creates%20a-,BUILD_TESTING,-option%20that%20selects) to guard everything testing related. This includes the symlink code above. [Example](https://github.com/friendlyanon/cmake-init-executable/blob/master/cmake/dev-mode.cmake#L3-L6).

CMakeLists.txt
```diff
@@ -0,0 +1,204 @@
...
+ProcessorCount(MOLD_NPROC)
+add_custom_target(
+    check
+    COMMAND "${CMAKE_COMMAND}" --build "${mold_BINARY_DIR}" --config "\$<CONFIG>" -j "${MOLD_NPROC}"
+    COMMAND "${CMAKE_CTEST_COMMAND}" -C "\$<CONFIG>" -j "${MOLD_NPROC}"
+    VERBATIM
+    USES_TERMINAL
+)
+
+install(TARGETS mold)
```
friendlyanon, line 182: Guard install rules with [CMAKE_SKIP_INSTALL_RULES](https://cmake.org/cmake/help/latest/variable/CMAKE_SKIP_INSTALL_RULES.html).

friendlyanon (changes requested):

CMakeLists.txt
clemenswasser: I wanted to ensure that it only links to a CMake Target and errors if it doesn't exist.
You are right, it is unnecessary, as it is always defined.

CMakeLists.txt
clemenswasser: Done, thanks!

CMakeLists.txt
clemenswasser: Done, thanks!

CMakeLists.txt
clemenswasser: Done, thanks!

CMakeLists.txt
clemenswasser: Done, thanks!

CMakeLists.txt
clemenswasser: Done, thanks!

CMakeLists.txt
clemenswasser: Done, thanks!

CMakeLists.txt
clemenswasser: Done, thanks!

CMakeLists.txt
clemenswasser: Done, thanks!

clemenswasser:
@3rav
Your 1. error was actually a issue on my side, which I now fixed. Thanks, good catch!
The 2. error however is really strange, did you accidentally deleted the third-party folder?

CMakeLists.txt
```diff
@@ -0,0 +1,201 @@
...
+    ProcessorCount(MOLD_NPROC)
+    add_custom_target(
+        check
+        COMMAND "${CMAKE_COMMAND}" --build "${mold_BINARY_DIR}" --config "\$<CONFIG>" -j "${MOLD_NPROC}"
+        COMMAND "${CMAKE_CTEST_COMMAND}" -C "\$<CONFIG>" -j "${MOLD_NPROC}"
+        VERBATIM
+        USES_TERMINAL)
+endif()
+
+if(CMAKE_SKIP_INSTALL_RULES)
```
friendlyanon, line 177:
```suggestion
if(NOT CMAKE_SKIP_INSTALL_RULES)
```

friendlyanon (approved):
Other than the single inverted logic I found, LGTM!

Would also be nice if downstream package managers chimed in regarding the symlink question.

CMakeLists.txt
clemenswasser: Oh, thanks.

CMakeLists.txt
```diff
@@ -0,0 +1,201 @@
...
+    mold_add_tbb()
+endif()
+
+include(CheckPIESupported)
+check_pie_supported()
+
+add_library(mold-wrapper SHARED)
+set_target_properties(mold-wrapper PROPERTIES
+    POSITION_INDEPENDENT_CODE ON
+    PREFIX "") # Remove the default `lib` prefix
```
friendlyanon, line 55:
```suggestion
add_library(mold-wrapper SHARED)
# Remove the default `lib` prefix
set_property(TARGET mold-wrapper PROPERTY PREFIX "")
```

Okay, one last thing. According to https://cmake.org/cmake/help/latest/prop_tgt/POSITION_INDEPENDENT_CODE.html

> This property is `True` by default for `SHARED` and `MODULE` library targets and `False` otherwise.

You also weren't checking the output either way, like in the [example](https://cmake.org/cmake/help/latest/module/CheckPIESupported.html#examples).

friendlyanon (changes requested):
Okay, I actually

joshcangit:
@clemenswasser
Using `make`, I can use the ` -C` option to set the working directory.
No such option exists for `cmake` as far as I know.

However, I do know how to work around this so to me this is a minor convenience.

3rav:
> @3rav Your 1. error was actually a issue on my side, which I now fixed. Thanks, good catch! The 2. error however is really strange, did you accidentally deleted the third-party folder?

The 2. is my bad, I used 1.3.1 version, new in main version are OK

friendlyanon:
> No such option exists for `cmake` as far as I know.

`cmake --build` **requires** a path argument: https://cmake.org/cmake/help/latest/manual/cmake.1.html#build-a-project

clemenswasser:
I rebased on current main and added the definition of `MOLD_VERSION` for `elf/lto.cc`, which it now uses.

rui314:
I applied these patches locally but I cannot build mold due to the following error:

```
ruiu@blue:~/mold$ mkdir build
ruiu@blue:~/mold$ cd build
ruiu@blue:~/mold/build$ cmake ..
-- The C compiler identification is Clang 14.0.0
-- The CXX compiler identification is Clang 14.0.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/lib/ccache/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/lib/ccache/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Error at /home/ruiu/mimalloc/cmake/mimalloc-config.cmake:1 (include):
  include could not find requested file:

    /home/ruiu/mimalloc/cmake/mimalloc.cmake
Call Stack (most recent call first):
  CMakeLists.txt:21 (find_package)


CMake Error at CMakeLists.txt:22 (target_compile_definitions):
  Cannot specify compile definitions for target "mimalloc" which is not built
  by this project.


CMake Error at CMakeLists.txt:36 (find_package):
  By not providing "FindTBB.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "TBB", but
  CMake did not find one.

  Could not find a package configuration file provided by "TBB" with any of
  the following names:

    TBBConfig.cmake
    tbb-config.cmake

  Add the installation prefix of "TBB" to CMAKE_PREFIX_PATH or set "TBB_DIR"
  to a directory containing one of the above files.  If "TBB" provides a
  separate development package or SDK, be sure it has been installed.


-- Configuring incomplete, errors occurred!
See also "/home/ruiu/mold/build/CMakeFiles/CMakeOutput.log".
```

clemenswasser:
As discussed in this thread: https://github.com/rui314/mold/pull/571#discussion_r916267809, using the system TBB and mimalloc is now the default. You either have to install them with your system package manager or set `-DMOLD_USE_SYSTEM_MIMALLOC=OFF -DMOLD_USE_SYSTEM_TBB=OFF` to keep using the vendored dependencies.

sylvestre:
You might want to improve the error message if TBB & mimalloc aren't found on the system

rui314:
I'm against changing the default. At least it shouldn't be done by this commit.

rui314:
It looks like `ninja test` runs all tests serially. Do you think you can run them in parallel?

clemenswasser:
the `check` target runs the tests in parallel by default and ensures that all targets get build first.
You can do `cmake --build <buildDir> -t check`.
When you execute ctest itself there it the typical `-j <jobs>` flag.

rui314:
Did you know the difference between `ninja test` and `ninja check`?

CMakeLists.txt
```diff
@@ -0,0 +1,198 @@
...
+    "NOT APPLE;NOT ANDROID" OFF)
+cmake_dependent_option(MOLD_USE_SYSTEM_MIMALLOC "Use system or vendored mimalloc" ON
+    MOLD_USE_MIMALLOC OFF)
+option(MOLD_USE_SYSTEM_TBB "Use system or vendored TBB" ON)
+
+# Setup mimalloc
+if(MOLD_USE_MIMALLOC)
+    if(MOLD_USE_SYSTEM_MIMALLOC)
+        find_package(mimalloc REQUIRED)
+        target_compile_definitions(mimalloc INTERFACE USE_SYSTEM_MIMALLOC)
```
friendlyanon, line 22: I forgot you can't modify IMPORTED targets like this.

```suggestion
        set_property(TARGET mimalloc APPEND PROPERTY INTERFACE_COMPILE_DEFINITIONS USE_SYSTEM_MIMALLOC)
```

clemenswasser:
> Did you know the difference between `ninja test` and `ninja check`?

The difference is that the test target is generated by CMake and only runs `ctest` on the build directory (no parallelism and doesn't build mold, which can result in strange file not found errors). The check target is a custom one by us, which builds and then runs the tests in parallel.

clemenswasser:
> You might want to improve the error message if TBB & mimalloc aren't found on the system 

Note that this error message is generated by CMake and not by us. In theory we could check the "*_FOUND" variables and output a different error message, but I have never seen this before...

friendlyanon (changes requested):

friendlyanon:
@rui314 The below script successfully builds mold and its dependencies:

```sh
#!/bin/bash

set -euo pipefail

git clone git@github.com:microsoft/mimalloc.git
(
cd mimalloc
cmake -GNinja -DBUILD_SHARED_LIBS=1 -DCMAKE_BUILD_TYPE=Release -B build
cmake --build build
cmake --install build --prefix ../mimalloc-prefix
)
rm -rf mimalloc

git clone git@github.com:oneapi-src/oneTBB.git
(
cd oneTBB
cmake -GNinja -DBUILD_SHARED_LIBS=1 -DCMAKE_BUILD_TYPE=Release -B build -DTBB_TEST=0 -DTBB_STRICT=0 -DTBBMALLOC_BUILD=0 -DTBB_ENABLE_IPO=0
cmake --build build
cmake --install build --prefix ../oneTBB-prefix
)
rm -rf oneTBB

git clone git@github.com:clemenswasser/mold.git -b cmake
cmake -S mold -B mold/build -GNinja -DBUILD_SHARED_LIBS=1 -DCMAKE_BUILD_TYPE=Release \
    "-DCMAKE_PREFIX_PATH=$PWD/oneTBB-prefix;$PWD/mimalloc-prefix" \
    '-DCMAKE_CXX_FLAGS=-fno-exceptions -fno-unwind-tables -fno-asynchronous-unwind-tables'
cmake --build mold/build
```

So I'm not sure where your error comes from.

CMakeLists.txt
clemenswasser: Shouldn't this be equivalent? Also I just checked and `USE_SYSTEM_MIMALLOC` gets defined when using `target_compile_definitions`.

clemenswasser:
I just checked, for debian sid/unstable it is the same problem as with ubuntu kinetic/devel. They include the cmake files in the `libmimalloc2.0` package but the include dir is incorrect as you can see from the build log:
```
CMake Error in CMakeLists.txt:
  Imported target "mimalloc" includes non-existent path

    "/usr/include/mimalloc-2.0"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.
```

friendlyanon:
Oh I see, there is something fishy going on with the Debian packages. [libmimalloc2.0](https://packages.debian.org/bookworm/amd64/libmimalloc2.0/filelist) contains the CMake package, but that is missing from [libmimalloc-dev](https://packages.debian.org/bookworm/amd64/libmimalloc-dev/filelist). I think this is an actual bug in the way mimalloc is packaged. There is no reason to provide the CMake package in the runtime component, since it's only ever used for developer purposes.

cc @alexmyczko Apologies for the ping again, just want to make sure you see this. Reach out to me via email if you want to talk about this.

alexmyczko:
@friendlyanon http://sid.ethz.ch/debian/mimalloc/2022b contains the CMake package in -dev instead of the library. can you confirm with 2.0.6+ds-2 if it's fixed?

friendlyanon:
I checked the -dev package and what mimalloc itself builds, and I'm not sure how the build rules on salsa achieve what goes into the .deb.

Expected output from mimalloc itself with `-DMI_BUILD_STATIC=0 -DMI_BUILD_OBJECT=0 -DMI_BUILD_TESTS=0 -DCMAKE_BUILD_TYPE=Release`:

```
mimalloc-prefix/
├── include
│   └── mimalloc-2.0
│       ├── mimalloc.h
│       ├── mimalloc-new-delete.h
│       └── mimalloc-override.h
└── lib
    ├── cmake
    │   └── mimalloc-2.0
    │       ├── mimalloc.cmake
    │       ├── mimalloc-config.cmake
    │       ├── mimalloc-config-version.cmake
    │       └── mimalloc-release.cmake
    ├── libmimalloc.so -> libmimalloc.so.2
    ├── libmimalloc.so.2 -> libmimalloc.so.2.0 # these ought to go in the runtime package
    └── libmimalloc.so.2.0                     # these ought to go in the runtime package
```

As you can see, the includes are prefixed with a version and the `mimalloc.cmake` propagates that same directory in its properties.

Output from your upload(s):

```
mimalloc-dev/usr/
├── include # where is mimalloc-2.0 ???
│   ├── mimalloc.h
│   ├── mimalloc-new-delete.h
│   └── mimalloc-override.h
├── lib
│   └── x86_64-linux-gnu
│       ├── cmake
│       │   └── mimalloc-2.0
│       │       ├── mimalloc.cmake
│       │       ├── mimalloc-config.cmake
│       │       ├── mimalloc-config-version.cmake
│       │       └── mimalloc-none.cmake
│       └── libmimalloc.so -> libmimalloc.so.2
```

As you can see, the subdirectory for the headers is not present and the `mimalloc.cmake` file still contains a path to that.

alexmyczko:
bugs.debian.org/1006249 for reference why the mimalloc subdir in include was removed (from debian/changelog)

friendlyanon:
While the paths are correct now, your bugreport was requesting mimalloc headers to be directly in the system includes of compilers. mimalloc upstream build file allows this without patches, so their CML isn't entirely unsalvageable.

I recommend applying this patch and distributing like that:

```diff
--- debian/rules
+++ debian/rules
@@ -21,6 +21,9 @@
 override_dh_auto_configure:
 	dh_auto_configure -- \
 		-DMI_BUILD_STATIC=OFF \
+		-DMI_BUILD_OBJECT=OFF \
+		-DMI_BUILD_TESTS=OFF \
+		-DMI_INSTALL_TOPLEVEL=ON \
 		-DMI_USE_LIBATOMIC=$(LIBATOMIC) \
 
 override_dh_missing:
```

This also disables tests and an object library, so you don't unnecessarily build things that are just discarded anyway.  
Worth noting that mimalloc says that `MI_INSTALL_TOPLEVEL` is deprectad, however that is the NORMAL way to do things. They don't even require this for any particular reason, since their includes aren't in a subdirectory (i.e. their usage is `#include <mimalloc.h>` not `#include <mimalloc/mimalloc.h>`, but then you wouldn't need an option for that either way).

Anyhow, a little offtopic, but the above allows all users to use mimalloc easily.

alexmyczko:
i agree will update accordingly asap (likely tomorrow)

alexmyczko:
it has been uploaded and built meanwhile https://buildd.debian.org/status/package.php?p=mimalloc&suite=sid

CMakeLists.txt
```diff
@@ -0,0 +1,198 @@
...
+add_library(mold-wrapper SHARED)
+
+# Remove the default `lib` prefix
+set_target_properties(mold-wrapper PROPERTIES PREFIX "")
+target_link_libraries(mold-wrapper PRIVATE ${CMAKE_DL_LIBS})
+target_sources(mold-wrapper PRIVATE elf/mold-wrapper.c)
+
+add_executable(mold)
+target_compile_features(mold PRIVATE cxx_std_20)
+target_include_directories(mold SYSTEM PRIVATE third-party)
```
jonesmz, line 58: This should probably be brought into the mold executable build via target_link_libraries, and not by adding the include directory to mold itself.

friendlyanon: Now that you mention it... the includes in this directory might be picked up before the system dependencies because of this. There should be a separate PR that fixes includes:

```diff
diff --git a/demangle.cc b/demangle.cc
--- demangle.cc
+++ demangle.cc
@@ -1,9 +1,9 @@
 #include "mold.h"
 
 #include <cstdlib>
 #include <cxxabi.h>
-#include <rust-demangle/rust-demangle.h>
+#include <rust-demangle.h>
 
 namespace mold {
 
 std::string_view demangle(std::string_view name) {
diff --git a/Makefile b/Makefile
--- Makefile
+++ Makefile
@@ -44,9 +44,10 @@
 CFLAGS = -O2
 CXXFLAGS = -O2
 
 MOLD_CXXFLAGS := -std=c++20 -fno-exceptions -fno-unwind-tables \
-                 -fno-asynchronous-unwind-tables -Ithird-party \
+                 -fno-asynchronous-unwind-tables \
+                 -Ithird-party/rust-demangle -Ithird-party/xxhash \
                  -DMOLD_VERSION=\"$(VERSION)\" -DLIBDIR="\"$(LIBDIR)\""
 
 MOLD_LDFLAGS := -pthread -lz -lm -ldl
 
diff --git a/mold.h b/mold.h
--- mold.h
+++ mold.h
@@ -26,9 +26,9 @@
 #include <unistd.h>
 #include <vector>
 
 #define XXH_INLINE_ALL 1
-#include <xxhash/xxhash.h>
+#include <xxhash.h>
 
 #ifdef NDEBUG
 #  define unreachable() __builtin_unreachable()
 #else
```

CMakeLists.txt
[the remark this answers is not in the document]
sergiud: This is by design. An unset `CMAKE_BUILD_TYPE` is an extra build type and thus does not need a reminder or a hard-coded override. As already mentioned, setting `CMAKE_BUILD_TYPE` does not have any effect in multi-config generators such as MSVC anyway.

CMakeLists.txt
friendlyanon: Submitted the above patch as #601

rui314:
I'm now trying to build mold on Windows. Let me merge this so that it's easier to build it on that platform.

clemenswasser:
> I'm now trying to build mold on Windows. Let me merge this so that it's easier to build it on that platform.

@rui314 FYI I've already done a lot of fixes for Windows on my `win` branch: https://github.com/clemenswasser/mold/commits/win. The only things left are porting the remaining `mmap` related calls and finding a workaround for the 128 branch limit of the MSVC compiler we currently exceed within the argument parsing.

rui314:
Could you send me PRs if they are ready?

PR #575, opened 2022-07-09
Fix double git installation in CI workflow

ZhongRuoyu:
The CI job `build-clang` has duplicated `git` installation.

Signed-off-by: Zhong Ruoyu <zhongruoyu@outlook.com>


rui314:
Thanks!

PR #576, opened 2022-07-09
Fix run.sh failure on musl

wangp:
The command "mold -run foo.ld" can fail when foo.ld is an empty file.
The musl execvp() call fails with "Exec format error" when the script
to be executed is empty. The fix is to create foo.ld with a valid
Unix shebang.


rui314:
Can you add a `Signed-off-by` line to your commit message as explained in https://github.com/rui314/mold/blob/main/CONTRIBUTING.md?

wangp:
Done.

rui314:
Thanks!

PR #578, opened 2022-07-09
Add macos usage verification method

ksato9700:
I added a description for macOS to check if mold was used to build the executable.

Signed-off-by: Kenichi Sato <ksato9700@gmail.com>


rui314:
I don't want to document mold/macOS at this moment as it's not for general use. We write documents for supported features and vice versa. mold/macOS is intentionally undocumented.

PR #585, opened 2022-07-14
Add support to install deps on Arch

aslikr42069:
Add support for Arch Linux in the script for installing dependencies

Signed-off-by: Lorena R. Contreras<arturo211234@gmail.com>

### DIFF
--- a/install-build-deps.sh
+++ b/install-build-deps.sh
@@ -29,6 +29,9 @@ opensuse-tumbleweed*)
 gentoo*)
   emerge dev-vcs/git dev-util/cmake sys-libs/zlib
   ;;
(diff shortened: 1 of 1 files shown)


rui314:
With this change, does `make test-all` pass?

aslikr42069:
It fails when testing absolute symbols because it looks for `x86_64-linux-gnu-gcc`, whereas Arch Linux names its gcc binary `x86_64-pc-linux-gnu-gcc`, therefore failing. Every step up until then, however, works; and the mold binary is successfully built

rui314:
OK, thank you for confirming.

PR #586, opened 2022-07-15
Fix spelling of "support".

ltratt:
Two (identical) minor typos. Spotted when checking whether mold supports `-b binary` or not.


rui314:
Could you add a `signed-off-by` line as explained in https://github.com/rui314/mold/blob/main/CONTRIBUTING.md?

ltratt:
Fixed!

PR #587, opened 2022-07-18
fix wording in readme to speed compared to cp

trunkensailor:
"2x slower than" is not really wording that makes sense, if you think about it.
I'm even just guessing in my correction of it, that this is what you meant, because it's ambiguous.

Firstly, you just measure the time something takes, and not its "slowness", so the term is kind of confusing as being "slow" is subjective and only makes sense when compared to something else, but in this case you are saying both `mold` and `cp` are slow.

Secondly, if some object is 1cm long, and another is 3cm long, you could say the second one is 3x as big as the first one or the second one is 2x bigger than the first one (1cm + 2 * (1cm) = 3cm).

So, saying it's "2x slower than" would, by this logic mean, if you take "slowness" as "the time it takes to run", that it would take 3x as much time, which I don't think is what you meant.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ on a simulated 8-core 16-threads machine.
 | Clang 13 (3.18 GiB)           | 64.12s   | 5.82s    | 2.90s
 | Firefox 89 libxul (1.64 GiB)  | 32.95s   | 6.80s    | 1.42s
 
-mold is so fast that it is only 2x _slower_ than `cp` on the same
+mold is so fast that it only takes twice as long to run as running `cp` takes on the same
 machine. Feel free to [file a bug](https://github.com/rui314/mold/issues)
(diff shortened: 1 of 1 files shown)


rui314:
I believe in computer science context, _n_ times faster is defined as it takes 1/_n_ of time compared to another program, and _n_ times slower is defined as the inverse of it (which is _n_/1). I believe I learned it from a decent text book, but I don't remember the title of it, though.

ishitatsuyuki:
Just my two cents but I also think this is fine. 2x speedup means running at half the time. Likewise, 2x slower means running at twice the time. There doesn't seem to be an ambiguity here.

The writing style might not be optimal, but we'd better spend time on useful stuff instead of bikeshedding about the usage of English.

Issue #532, opened 2022-06-04
Feature request: --export-dynamic-symbol for gold & lld compatibility


kinke:
To selectively export some symbols directly in the cmdline, e.g.: `--export-dynamic-symbol=sym1 --export-dynamic-symbol=sym2`.

rui314:
Yes, we need to support that option as well as `--export-dynamic-symbol-list` for compatibility with other linkers.

ishitatsuyuki:
I'd like to look into this.

rui314:
Thanks. Please go ahead. I believe you can use our `MultiGlob` class for glob pattern matching.

ishitatsuyuki:
@rui314 I have a question on how dynamic and version lists are handled currently.

The `--dynamic-list` and `--version-script` arg are handled in `read_input_files`:

https://github.com/rui314/mold/blob/990b0ac4cd78081b57231a3658f34f8957797576/elf/main.cc#L217-L221

... which implies their behavior depends on argument order. But I couldn't figure out what kind of options can change their behavior. Do you have an example where this becomes order dependent, or in case it's order independent could we handle it in `parse_nonpositional_args` instead?

rui314:
I don't remember why I wanted to handle these options in `read_input_files`. You can probably move them to `parse_nonpositional_args` safely.

PR #588, opened 2022-07-20
[elf] Implement --export-dynamic-symbol[-list]. 

ishitatsuyuki:
See descriptions in commit messages.

Closes: #532

### DIFF
--- a/elf/linker-script.cc
+++ b/elf/linker-script.cc
@@ -349,6 +349,14 @@ void parse_version_script(Context<E> &ctx, std::string path) {
 }
 
 template <typename E>
+void parse_dynamic_glob(Context <E> &ctx, std::string_view arg, bool is_cpp) {
+  if (arg == "*")
+    ctx.default_version = VER_NDX_GLOBAL;
+  else
+    ctx.version_patterns.push_back({unquote(arg), VER_NDX_GLOBAL, is_cpp});
+}
+
+template <typename E>
 void read_dynamic_list_commands(Context<E> &ctx, std::span<std::string_view> &tok,
                                 bool is_cpp) {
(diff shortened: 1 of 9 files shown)


elf/linker-script.cc
```diff
@@ -348,6 +348,14 @@ void parse_version_script(Context<E> &ctx, std::string path) {
     SyntaxError(ctx, tok[0]) << "trailing garbage token";
 }
 
+template <typename E>
+void parse_dynamic_glob(Context <E> &ctx, std::string_view arg, bool is_cpp) {
+  if (arg == "*")
+    ctx.default_version_from_glob = VER_NDX_GLOBAL;
```
rui314, line 354: I believe linker scripts are parsed after `parse_nonpositional_args`, so isn't it too late to assign a value to this variable?

ishitatsuyuki: I changed their parsing order in https://github.com/rui314/mold/pull/588/commits/0525da2c0385a5498e8b66c2df121737a417a863.

ishitatsuyuki: (Or did you actually meant linker script not version script? IIRC linker scripts are not implemented yet?)

rui314: Linker script is supported, though we support only a limited set of directive such as `INPUT`, `OUTPUT_FORMAT` or `VERSION`. And with `VERSION`, you can embed a version script into a linker script.

ishitatsuyuki: OK, I see, will try to get this fixed.

ishitatsuyuki: I've reverted to parsing these as positional args in the latest revision.

PR #590, opened 2022-07-28
[ELF][RISCV] add R_RISCV_SET32 in EhFrameSection

alexfanqi:
Hi,

Compiling llvm and mesa with mold gives some linking errors like:
```
mold: fatal: unsupported relocation in .eh_frame: R_RISCV_SET32
```
ref: https://bugs.gentoo.org/861488

From my understanding of llvm riscv backend, it will emit R_RISCV_SET32 as part of DW_CFA_advance_loc4. So is R_RISCV_SET32 missing here or intentionally avoided for some reasons?


rui314:
Looks good, but can you add a `signed-off-by` line to your commit message as explained in https://github.com/rui314/mold/blob/main/CONTRIBUTING.md?

Issue #592, opened 2022-07-29
symbol export issues


zmike:
We're evaluating mold for use in Mesa CI, but there's some issues with unit tests related to symbol exports failing with the latest version. See also comments here https://gitlab.freedesktop.org/mesa/mesa/-/issues/6877#note_1487555

rui314:
Thank you for your report. Let me take a look. What was the command you used to build Mesa?

zmike:
I think something like `meson . build -Dgallium-drivers=swrast -Dbuild-tests=true` should be enough to have the unit tests built? I imagine if you comment on that ticket then people will rapidly reply to help get you whatever info you need.

ishitatsuyuki:
The symbols are specified through a version script (using AMD Vulkan as an example):

https://github.com/mesa3d/mesa/blob/ae2947fbc5a7ad3879090777c829e44b7c339eb3/src/amd/vulkan/meson.build#L152-L155

For my setup (clang + ThinLTO), it looks like I get a few redundant symbols related to splitstack (only redundant one noted below):
```
  3623: 0000000000000000   184 TLS     GLOBAL DEFAULT   21 __morestack_initial_sp
  3625: 00000000000000b8     8 TLS     GLOBAL DEFAULT   21 __morestack_current_segment
  3626: 00000000000000c0     8 TLS     GLOBAL DEFAULT   21 __morestack_segments
  3633: 000000000005e590   356 FUNC    GLOBAL DEFAULT   13 __splitstack_find
  3634: 000000000005e700    98 FUNC    GLOBAL DEFAULT   13 __splitstack_block_signals
  3635: 000000000005e770   163 FUNC    GLOBAL DEFAULT   13 __splitstack_getcontext
  3636: 000000000005e820   116 FUNC    GLOBAL DEFAULT   13 __splitstack_setcontext
  3637: 000000000005e8a0   409 FUNC    GLOBAL DEFAULT   13 __splitstack_makecontext
  3638: 000000000005ea40   112 FUNC    GLOBAL DEFAULT   13 __splitstack_resetcontext
  3639: 000000000005eab0    14 FUNC    GLOBAL DEFAULT   13 __splitstack_releasecontext
  3640: 000000000005eac0    40 FUNC    GLOBAL DEFAULT   13 __splitstack_block_signals_context
  3641: 000000000005eaf0    88 FUNC    GLOBAL DEFAULT   13 __splitstack_find_context
```

ishitatsuyuki:
[...]

PR #593, opened 2022-07-30
[elf] Align .symtab's ELF attributes with .dynsym's. 

ishitatsuyuki:
Plus a few related refactors.

I found that the reference values for elf/symtab.sh was actually wrong, and after updating the reference values I think it's sufficient to act as a regression test for this.

Closes: #592

Unresolved question:
- [ ] The `.dynsym` logic only writes `st_visibility` if the symbol is not absolute nor a TLS object. Is the condition safe? Or should we always carry over `st_visibility` from `sym->visibility` (which is a minimum among input files)?


ishitatsuyuki:
Hmm, no idea why tsan is failing...

ishitatsuyuki:
Turns out that a designated initializer is not guaranteed to initialize the padding, and `Elf64Sym` contained such bit paddings and readelf was complaining about reading garbage. Learned my lesson.

PR #599, opened 2022-08-05
Support derivative OSes when installing dependencies

nbarrios1337:
This PR expands `install-build-deps.sh` to support derivative OSes that set the ID_LIKE variable in `/etc/os-release`.

These changes, in summary, move the existing work into a function, and proceed to call that function with the original "$ID$VERSION_ID" string. If that fails to install dependencies, we check again with the first OS listed in the ID_LIKE variable.

### DIFF
--- a/install-build-deps.sh
+++ b/install-build-deps.sh
@@ -33,16 +33,16 @@ test_id_version() {
     pacman -S --needed --noconfirm base-devel zlib openssl cmake util-linux git
     ;;
   *)
+    echo "Error: don't know anything about build dependencies on $1 $2"
     return 1
     ;;
   esac
 }
 
 if ! test_id_version "$ID" "$VERSION_ID" ; then
-  echo "Error: don't know anything about build dependencies on $ID $VERSION_ID"
   echo "Testing related OS"
   if ! test_id_version "${ID_LIKE%% *}" "$VERSION_ID" ; then
-    echo "Error: don't know anything about build dependencies on ${ID_LIKE%% *} $VERSION_ID"
+    echo "Error: Failed to find matching dependencies for the system"
     exit 1
   fi
 fi
--- a/install-build-deps.sh
+++ b/install-build-deps.sh
@@ -4,35 +4,45 @@ source /etc/os-release
 
 # The first line for each distro installs a build dependency.
 # The second line installs extra packages for `make test`.
+test_id_version() {
+  case "$1$2" in
+  ubuntu20.*)
+    apt-get install -y git cmake libssl-dev zlib1g-dev gcc g++ g++-10
+    apt-get install -y file bsdmainutils
+    ;;
+  ubuntu22.* | debian11)
+    apt-get install -y git cmake libssl-dev zlib1g-dev gcc g++
+    apt-get install -y file bsdmainutils
+    ;;
+  fedora*)
+    dnf install -y git gcc-g++ cmake openssl-devel zlib-devel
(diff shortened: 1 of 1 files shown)


rui314:
What distros do you actually want to support? I don't think there is always `FooLike` for distro `Foo`, so I don't think we need to support all possible cases.

nbarrios1337:
I personally use Pop!_OS, which sets `ID=pop` and `ID_LIKE="ubuntu debian"`, and I think this would be of great help to the myriad of arch derivatives out there.

Note: There is no <OS>_LIKE variables (i.e. UBUNTU_LIKE, ARCH_LIKE), there's only one ID_LIKE that OS maintainers set to be one or more values. More info in the [`/etc/os-release` man-page](https://www.freedesktop.org/software/systemd/man/os-release.html)

Because of that, the responsbility of supporting some arbitrary derivative falls on that derivative's maintainer(s) setting `ID_LIKE` to one of the more mainstream OSes that this script happens to support

install-build-deps.sh
```diff
@@ -4,35 +4,45 @@ source /etc/os-release
...
+    ;;
+  *)
+    echo "Error: don't know anything about build dependencies on $1 $2"
+    return 1
+    ;;
+  esac
+}
 
-case "$ID$VERSION_ID" in
-ubuntu20.*)
```
rui314, line 9: Then I think we should simply add `pop20.*` here and `pop22.*` below.

nbarrios1337: Doesn't that leave behind possible other derivatives?

From my point of view, I see these changes as way to stop the script from increasing in scope with every derivative someone submits a PR for.

I think that anyone tech-savvy enough to compile mold from source, and happens to use a derivative OS, would be able to copy the relevant dependency command out from `install-build-deps.sh` for their system.

rui314: I don't think there are too many derived distros. Do you have anything other than Pop!_OS? Also, users of minor distros are likely to be tech savvy, so they can easily run commands by hand. To make it crystal clear that what commands have to be run, I want to keep this script as simple as possible, so I don't want to use anything other than table lookup.

nbarrios1337: That makes sense to me, thank you for the clarification. Copying the apt command is what I did when installing v1.4.0.

In light of the expectations of small distro users and the goal of keeping this script simple, I think it would be best not to merge this PR at all

rui314: I'm fine to add `pop20.*` and `pop22.*` to the table. As long as we don't add any logic, adding new entries to these pattern matches doesn't increase complexity.

nbarrios1337: If that's the case, I'll make a separate PR for that; modifying this PR and its commit history to reflect only that small change seems misleading and disconnected from the earlier discussion.

rui314: OK, then let me close this PR.

rui314: I added pop-20 and pop-22 to the list. I noticed that `ID_LIKE` wasn't suitable for our purpose because it doesn't specify a version number. For example, `ubuntu` is defined as a `debian`-like in their `/etc/os-release`, but it doesn't say anything about what version it corresponds to. Trying to install as if it's `debian-20` isn't right (the latest debian version is 11.) So just enumerating all possible OSes and version numbers seems to be the best we can do.

nbarrios1337: That was definitely a problem I failed to consider. Thank you for finding a solution for this that works for your project, I enjoyed being a (small) part of the process :)

PR #601, opened 2022-08-06
Include only what is needed from third-party

friendlyanon:
Both rust-demangle and xxhash provide usage instructions with include
paths that do not have an extra directory component. This patch fixes
uses of these libraries in the code and avoids including the third-party
directory wholesale.

Found by https://github.com/rui314/mold/pull/571#discussion_r937135058


rui314:
Thanks!

PR #608, opened 2022-08-11
Initial Windows fixes

clemenswasser:
This fixes some of the errors when compiling for Windows but **not** all of them.
The remaining ones are mostly related to `mmap` calls and one in `elf\cmdline.cc`:
`fatal error C1061: compiler limit: blocks nested too deeply`
MSVC does only support up to 128 `else if` branches.
@rui314 I have reverted your changes related to the usage of the vendored sha256 implementation as I have already a implementation using CNG.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,8 +24,12 @@ endif()
 add_executable(mold)
 target_compile_features(mold PRIVATE cxx_std_20)
 target_link_libraries(mold PRIVATE ${CMAKE_DL_LIBS})
-target_compile_options(mold PRIVATE -fno-exceptions -fno-unwind-tables
-  -fno-asynchronous-unwind-tables)
+if(NOT MSVC)
+    target_compile_options(mold PRIVATE
+        -fno-exceptions
+        -fno-unwind-tables
+        -fno-asynchronous-unwind-tables)
+endif()
 
 # Build mold with -flto if MOLD_LTO=On
 option(MOLD_LTO "Build mold with link-time optimization enabled")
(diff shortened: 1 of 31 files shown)


CMakeLists.txt
```diff
@@ -197,7 +204,11 @@ target_sources(mold PRIVATE
   macho/tapi.cc
   macho/yaml.cc
   third-party/rust-demangle/rust-demangle.c
-  ${CMAKE_CURRENT_BINARY_DIR}/git-hash.cc)
+  git-hash.cc)
+
+if(MSVC)
+    target_compile_options(mold PRIVATE /bigobj)
```
rui314, line 210: Is this option needed?

demangle.cc
```diff
@@ -29,6 +32,9 @@ std::optional<std::string_view> cpp_demangle(std::string_view name) {
   static thread_local char *buf;
   static thread_local size_t buflen;
 
+  // TODO(cwasser): Actually demangle Symbols on Windows using e.g.
```
rui314, line 35: My goal is to make mold work exactly the same way on any platform as long as the same command line options and the same set of input files are given. This is important for cross-compilation. To achieve this goal, we eventually need to be able to demangle Itanium-mangled symbols, or otherwise we can't handle `extern "C"` symbols in version scripts.

I think we can just copy demangler source files from libcxxabi.

rui314:
Overall looking good. Thanks!

CMakeLists.txt
clemenswasser: Yes, else I wouldn't have added it 😉
On a more serious note, MSVC only supports up to 64k addressable sections by default, `/bigobj` raises this limit. It's used relatively frequently.

demangle.cc
clemenswasser: Sounds like a good solution 👍

PR #612, opened 2022-08-13
Windows: does not test pow() when compiling also with mingw

vtorri:


rui314:
Could you add `Signed-off-by` line to your commit message as explained in https://github.com/rui314/mold/blob/main/CONTRIBUTING.md?

With this change, does mold compile with mingw?

vtorri:
this patch is just that it disable a useless test as, like the Microsoft compiler, pow() always exists when using gcc.

i'll redo the commit

Issue #550, opened 2022-06-19
Even the "hello world" Rust program segfaults on ARMV7 when built with mold.


turkerali:
Hi again.
I reopened this issue as requested. Steps to reproduce segfault:
- Install rust on armv7 via "curl https://sh.rustup.rs -sSf | sh" 
- cargo new mold_test
- Edit .cargo/config.toml and add:
[target.armv7-unknown-linux-gnueabihf]
linker = "clang"
rustflags = ["-C", "link-arg=-fuse-ld=/usr/local/bin/mold"]
- Cargo run
- And the result is:
Compiling mold_test v0.1.0 (/root/aet_rust/mold_test)
    Finished dev [unoptimized + debuginfo] target(s) in 22.73s
     Running `target/debug/mold_test`
Segmentation fault

Hope this helps to debug this issue.

jpalus:
Can confirm issue happens for me as well both on armv6 and armv7. Segfaulting binary: [mold_test.zip](https://github.com/rui314/mold/files/8938755/mold_test.zip). Works fine on aarch64 though.

PR #625, opened 2022-08-15
Make sure symbol points to correct section.

rbartlensky:
I was debugging #550, and I noticed a symbol was pointing to the
start of a section, but its shndx was set to SHN_UNDEF. Looking over
the code, I realized that for each section you create a corresponding
output section, but if the input section is mergeable, you also create
a fragment for it whose parent is MergeableSection, which in the end
will be copied into the output binary instead of the output section.

Since the symbol was pointing to the start of a mergeable section, I
realized that it is probably wrong to use the shndx of
`isec->output_section`, because that particular Chunk never actually
makes it to the list of final Chunks for mergeable sections.

Before:
```
$ eu-readelf -s ./target/debug/mold_test | grep _rustc
  Num:    Value   Size Type    Bind   Vis          Ndx Name
   77: 000d563c     34 OBJECT  LOCAL  DEFAULT    UNDEF __rustc_debug_gdb_scripts_section__

$ eu-readelf -S ./target/debug/mold_test
[Nr] Name                 Type         Addr     Off    Size   ES Flags Lk Inf Al
...
[22] .debug_gdb_scripts   PROGBITS     000d563c 0d563c 000022  0 A      0   0  1
...
```

After:
```
$ eu-readelf -s ./target/debug/mold_test | grep _rustc
  Num:    Value   Size Type    Bind   Vis          Ndx Name
   77: 000d563c     34 OBJECT  LOCAL  DEFAULT       22 __rustc_debug_gdb_scripts_section__

$ eu-readelf -S ./target/debug/mold_test
[Nr] Name                 Type         Addr     Off    Size   ES Flags Lk Inf Al
...
[22] .debug_gdb_scripts   PROGBITS     000d563c 0d563c 000022  0 A      0   0  1
...
```

### DIFF
--- a/elf/mold.h
+++ b/elf/mold.h
@@ -1787,6 +1787,7 @@ public:
   bool has_plt(Context<E> &ctx) const;
   bool has_got(Context<E> &ctx) const;
 
+  u64 get_st_shndx() const;
   bool is_absolute() const;
   bool is_relative() const { return !is_absolute(); }
   bool is_local() const;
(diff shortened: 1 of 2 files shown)


rui314:
Thank you!

PR #628, opened 2022-08-16
tests: grep for output of nm in defsym2.sh test

marxin:
Right now, the test prints to the output:

```
Testing defsym2 ...                  w _ITM_deregisterTMCloneTable
                 w _ITM_registerTMCloneTable
                 U __cxa_finalize
                 w __gmon_start__
00000000000010f9 T bar
00000000000010f9 T foo
OK
```

which is changed to:

```
Testing defsym2 ... OK
```


rui314:
Thanks!

PR #635, opened 2022-08-18
remove unused to_string impl for Digest (aka array<uint8_t, ...>)

nevack:
Fixes this warning when building with make or Cmake.

```
elf/icf.cc:93:5: warning: 'sprintf' is deprecated: This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead. [-Wdeprecated-declarations]
    sprintf(buf + i * 2, "%02x", digest[i]);
    ^
/Applications/Xcode-14.0.0-Beta.5.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/stdio.h:188:1: note: 'sprintf' has been explicitly marked deprecated here
__deprecated_msg("This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead.")
^
/Applications/Xcode-14.0.0-Beta.5.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/sys/cdefs.h:214:48: note: expanded from macro '__deprecated_msg'
        #define __deprecated_msg(_msg) __attribute__((__deprecated__(_msg)))
                                                      ^
1 warning generated.
```

Seems to be safe to use `snprintf` here with buffer size equal to 2.

**EDIT:** this `to_string` impl appears to be unused, so just remove it.

### DIFF
--- a/elf/icf.cc
+++ b/elf/icf.cc
@@ -86,13 +86,6 @@ template<> struct hash<Digest> {
     return *(int64_t *)&k[0];
   }
 };
-
-std::string to_string(Digest digest) {
-  char buf[HASH_SIZE * 2];
-  for (int i = 0; i < HASH_SIZE; i++)
-    sprintf(buf + i * 2, "%02x", digest[i]);
-  return {buf, sizeof(buf)};
-}
 }
 
 namespace mold::elf {


rui314:
Can you add a `signed-off-by` line to your commit message as instructed in https://github.com/rui314/mold/blob/main/CONTRIBUTING.md?

Alcaro:
That's not gonna work - the buffer size includes the nul terminator, so buffer size 2 will replace every second digit with a nul byte. Use buffer size 3 instead.

And you'll need to increment the buffer size, so the final char's nul fits. And subtract that in the sizeof in the return. Feels like 'convert byte array to hex string' is common enough that it should be in the standard library, but I haven't found one.

nevack:
> Can you add a `signed-off-by` line to your commit message as instructed in

Done!

> That's not gonna work - the buffer size includes the nul terminator, so buffer size 2 will replace every second digit with a nul byte. Use buffer size 3 instead.

3 won't work either, no? Like the last write will overflow buffer.

sicherha:
> 3 won't work either, no? Like the last write will overflow buffer.

Indeed. To fix this, the buffer size would have to be increased by 1 (in line 91), which would then have to be subtracted again in line 94.

sicherha:
Looks okay to me in principle, but your suggested change would introduce a subtle semantic difference: while `"%02x"` prints the hex digits *a-f* in lowercase, your code would print them in uppercase.

But if the function is indeed unused, I agree we should consider removing it.

rui314:
Maybe I added this function for debugging and forgot to remove. If it's unused, you can just remove.

PR #637, opened 2022-08-18
Use portable shebang for python3 in update-git-hash.py

nevack:
As per packagers recommendation here https://github.com/Homebrew/homebrew-core/pull/108346#discussion_r949151369

### DIFF
--- a/update-git-hash.py
+++ b/update-git-hash.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
 #
 # This script updates git-hash.cc so that the file contains the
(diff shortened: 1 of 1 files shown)


rui314:
I think we should remove the dependency to python3 by rewriting this script in CMake. Let me try to do that.

ltratt:
FWIW, I think the dependency on python3 is OK (it's widely installed) provided one uses `#!/usr/bin/env` to locate it (rather than hard-coding a path).

rui314:
Let me merge this so that you guys can compile mold now. Thanks!

PR #639, opened 2022-08-18
Add Rust macOS instructions to README

nickgerace:
Since version 1.4.1 contains alpha support for macOS, it may be a good idea to show Rust users how to try it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ take an absolute path as an argument for `-fuse-ld` though.
 
 Create `.cargo/config.toml` in your project directory with the following:
 
-```
+```toml
 [target.x86_64-unknown-linux-gnu]
 linker = "clang"
 rustflags = ["-C", "link-arg=-fuse-ld=/path/to/mold"]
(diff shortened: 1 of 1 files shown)


rui314:
Please add a `signed-off-by` line to your commit message as instructed by https://github.com/rui314/mold/blob/main/CONTRIBUTING.md.

As to this change, I actually wonder if you can generalize

```
[target.x86_64-unknown-linux-gnu]
```

and

```
[target.aarch64-apple-darwin]
```

lines. Isn't there any way to just pass a linker flag in all targets?

nickgerace:
Oops! I'm sorry. I usually sign my commits. I made this PR from a different machine than usual, my mistake.

As for the change: I believe you can, but those targets can include Windows and the path to `ld64.mold` or `mold` might be different depending on the distribution or OS.

What we might be able to do is something like the following...

```toml
[target.your-linux-or-macOS-target]
linker = "clang"
rustflags = ["-C", "link-arg=-fuse-ld=/path/to/mold"]
```

And then add a comment reminding the user that `ld64.mold` should be specified for building darwin binaries.

Thoughts?

rui314:
OK, I think if we can't generalize it, we should write the actual targets. I want to make it something that you can just copy-n-paste, so replacing it with `target.your-linux-or-darwin-target` doesn't seem like a good idea.

nickgerace:
Thank you! I'll keep an eye out to see if this can be improved in the future too.

PR #640, opened 2022-08-19
Fix `ld64.mold` version flag

ZhongRuoyu:
Currently, `ld64.mold -v` returns a nonzero exit code, with the following output:

```plaintext
mold 1.4.1 (compatible with GNU ld)
mold: fatal: no input files
```

- 6d8dcc9cea037af7b6d938df4fb21945f4299699 corrects this in a similar manner as in `elf/cmdline.cc`.
- ~~In addition, I would like to propose adding a `--version` flag, as present in `ld64.lld` (156aad6354bd90d97e7170015600ddb59a861338).~~
- A test `test/macho/version.sh` is added to cover the above (6d8dcc9cea037af7b6d938df4fb21945f4299699).

### DIFF
--- a/macho/cmdline.cc
+++ b/macho/cmdline.cc
@@ -104,7 +104,7 @@ Options:
   -unexported_symbol <SYMBOL> Export all but a given symbol
   -unexported_symbols_list <FILE>
                               Read a list of unexported symbols from a given file
-  -v, --version               Report version information
+  -v                          Report version information
   -weak_framework <NAME>[,<SUFFIX>]
                               Search for a given framework
   -weak-l<LIB>                Search for a given library)";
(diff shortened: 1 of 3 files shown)


rui314:
`ld64.mold` is intended to be a drop-in replacement for Apple's ld which doesn't have the `--version` option. So I don't see an immediate need to add that flag to mold.

ZhongRuoyu:
> `ld64.mold` is intended to be a drop-in replacement for Apple's ld which doesn't have the `--version` option. So I don't see an immediate need to add that flag to mold.

I see. I was checking the behaviour of `ld64.lld` installed from Homebrew, and observed this inconsistency. Then I'll revert the changes related to `--version`?

rui314:
Yes, please remove `--version` from your changes.

ZhongRuoyu:
Sure, done.

PR #643, opened 2022-08-19
Use /usr/bin/env rather than a hard-coded path to python3.

ltratt:
This works well on platforms that store python3 elsewhere than /usr/bin/ (e.g. on OpenBSD it's stored in /usr/local/bin/).


Alcaro:
Isn't this same as #637?

ltratt:
It is! Closing in favour of #637.

PR #652, opened 2022-08-21
[ELF] implement -Y flag

rawoul:
Please consider supporting the -Y flag, which is supported by the bfd linker (but not gold):

```
       -Y path
           Add path to the default library search path.  This option exists for Solaris compatibility.
```

While this is a compatibility option in bfd, it is still very useful when you want to share a cross-toolchain to build binaries and libraries for different targets, without having to copy the full toolchain to their staging dir and using the --sysroot option.

### DIFF
--- a/elf/cmdline.cc
+++ b/elf/cmdline.cc
@@ -39,6 +39,7 @@ Options:
   -S, --strip-debug           Strip .debug_* sections
   -T FILE, --script FILE      Read linker script
   -X, --discard-locals        Discard temporary local symbols
+  -Y DIR                      Add DIR to default library search path
   -e SYMBOL, --entry SYMBOL   Set program entry point
   -f SHLIB, --auxiliary SHLIB Set DT_AUXILIARY to the specified value
   -h LIBNAME, --soname LIBNAME
@@ -338,6 +339,7 @@ std::vector<std::string> parse_nonpositional_args(Context<E> &ctx) {
   ctx.page_size = E::page_size;
   ctx.arg.color_diagnostics = isatty(STDERR_FILENO);
 
+  std::vector<std::string> default_library_paths;
   bool version_shown = false;
   bool warn_shared_textrel = false;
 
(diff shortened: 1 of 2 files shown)


rui314:
I think we should just handle it as an unknown flag to let people to stop using this old command line flag, if any.

rawoul:
As you wish, thanks for the quick reply. I'm fine maintaining this patch in my tree as I need it, feel free to close the PR.

rui314:
What program do you need the `-Y` flag to compile?

rawoul:
Not a specific program, I use it when cross compiling root filesystems for embedded targets. The standard way of doing this is:

- copy the cross-toolchain from a common location to the embedded target staging build dir.
- build libraries and programs using the cross-toolchain with the `--sysroot` flag pointing to the staging build dir, so that both the toolchain libraries and the built libraries can be resolved, and install those binaries in the staging dir.

But to avoid copying the toolchain you can do (with ld.bfd):

- use the cross-toolchain directly from its common install location, without specifying the `--sysroot` flag (otherwise the toolchain libraries won't be found).
- build libraries and programs with the `-Y` and `-Wl,-rpath-link` flags pointing to the staging build dir and install them there.

Using `-L` instead of `-Y` is not a good idea as it is hard to enforce the staging dir to be added first in the library search path with all the different build systems used by all the target packages (autotools, cmake, meson, etc...).

elf/cmdline.cc
```diff
@@ -998,6 +1002,10 @@ std::vector<std::string> parse_nonpositional_args(Context<E> &ctx) {
     }
   }
 
+  ctx.arg.library_paths.insert(ctx.arg.library_paths.end(),
+          default_library_paths.begin(),
+          default_library_paths.end());
```
david-alvarez-rosa, line 1007: I'm not fully getting what is difference between `-Y` and `-L` options -- at the end, we are just concatenating them.

david-alvarez-rosa: Oh, I see now.  It does make sense, thanks.

david-alvarez-rosa (approved):
LGTM.

Issue #655, opened 2022-08-22
sorry, unimplemented: Thumb-1 'hard-float' VFP ABI for armv6l


marxin:
The following test fails on `armv6l`:

```
[ 1174s] Testing arm32-thumb-interwork ... <stdin>: In function 'foo':
[ 1174s] <stdin>:3:1: sorry, unimplemented: Thumb-1 'hard-float' VFP ABI
[ 1174s] make[1]: *** [Makefile:207: test/elf/arm32-thumb-interwork.sh] Error 1
```

rui314:
Is this the line that failed in your environment?

```
cat <<EOF | $CC -o $t/a.o -c -xc - -mthumb
```

If you pass an empty string as an input to the compiler, does it compile or fail?

marxin:
> Is this the line that failed in your environment?
> 
> ```
> cat <<EOF | $CC -o $t/a.o -c -xc - -mthumb
> ```

Yes.

> 
> If you pass an empty string as an input to the compiler, does it compile or fail?

It's going to fail as the `sorry` message comes from option parsing that we run even for an empty input file.

rui314:
Thanks! Then I think the above fix should work.

PR #658, opened 2022-08-23
armv6l: properly skip arm32-thumb-interwork.sh

marxin:
I was not correct as the 'sorry, unimplemented: Thumb-1 'hard-float' VFP ABI
really comes late in target register allocator. Thus one needs to
assembly a function.

Fixes: #655


rui314:
Thanks!

PR #662, opened 2022-08-23
Fix #661: Restore ld location to $(LIBEXECDIR)/mold/ld

### LINKED ISSUE #661: The location of `{PREFIX}/libexec/mold/ld` changed to `{PREFIX}/libexec/ld`
The install target of mold changed the location of `ld`, this looks to be caused by 49ec4671fb48aadc7ffedb96bf89c62bcd590a2e and seems like an accidental change.

ideasman42:


rui314:
Thanks!

PR #668, opened 2022-08-27
[RISCV32] Fix R_RISCV_HI20 relocations

ksco:
Fixed the `exclude-libs3` test case with the following command line:
```
make test-arch TRIPLE=riscv32-unknown-elf MACHINE=riscv32
```

> I'm still learning how linkers work in general as well as the codebase itself, please forgive me if this is the wrong way to fix it.

### DIFF
--- a/elf/arch-riscv.cc
+++ b/elf/arch-riscv.cc
@@ -565,13 +565,11 @@ void InputSection<E>::scan_relocations(Context<E> &ctx) {
 
     switch (rel.r_type) {
     case R_RISCV_32:
+    case R_RISCV_HI20:
       if constexpr (sizeof(Word<E>) == 8)
         handle_abs_rel(ctx, sym, rel);
       else
         handle_abs_dyn_rel(ctx, sym, rel);
-      break;
-    case R_RISCV_HI20:
-      handle_abs_rel(ctx, sym, rel);
       break;
     case R_RISCV_64:
       if constexpr (sizeof(Word<E>) == 4)


rui314:
I believe this is a correct fix, but I wonder how you verified it. Is there any RV32 computer and a Linux distribution I can easily get?

ksco:
I'm currently using Ubuntu 20.04 in Docker. And I installed riscv-gnu-toolchain with the following instruction:

```
./configure --prefix=/opt/riscv --with-arch=rv32gc --with-abi=ilp32d
make -j$(nproc)
ln -sf /opt/riscv/bin/riscv32-unknown-elf-g++ /usr/bin/riscv32-unknown-elf-g++
ln -sf /opt/riscv/bin/riscv32-unknown-elf-gcc /usr/bin/riscv32-unknown-elf-gcc
```

Then I just run `make test-arch TRIPLE=riscv32-unknown-elf MACHINE=riscv32`

With this patch, it now outputs:

```
TEST_CC=riscv32-unknown-elf-gcc \
TEST_CXX=riscv32-unknown-elf-g++ \
TEST_GCC=riscv32-unknown-elf-gcc \
TEST_GXX=riscv32-unknown-elf-g++ \
OBJDUMP=riscv32-unknown-elf-objdump \
MACHINE=riscv32 \
QEMU="qemu-riscv32 -L /usr/riscv32-unknown-elf" \
make test
make[1]: Entering directory '/mold'
cmake -DSOURCE_DIR=. -DOUTPUT_FILE=out/git-hash.cc -P update-git-hash.cmake
Testing sysroot-linker-script ... OK
Testing exclude-libs3 ... OK
Testing stdout ... OK
Testing warn-shared-textrel ... skipped
Testing common-archive ... OK
Testing exception-mcmodel-large ... skipped
Testing gnu-hash ... OK
Testing large-alignment ... OK
Testing tls-gd-mcmodel-large ... skipped
Testing unique ... skipped
Testing z-text ... skipped
Testing gnu-linkonce ... skipped
Testing dead-debug-sections ... skipped
Testing nostdlib ... OK
Testing version-script7 ... OK
Testing init ... make[2]: *** [Makefile:212: test/elf/init.sh] Error 1
make[1]: *** [Makefile:174: test] Error 2
make[1]: Leaving directory '/mold'
make: *** [Makefile:184: test-arch] Error 2
```

rui314:
Thanks. Our RV32 support must be buggy since I actually didn't test at all. It's nice to know at least a few tests passed in your environment.

rui314:
On second thought, I don't think this is not a correct fix. We can't emit a R_RISCV_32 dynamic relocation for R_RISCV_HI20 relocation. If we do, the dynamic loader will override an entire 32 bit word with an address of a specified symbol. HI20 is expected to set a high 20 bits as an instruction operand, so they are not compatible.

ksco:
Oh, that makes sense, I'm looking for other ways to fix this.

ksco:
This seems related to the newlib version of GCC (`riscv32-unknown-elf-`), it's not working with shared libraries. It seems necessary to use the linux version of GCC (`riscv32-linux-gnu-`) just like other archs.

I found out that [Linux 5.19 supports running RV32 binaries](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=35b51afd23c98e2f055ac563aca36173a12588b9), I will try to build a dev environment based on this.

rui314:
You don't have to user a specific version of the Linux kernel as long as you are using qemu-user, no? I think everything is emulated in that environment, so the kernel version of the host machine didn't matter.

ksco:
You are right, I will try using riscv32-linux-gnu-gcc and qemu-user to test. Thanks for correcting me.

PR #670, opened 2022-08-27
[RV32] Skip all test cases not applicable to RV32

ksco:
Along with https://github.com/rui314/mold/pull/669, all test cases for RV32 are passed.

### DIFF
--- a/test/elf/discard.sh
+++ b/test/elf/discard.sh
@@ -12,6 +12,7 @@ echo -n "Testing $testname ... "
 t=out/test/elf/$MACHINE/$testname
 mkdir -p $t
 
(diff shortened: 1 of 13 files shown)


rui314:
I don't think adding `-L/usr/riscv32-linux-gnu` to the command line is the right thing to do. Other targets don't need `-L /usr/target-name`. I think you should compile your compiler so that the path is automatically added to the library search path.

ksco:
OK, I've dropped all relevant changes. Also, adding CI for RV32 in GitHub Action doesn't seem to be very easy since gcc-10-riscv32-linux-gnu package is not available in Ubuntu.

PR #671, opened 2022-08-28
[RV64] Fix jalr assemble error

ksco:
If I understand correctly, the line of code that this patch modifies is trying to take the rd from the `jalr` and put it into the relaxed `jal`. If so, then the `jal` opcode should be `1101111`, not `101111`.

The reason the old code works, I think is because the rd register of the `call rd, func` instruction is usually x1.


rui314:
Thank you for finding this error!

Issue #358, opened 2022-02-22
mold 1.1 failed to find libraries with gcc 11.2.0 on riscv64


xctan:
Currently `gcc` (version 11.2.0) behaves differently on x86-64 and riscv64 platforms. The riscv64 version of `gcc` doesn't pass enough search dirs (such as `/lib`, `/usr/lib`) to mold, resulting in library resolution failure.

Tested on Arch Linux RISC-V:
```
[root@archriscv2 tmp]# uname -a
Linux archriscv2 5.16.9-arch1-1 #1 SMP PREEMPT Fri, 11 Feb 2022 22:42:06 +0000 riscv64 GNU/Linux
[root@archriscv2 tmp]# pacman -Qe gcc
gcc 11.2.0-3
[root@archriscv2 tmp]# echo 'int main() { return 0; }' > b.c
[root@archriscv2 tmp]# mold -run gcc -o ./exe b.c --verbose
  ......
LIBRARY_PATH=/usr/lib/gcc/riscv64-unknown-linux-gnu/11.2.0/:/lib/:/usr/lib/
[...]

PR #672, opened 2022-08-28
[RISCV] Relax AUIPC+JALR to C.JAL or C.J

ksco:
This patch completes the R_RISCV_CALL_PLT relaxation function, as it is done in [LLVM lld](https://github.com/llvm/llvm-project/blob/main/lld/ELF/Arch/RISCV.cpp#L554). However, I don't know if there is any case where auipc+jalr can be relaxed to c.j.

### DIFF
--- a/elf/arch-riscv.cc
+++ b/elf/arch-riscv.cc
@@ -45,6 +45,10 @@ static u32 cjtype(u32 val) {
          bit(val, 1)  << 3  | bit(val, 5)  << 2;
 }
 
+static u32 rd_reg(u32 val) {
+  return bits(val, 11, 7);
+}
+
 static void write_itype(u8 *loc, u32 val) {
   u32 mask = 0b000000'00000'11111'111'11111'1111111;
   *(ul32 *)loc = (*(ul32 *)loc & mask) | itype(val);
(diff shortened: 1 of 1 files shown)


ksco:
I forgot to add Signed-off-by, force-pushed.

rui314:
I think it's a bit risky to work on optional relaxations at the moment. We should focus on making RV32 work first, shouldn't we?

ksco:
RV32 has been adapted to existing tests, just like RV64. But I haven't tried compiling any serious programs, and yes, I think you're right.

So, what should I do next if I want to make RV32 production-ready?

ksco:
Btw, `auipc+jalr -> c.j` can happens in tail call optimization.

rui314:
Do all tests pass? If so, the next step would be to set up a CI.

ksco:
```
sysroot-linker-script.sh
sysroot.sh
sysroot2.sh
```

These still fail due to [this](https://github.com/rui314/mold/pull/670#issuecomment-1229282925), but I think it has something to do with GCC, not mold. It seems that this issue is similar to https://github.com/rui314/mold/issues/358. riscv64-gcc has fixed it, but it seems riscv32-gcc still has it.

rui314:
That's a great news! I didn't know that our RV32 support works that well. Did you have any idea as to how to set up a CI?

ksco:
Yeah, I didn't expect it to be this smooth at first! 😄

Currently, Ubuntu does not provide the `gcc-10-riscv32-linux-gnu` package. I think we can use pre-built binaries from riscv-gnu-toolchain like [this one](https://github.com/riscv-collab/riscv-gnu-toolchain/releases/tag/2022.08.26#:~:text=riscv32%2Dglibc%2Dubuntu%2D20.04%2Dnightly%2D2022.08.26%2Dnightly.tar.gz).

But the sysroot-related cases don't seem to pass the test, unfortunately.

rui314:
How about writing a script like this and place it in `$PATH` as riscv32-linux-gnu-gcc?

```
#!/bin/bash
/absolute/path/to/riscv32-linux-unknown-gcc -L /usr/riscv32-linux-gnu "$@"
```

rui314:
You may also want to add a new CMake option `MOLD_ENABLE_RV32_QEMU_TESTS` to enable tests with qemu-riscv32. By default, we don't want to run them even if `MOLD_ENABLE_QEMU_TESTS` is `ON`, as they don't run correctly in most environments at the moment.

ksco:
> How about writing a script like this and place it in `$PATH` as riscv32-linux-gnu-gcc?
> 
> ```
> #!/bin/bash
> /absolute/path/to/riscv32-linux-unknown-gcc -L /usr/riscv32-linux-gnu "$@"
> ```

Looks great to me ;)

ksco:
> You may also want to add a new CMake option `MOLD_ENABLE_RV32_QEMU_TESTS` to enable tests with qemu-riscv32. By default, we don't want to run them even if `MOLD_ENABLE_QEMU_TESTS` is `ON`, as they don't run correctly in most environments at the moment.

Okay, I'll do that in a separate PR.

ksco:
Hi, Rui,

Do you have time to add RV32 to CI? I can also try to do this.

rui314:
I added RV32 to CI, and all the tests are green!

ksco:
🎉

rui314:
Is this true you can relax `AUIPC+JALR` to `C.J`? `C.J` is just a jump instruction and not jump-and-link-register instruction, no?

rui314:
My bad, `C.J` is defined as `jal x0, offset[11:1]`.

ksco:
It can happen with tail call optimization. I'm going to write a test case to cover this situation, but I'm not quite sure if this test is necessary for ARM and X86.

rui314:
At this point you probably want to try to link a large program like mold itself or even gcc using mold. It's amazing that we get to this point so quickly.

PR #673, opened 2022-08-28
Add MOLD_ENABLE_RV32_QEMU_TESTS option

ksco:


rui314:
I don't think we need to add it to Makefile, but this is fine too.

PR #681, opened 2022-09-01
[RISCV] Check rvc support when relaxing R_RISCV_CALL_PLT

ksco:


ksco:
Relevant description in spec

> EF_RISCV_RVC (0x0001)
> This bit is set when the binary targets the C ABI, which allows instructions to be aligned to 16-bit boundaries (the base RV32 and RV64 ISAs only allow 32-bit instruction alignment). When linking objects which specify EF_RISCV_RVC, the linker is permitted to use RVC instructions such as C.JAL in the linker relaxation process.

PR #686, opened 2022-09-02
Fix build error while using the system version of mimalloc and CMake

ericriff:
Currently we get the following error:
```
CMake Error at source_subfolder/CMakeLists.txt:82 (target_compile_definitions):
  Cannot specify compile definitions for target "mimalloc" which is not built
  by this project.
```
This is because we're trying to add definitions to a target which is not built by this project. As a workaround we set the definition on the `mold` target, privately since it is only used on `main.cc`.

BTW: There are no docs on how to build with CMake. Would you consider CMake to be production ready or experimental?


rui314:
We just added CMakeLists.txt in the previous release cycle, so it's somewhat experimental. Our official doc still recommends users use Make. We need to add a documentation for CMake though.

Issue #461, opened 2022-04-25
RV64: Implement TLS relaxation


rui314:
mold currently doesn't relax any TLS relocations at all. We should implement GD → LE and LD → LE relaxations just like we did to x86-64.

ksco:
Hi Rui,

I would like to implement this but it may take some time as I am not familiar with the ELF TLS model.

rui314:
Sure. I'll assign this bug to you to avoid duplicate effort. Feel free to ask any questions.

ksco:
Firstly, I'm trying to open all the `tls-*` tests for RV32 and RV64 and everything works fine except for one issue.

In [tls-gd.sh](https://github.com/rui314/mold/blob/main/test/elf/tls-gd.sh), in the end, we are trying to statically link the three PIC object files together. In this very case, the output becomes 0 2 0 0 0 0 instead of the expected 1 2 3 4 5 6, I haven't found the reason yet.

https://github.com/ksco/mold/commit/06a2daa19df8c406a2dd1eba52fdbc6523c8ad52#diff-8d8a35fafde296bbdfcf817940e5559dc0fb2a1bc4aaea5822870253cdbaac74R68

rui314:
TLS is super tricky. I don't know what's wrong, let me take a look.

rui314:
I fixed the issue in the above commit.

ksco:
That's amazing!

ksco:
I think this is done since we already handled the Thread-pointer Relaxation, right?

rui314:
By TLS relaxation, I meant conversion of General Dynamic → Local Exec and Local Dynamic → Local Exec, and not the RISC-V-specific %tp optimization. But I found that the RISC-V psABI lacks a mechanism to allow GD → LD (https://github.com/riscv-non-isa/riscv-elf-psabi-doc/issues/308), so maybe this is infeasible. It's a spec bug, though.

ksco:
Oh, seems they're gonna add this mechanism after spec 1.0 is released, so I think I'll just wait till that happens.

rui314:
TLSDESC for RISC-V has now been ratified, and we've implement it to mold. I think we can close this issue now.

Issue #460, opened 2022-04-25
RV64: Implement relaxation against __global_pointer$


rui314:
On RV64, the register `gp` always points to .sdata+0x800 to allow a linker to relax global variable accesses if they are within [.sdata, .sdata+0x1000). mold does not implement this relaxation yet.

PR #687, opened 2022-09-02
[RISCV] Add zero-page relaxation 

ksco:
Currently, this patch fails at tests with -static option and needs more work to do before merging.

Also, seems TLS relaxation(https://github.com/rui314/mold/issues/461) is already supported. We only have global-pointer relaxation(https://github.com/rui314/mold/issues/460) left to do for RISC-V.

### DIFF
--- a/elf/arch-riscv.cc
+++ b/elf/arch-riscv.cc
@@ -75,12 +75,12 @@ static void write_jtype(u8 *loc, u32 val) {
 }
 
 static void write_cbtype(u8 *loc, u32 val) {
-  u32 mask = 0b1110001110000011;
+  u32 mask = 0b111'000'111'00000'11;
   *(ul16 *)loc = (*(ul16 *)loc & mask) | cbtype(val);
 }
 
 static void write_cjtype(u8 *loc, u32 val) {
-  u32 mask = 0b1110000000000011;
+  u32 mask = 0b111'00000000000'11;
   *(ul16 *)loc = (*(ul16 *)loc & mask) | cjtype(val);
 }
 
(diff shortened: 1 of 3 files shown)


rui314:
Generally looking good, but unlike `R_RISCV_CALL_PLT`, data may be referenced by HI20+LUI, so we need to handle section fragments when computing target addresses, no?

ksco:
Thanks, the newly pushed code handles section fragments.

elf/arch-riscv.cc
```diff
@@ -812,6 +841,43 @@ static void shrink_section(Context<E> &ctx, InputSection<E> &isec, bool use_rvc)
...
     }
+    case R_RISCV_HI20: {
+      // This relocation refers to an LUI instruction containing the high
+      // 20-bits to be relocated to an absolute symbol address.
+      // In some cases, LUI can be removed or replaced by C.LUI.
+
+      // If it's a static linking, we skip this relaxation for now, as it's
+      // currently not working.
+      if (ctx.arg.is_static)
+        break;
```
ksco, line 852: I still don't understand why it fails in static linking. Looks like we accidentally removed some LUI instructions that shouldn't have been removed.

ksco:
Also, the LUI -> C.LUI relaxation seems too restrictive and I don't know under what circumstances it would be triggered. For example, how do we write tests for this relaxation?

elf/arch-riscv.cc
```diff
@@ -812,6 +841,43 @@ static void shrink_section(Context<E> &ctx, InputSection<E> &isec, bool use_rvc)
...
+
+      const SectionFragmentRef<E> *frag_ref = nullptr;
+      if (isec.rel_fragments && isec.rel_fragments[frag_idx].idx == i)
+        frag_ref = &isec.rel_fragments[frag_idx++];
```
rui314, line 856: You shouldn't write this inside a switch-case because frag_idx would be increment only when rel is R_RISCV_HI20. We should increment it regardless of relocation type.

rui314:
> Also, the LUI -> C.LUI relaxation seems too restrictive and I don't know under what circumstances it would be triggered. For example, how do we write tests for this relaxation?

I believe you can place a function in its own section using `__attribute__((section(".low")))` and place it at the very beginning of the virtual address space with `-Wl,--section-start=.low=0x16`.

I don't think LUI → C.LUI is useful on Unix, but it might be in embedded programming.

elf/arch-riscv.cc
ksco: Thanks, done.

ksco:
> I believe you can place a function in its own section using `__attribute__((section(".low")))`...

Thanks, I added a c.lui test, but it segfaulted under riscv64, but I believe it's unrelated to this patch.

rui314:
Maybe the first few pages are kept unmapped to catch null pointer accesses?

There might be no way to test this on Unix. If you wrote a bootloader-ish program, you could test it, but I doubt if it's necessary. I wouldn't implement LUI → C.LUI relaxation in the first place as it doesn't seem useful.

ksco:
I created a separate bug https://github.com/rui314/mold/issues/688.

ksco:
> I wouldn't implement LUI → C.LUI relaxation in the first place as it doesn't seem useful.

Agreed, I realized this after implementing, maybe I just remove this relaxation then.

ksco:
@rui314 Thanks for the help, I think now it's ready for review.

PR #691, opened 2022-09-04
Add ENTRY command

ksco:


rui314:
I want to keep the set of supported linker script features as minimum as possible, as we'll eventually implement something that compete with it rather than trying to be 100% compatible with it. So could you explain a little bit about your motivation to add this directive to our implementation?

ksco:
Ah, I'm just trying to do something random and get familiar with the codebase. The ENTRY directive seems simple enough to add. I don't have a strong will to add this directive.

rui314:
Thanks for the explanation. If that's the case, let's not add the feature to our codebase.

Issue #268, opened 2022-01-10
Object merging with linker script produces object file with no symbols


mpickering:
It appears that object merging when using a linker script produces object files with no symbols.

```
> tree .
.
├── ghc_14.ldscript
├── input.o
├── lib.o
├── mold-cmd
└── out.dyn_o
```

```
> cat ghc_14.ldscript
INPUT("lib.o" "input.o")
```

```
[nix-shell:~/ghc-scratch/mold-bug]$ nm lib.o 
                 U registerForeignExports
0000000000000000 b stg_exports_GHCziCString
0000000000000000 t stginit_export_GHCziCString

[nix-shell:~/ghc-scratch/mold-bug]$ nm input.o 
0000000000000000 D ghczmprim_GHCziCString_Bytes_closure_tbl
0000000000000230 D ghczmprim_GHCziCString_cstringLengthzh_closure
0000000000000fb8 T ghczmprim_GHCziCString_cstringLengthzh_info
0000000000000228 D ghczmprim_GHCziCString_czustrlen_closure
0000000000000f88 T ghczmprim_GHCziCString_czustrlen_info
0000000000000250 D ghczmprim_GHCziCString_Four_closure
0000000000001038 T ghczmprim_GHCziCString_Four_con_info
0000000000000238 D ghczmprim_GHCziCString_One_closure
0000000000000fd8 T ghczmprim_GHCziCString_One_con_info
0000000000000248 D ghczmprim_GHCziCString_Three_closure
0000000000001018 T ghczmprim_GHCziCString_Three_con_info
0000000000000240 D ghczmprim_GHCziCString_Two_closure
0000000000000ff8 T ghczmprim_GHCziCString_Two_con_info
0000000000000028 D ghczmprim_GHCziCString_unpackAppendCStringUtf8zh_closure
00000000000009d0 T ghczmprim_GHCziCString_unpackAppendCStringUtf8zh_info
0000000000000008 D ghczmprim_GHCziCString_unpackAppendCStringzh_closure
0000000000000180 T ghczmprim_GHCziCString_unpackAppendCStringzh_info
0000000000000030 D ghczmprim_GHCziCString_unpackCStringUtf8zh_closure
0000000000000d40 T ghczmprim_GHCziCString_unpackCStringUtf8zh_info
0000000000000000 D ghczmprim_GHCziCString_unpackCStringzh_closure
0000000000000058 T ghczmprim_GHCziCString_unpackCStringzh_info
[...]

PR #692, opened 2022-09-04
[ELF] Add linker script support to --relocatable

ksco:
This patch doesn't solve #268, but at least made some progress.

### DIFF
--- a/elf/linker-script.cc
+++ b/elf/linker-script.cc
@@ -187,14 +187,15 @@ static MappedFile<Context<E>> *resolve_path(Context<E> &ctx, std::string_view to
 
 template <typename E>
 static std::span<std::string_view>
-read_group(Context<E> &ctx, std::span<std::string_view> tok) {
+read_group(Context<E> &ctx, std::span<std::string_view> tok,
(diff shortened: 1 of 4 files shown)


ksco:
Force-pushed to make clang happy.

rui314:
This is again a bit controversial. Our implementation for `-r` is intentionally separated from the main execution passes because `-r` is a very different operation mode than outputting an executable or a shared library. And our linker script implementation is also limited so that the mold is simple and fast. I think we don't want to add more code there unless it is essential.

In this case, they could just pass a list of files instead of a linker script. If they wanted to pass a list of files as a file, they could also use a response file. So I don't think the need is essential.

ksco:
Thanks for the explanation, it makes sense to me. I’ll close this PR now.

rui314:
If you are looking for something to do in this area, you might want to try to build an OS kernel with mold. I don't think you can build the Linux kernel with mold, but for smaller OSes, `--section-start` might be enough.

We want to implement something better than the linker script, but before doing that, we need to understand the existing uses of the script. Trying to link a kernel should give us insights.

ksco:
Thanks for the suggestion, sounds good, I will try to do it.

PR #696, opened 2022-09-07
[RISCV] Add global-pointer relaxation for R_RISCV_HI20

ksco:
Related #460

### DIFF
--- a/elf/arch-riscv.cc
+++ b/elf/arch-riscv.cc
@@ -333,23 +333,33 @@ void InputSection<E>::apply_reloc_alloc(Context<E> &ctx, u8 *base) {
         write_utype(loc, val);
       } else {
         assert(removed_bytes == 4);
-        assert(sign_extend(val, 11) == val);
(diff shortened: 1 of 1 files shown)


rui314:
Looks good but I think we need a test for this relaxation.

ksco:
Sure, let's keep this PR open until I added a test for it.

rui314:
Not specific to this particular change, but I'm genuinely curious how much useful the %gp optimization is. It might be useful for small embedded programs, but for usual Unix programs, is it really worth to reserve a register just for enabling this optimization?

ksco:
Yes, it is useful for bare metal programs, but even in a non-embedded environment, the gp register is always occupied and not changed after initialization. Maybe just for the sake of simplicity?

ksco:
This patch is not right, `__global_pointer` is a linker-synthesized symbol, we don't know its value until `fix_synthetic_symbols` is called.

PR #701, opened 2022-09-10
Remove 'Ignored' helptext for lto_library and order_file

rod-salazar:
I see that these 2 flags are implemented and not ignored, albeit potentially incomplete (order file code does not handle colon delimiter yet for example).

If the intent was to leave these as undocumented/Ignored for the time being, you can ignore this pull request :)


rui314:
Thanks! Long descriptions should be written to the man page (which doesn't yet), so I'll update the help text after merging your commit.

Issue #647, opened 2022-08-20
[macOS Apple Silicon] Add support for -undefined dynamic_lookup


nsabovic:
When compiling [napi-rs](https://napi.rs) (Node.js addons in Rust) the Rust compilation fails with

```
note: mold: fatal: unknown command line option: -undefined
```

The relevant flags are `"-fuse-ld=/opt/homebrew/bin/ld64.mold" "-Wl" "-undefined" "dynamic_lookup"`.

pan93412:
I also encountered that, and seems like Mold has not been supported `-undefined dynamic_lookup` in their codebase. Hopefully it can be fixed so I can apply `mold` to my projects ;)

kiufta:
I've got this issue with Paddle: https://github.com/PaddlePaddle/Paddle/issues/45761

rui314:
Thank you guys for reporting. Let me take a look.

PR #706, opened 2022-09-12
[Mach-O] Add -undefined dynamic_lookup option

ksco:
Related #647


ksco:
Hi @rui314, is it possible for this to be merged?

rui314:
I think we want to make `sym->is_imported` true if an undefined symbol is promoted to a dynamic symbol, just like what we are doing for ELF.

PR #711, opened 2022-09-14
tests: come up with common base.sh

marxin:
All tests have a common start, factor it out.

### DIFF
--- a/test/elf/base.sh
+++ b/test/elf/base.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+export LC_ALL=C
+set -e
+CC="${TEST_CC:-cc}"
+CXX="${TEST_CXX:-c++}"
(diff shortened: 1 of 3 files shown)


marxin:
Would you be interested in a patch that would do the described refactoring?

rui314:
I kind of like the current way of doing this for a few reasons:

 - it's self-contained, and
 - you can't write lots of code in the boilerplate, so it's guaranteed to be simple.

But I do see the benefit of creating a `base.sh` file, too. Let me think about it.

marxin:
Sure, think about it.

aminya (approved):

rui314:
I did it in e1275942c8271bafd5e36eeb502295d853d092b8.

marxin:
Oh, great!

PR #712, opened 2022-09-17
Don't use which(1) or perl

wangp:
`which` is not always installed.
I also noticed `perl` was used in a single test, that could be replaced with `sed`.


joshcangit:
I agree. `which` is deprecated since debianutils (5.0-1).
That could mean Debian maintainers might remove `which` in the future.
I wonder if it is better to use `command -v` or `type -P`?
Both seem the same to me.

rui314:
Thank you for the PR! I didn't know that `which` isn't a POSIX command.

PR #713, opened 2022-09-18
Convert hidden global symbols into local ones.

### LINKED ISSUE #528: Hidden weak undefined symbol generates dynamic relocation in shared library
Here is a simple reproducer:
```cpp
// foo.cpp
[[gnu::weak]] [[gnu::visibility("hidden")]] void foo();

int bar() {
	if (foo) {
		foo();
		return 0;
	} else {
		return 1;
	}
}
```
Then compile with
```
$ clang++ foo.cpp -mno-relax --target=riscv64-unknown-linux-none -Wall -c -o foo.o
```
and link with
```
$ mold foo.o -shared -o foo.so
```

Then we have:
```
$ llvm-readelf -r foo.so

Relocation section '.rela.dyn' at offset 0x290 contains 1 entries:
    Offset             Info             Type               Symbol's Value  Symbol's Name + Addend
0000000000003000  0000000100000002 R_RISCV_64             0000000000000000 _Z3foov + 0
```

This behaviour differs to `lld` and GNU `ld`, e.g:
```
$ ld.lld foo.o -shared -o foo.so
$ llvm-readelf -r foo.so

There are no relocations in this file.
```

Mold version used:
```
$ ./mold -v
mold 1.2.1 (826bb56d59312bd8a4ad3fc1d60aa7b430cee94b; compatible with GNU ld)
```

rbartlensky:
I decided to add an extra pass for this since it makes sense to do it after all symbols have been processed, but before scanning relocations.

Fixes #528.

```
$ clang++ test.cc -mno-relax --target=riscv64-unknown-linux-none -Wall -c -o foo.o
$ ../mold foo.o -shared -o foo.so
$ readelf -r foo.so

There are no relocations in this file.
```

### DIFF
--- a/test/elf/hidden-conversion.sh
+++ b/test/elf/hidden-conversion.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+export LC_ALL=C
+set -e
+CC="${TEST_CC:-cc}"
+CXX="${TEST_CXX:-c++}"
+GCC="${TEST_GCC:-gcc}"
+GXX="${TEST_GXX:-g++}"
+OBJDUMP="${OBJDUMP:-objdump}"
+MACHINE="${MACHINE:-$(uname -m)}"
+testname=$(basename "$0" .sh)
+echo -n "Testing $testname ... "
+t=out/test/elf/$MACHINE/$testname
+mkdir -p $t
+
+cat <<EOF | $CC -o $t/foo.o -fPIC -c -xc -
+__attribute__((visibility("hidden"))) void foo();
+int main() { foo(); }
+EOF
+
+./mold -shared -o $t/foo.so $t/foo.o
+readelf --symbols $t/foo.so > $t/log
+grep -q '.*LOCAL  DEFAULT  UND foo' $t/log
+
+echo OK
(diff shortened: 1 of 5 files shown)


test/elf/hidden-conversion.sh
```diff
@@ -0,0 +1,24 @@
...
+OBJDUMP="${OBJDUMP:-objdump}"
+MACHINE="${MACHINE:-$(uname -m)}"
+testname=$(basename "$0" .sh)
+echo -n "Testing $testname ... "
+t=out/test/elf/$MACHINE/$testname
+mkdir -p $t
+
+cat <<EOF | $CC -o $t/foo.o -fPIC -c -xc -
+__attribute__((visibility("hidden"))) void foo();
+int main() { foo(); }
```
rui314, line 17: I think this should actually cause an undefined symbol error, no? `foo` is not weak.

rui314:
I think `ObjectFile<E>::claim_unresolved_symbols` is the one that promotes undefined symbols to dynamic symbols, and we need to add a new check there. It's going to be a single-line change, so let me do that.

rbartlensky:
@rui314 thanks for looking over my PR. Indeed your changes seems to do the trick :)

PR #717, opened 2022-09-18
tests: unify usage of '[ test ]'

marxin:
Do not use double square brackets.


rui314:
`[[` is a bash command that interprets given strings as glob patterns, while `[` is an external command that just does string comparison, so I don't think they are compatible.

marxin:
Oh, I see!

PR #720, opened 2022-09-19
[ELF] Simplify

ksco:


ksco:
Oh, I didn't notice this part was just changed.

rui314:
We happened to be editing the same line!

Issue #427, opened 2022-04-13
Address of nested function crashes with mold


marxin:
The following crashes:

```
$ cat nested.c
void callme (void (*callback) (void));

int
main (void)
{
  int ok = 0;
  void callback (void) { ok = 1; }

  callme (&callback);

  if (!ok)
   __builtin_abort ();
  return 0;
}

__attribute__((noinline, noclone))
void
callme (void (*callback) (void))
{
  (*callback) ();
}
$ gcc nested.c && ./a.out
Segmentation fault (core dumped)
```

rui314:
GCC's nested function is implemented by dynamically constructing stub code in the stack area to jump from the stack to an actual function. So the stack area must be executable to use it, which is very rare nowadays for the obvious security reason.

mold unconditionally makes the stack area non-executable. You need to pass `-z execstack` to mold to make it executable.

GNU linker makes the stack area executable if at least one input object file has a section `.note.GNU-stack` with `SHF_EXECINSTR`. GCC emits such section if the code uses nested functions.

To me, the GNU linker's behavior seems dangerous. If one person accidentally uses a nested function at somewhere in the entire codebase, or if you statically-link a library that uses a nested function, it makes your program's entire stack area executable without any notice. That would be a big surprise to users, and it feels like that's planting a time bomb waiting to explode.

I think this feature should be re-implemented with an assumption that the stack area is not executable, or if you can't, it should be deprecated because it's dangerous to use.

marxin:
Well, I think various projects are leaving usage of nested functions and Clang does not support them if I'm correct. However, GCC still emits them and e.g. Fortran code contains quite many of them.

I respect your design decision, but on the other hand, you claim `mold` is a GNU-compatible linker. So, would it be possible to add a build option that would allow GNU behavior w/o need to pass `-z execstack`?

rui314:
Currently, there's no compile-time options to customize mold's run-time behavior, and that helps me a lot to support users. As long as mold binaries are built at a specific git commit, they are guaranteed to behave the same (modulo a possible compiler bug). In addition to that, I don't want to make `-z execstack` implicit for the reason as explained above. So, I don't think that that's a right decision to make it configurable.

I committed d0e4eee204ad14191b219313d0d4e129b91e5bcc so that users can change their code for mold. Isn't this enough?

This is security-related. It's hard to make a decision to make it vulnerable to a simple buffer overflow attack.

marxin:
[...]

PR #725, opened 2022-09-22
list supported debug section compression algorithms

marxin:
The feature would be useful for e.g. GCC where we will add a configure check for `--compress-debug-sections=zstd` based on mold -V.

### DIFF
--- a/elf/cmdline.cc
+++ b/elf/cmdline.cc
@@ -194,7 +194,8 @@ Options:
     -z textoff
 
 mold: supported targets: elf32-i386 elf64-x86-64 elf32-littlearm elf64-littleaarch64 elf32-littleriscv elf64-littleriscv elf64-sparc
-mold: supported emulations: elf_i386 elf_x86_64 armelf_linux_eabi aarch64linux aarch64elf elf32lriscv elf64lriscv elf64_sparc)";
+mold: supported emulations: elf_i386 elf_x86_64 armelf_linux_eabi aarch64linux aarch64elf elf32lriscv elf64lriscv elf64_sparc
+mold: supported debug section compressions: zlib zlib-gabi zstd)";
 
 static std::vector<std::string> add_dashes(std::string name) {
   // Single-letter option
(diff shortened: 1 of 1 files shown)


rui314:
This is OK as long as we are adding the same message to GNU ld. Was that proposed?

rui314:
I also wonder if we need to list `zlib-gabi`. It's just an alias for `zlib`.

marxin:
After rethinking, I'm going to make the detection from `$ld --help`.

rui314:
By the way, what do you think about adding a `./configure`-time option to GCC to use `ld.mold` by default instead of `ld`? Currently, in order to mold, we need to either modify each project's build file (which is bothersome) or replace `/usr/bin/ld` entirely (which is scary). I think it's nice if we can make gcc to use `ld.mold` by default unless it is told to do otherwise with `-fuse-ld=bfd`.

marxin:
> By the way, what do you think about adding a `./configure`-time option to GCC to use `ld.mold` by default instead of `ld`? 

It's definitely something you/we should discuss at `gcc@gcc.gnu.org` mailing list. But generally speaking, we (as GNU toolchain) prefer our linker (BFD, GOLD is obsolete). Anyway, please start a discussion.

One discrepancy between mold and ld.bfd is that you don't emit executable stack when stack note is missing:
https://github.com/rui314/mold/issues/427#issuecomment-1102531554

`ld.bfd` newly emits warning, but the behavior is unchanged.

Currently, in order to mold, we need to either modify each project's build file (which is bothersome) or replace `/usr/bin/ld` entirely (which is scary). I think it's nice if we can make gcc to use `ld.mold` by default unless it is told to do otherwise with `-fuse-ld=bfd`.

Well, distros have a nice mechanism and it's `update-alternative`, which I use:
```
$ update-alternatives --set ld /usr/bin/ld.mold
$ ld --version
mold 1.4.2 (compatible with GNU ld)
```

PR #727, opened 2022-09-22
Enable compress-debug-sections-zstd for GCC

marxin:
The compression option is directly passed to linker, so GCC can handle it as well.


marxin:
All right.

PR #738, opened 2022-09-27
cmake: use unambiguously absolute paths for install directories

jpalus:
as per https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html CMAKE_INSTALL_<dir> can be both relative and absolute. so far mold made assumption these are relative only, but handle both cases by using CMAKE_INSTALL_FULL_<dir> instead.

Signed-off-by: Jan Palus <jpalus@fastmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@ endif()
 add_executable(mold)
 target_compile_features(mold PRIVATE cxx_std_20)
 target_compile_definitions(mold PRIVATE
-  "LIBDIR=\"${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}\"")
+  "LIBDIR=\"${CMAKE_INSTALL_FULL_LIBDIR}\"")
 target_link_libraries(mold PRIVATE ${CMAKE_DL_LIBS})
 
 if(NOT "${CMAKE_CXX_COMPILER_FRONTEND_VARIANT}" STREQUAL "MSVC")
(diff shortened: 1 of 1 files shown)


rui314:
With what cmake option can I see a difference of CMAKE_INSTALL and CMAKE_INSTALL_FULL_?

jpalus:
Let me explain on simplified CMake file:
```cmake
project(test NONE)

include(GNUInstallDirs)
message("CMAKE_INSTALL_LIBDIR: ${CMAKE_INSTALL_LIBDIR}")
message("CMAKE_INSTALL_FULL_LIBDIR: ${CMAKE_INSTALL_FULL_LIBDIR}")
message("CMAKE_INSTALL_PREFIX/CMAKE_INSTALL_LIBDIR: ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}")
```
First two messages show difference between `CMAKE_INSTALL_LIBDIR` and `CMAKE_INSTALL_FULL_LIBDIR`. Third message is mimicking what `mold` does ie here:
https://github.com/rui314/mold/blob/56b9ad9c6275d1b3679154efb91172e8c0008752/CMakeLists.txt#L16

`CMAKE_INSTALL_LIBDIR` can be both relative (in which case it is relative to `CMAKE_INSTALL_PREFIX`) or absolute so let's try both:

**Relative**:
```
$ cmake -B build -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_INSTALL_LIBDIR=lib64 
CMAKE_INSTALL_LIBDIR: lib64
CMAKE_INSTALL_FULL_LIBDIR: /usr/lib64
CMAKE_INSTALL_PREFIX/CMAKE_INSTALL_LIBDIR: /usr/lib64
```

That's all correct and expected.

**Absolute**:

```
$ cmake -B build -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_INSTALL_LIBDIR=/usr/lib64
CMAKE_INSTALL_LIBDIR: /usr/lib64
CMAKE_INSTALL_FULL_LIBDIR: /usr/lib64
CMAKE_INSTALL_PREFIX/CMAKE_INSTALL_LIBDIR: /usr//usr/lib64
```

As you can see last line has `CMAKE_INSTALL_PREFIX` applied even though it shouldn't.

Instead of manually applying `CMAKE_INSTALL_PREFIX` the solution is to use `CMAKE_INSTALL_FULL_*` since it always correctly resolves both relative and absolute paths.

ADKaster:
I thought that common best practice was to always use a path relative to PREFIX for CMAKE_INSTALL_LIBDIR and all the other GnuInstallDirs options. Is there a common distro that packages CMake packages using an absolute path for those options?

jpalus:
I wouldn't really know about _common distro_ but distribution that I'm using (PLD Linux) does. Historically RPM based distributions define macros with common locations like `%_bindir` or `%_libdir` which are absolute and that's what's used in our `%cmake` macro as well, same as equivalent `%configure` macro for autotools. On the contrary I see Fedora is inconsistent here and while they use `%_bindir` or `%_libdir` in `%configure`, their `%cmake` macro picks only a few ignoring global macros:
```
        -DCMAKE_INSTALL_PREFIX:PATH=/usr \
        -DINCLUDE_INSTALL_DIR:PATH=/usr/include \
        -DLIB_INSTALL_DIR:PATH=/usr/lib64 \
        -DSYSCONF_INSTALL_DIR:PATH=/etc \
```

Not sure about best practices in `cmake` but it supports both relative and absolute paths consistently throughout all directives (see ie `install` command https://cmake.org/cmake/help/latest/command/install.html) and issues with absolute paths are extremely rare.

rui314:
I'm not sure if specifying a full path to CMAKE_INSTALL_LIBDIR is a good practice, but at least applying this patch doesn't seems to hurt existing usages, so I'm fine with this.

PR #743, opened 2022-09-29
Detect skipped tests

sicherha:
This allows CTest to differentiate between tests that were executed successfully and tests that were skipped.

The [`SKIP_REGULAR_EXPRESSION`](https://cmake.org/cmake/help/latest/prop_test/SKIP_REGULAR_EXPRESSION.html) property was added in CMake 3.16, so we need conditional `if`-blocks around the places where we set it. Alternatively, we could increase `cmake_minimum_required(VERSION 3.13)` in the main `CMakeLists.txt`.

Side note: Ubuntu 20.04 comes with CMake 3.16.3.

Which would you prefer?

### DIFF
--- a/test/elf/CMakeLists.txt
+++ b/test/elf/CMakeLists.txt
@@ -13,6 +13,12 @@ function(add_qemu_test TEST TRIPLE MACHINE)
       COMMAND bash -x ${CMAKE_CURRENT_LIST_DIR}/${TEST}
       WORKING_DIRECTORY ${mold_BINARY_DIR})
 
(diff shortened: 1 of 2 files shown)


rui314:
I think I want to keep the required Cmake version as minimum as possible so that the transition from Make to Cmake will be smooth.

rui314:
I rolled it back because cmake printed out too many lines for skipped test names which pushes the test summary lines out of the screen. If there's a way to print out only the number of skipped tests, but we don't need to print out names of skipped tests.

PR #745, opened 2022-09-30
Optionally detect skipped tests

sicherha:
Second attempt; I'm making it opt-in this time. 🙂

This allows CTest to differentiate between tests that were executed successfully and tests that were skipped.

Since CTest prints a list of skipped tests at the very end of its output, skipping many tests may cause the overall result (e.g. '100% tests passed, 0 tests failed out of 275') to scroll out of sight. Hence, make this feature opt-in under the flag
`MOLD_SHOW_SKIPPED_TESTS`.

The `SKIP_REGULAR_EXPRESSION` test property was added in CMake 3.16, so make that minimum version a prerequisite for enabling `MOLD_SHOW_SKIPPED_TESTS`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -267,6 +267,13 @@ target_sources(mold PRIVATE
 
 include(CTest)
 
+option(MOLD_SHOW_SKIPPED_TESTS "Print list of skipped tests in the summary" OFF)
+if(MOLD_SHOW_SKIPPED_TESTS AND ${CMAKE_VERSION} VERSION_LESS "3.16")
+  # This feature uses `SKIP_REGULAR_EXPRESSION`, which requires CMake >= 3.16
+  message(WARNING "MOLD_SHOW_SKIPPED_TESTS=ON requires CMake >= 3.16; forcing it to OFF")
+  set(MOLD_SHOW_SKIPPED_TESTS OFF)
+endif()
+
 if(BUILD_TESTING)
   # Create the ld and ld64 symlinks required for testing
   if(NOT WIN32)
--- a/test/elf/CMakeLists.txt
+++ b/test/elf/CMakeLists.txt
@@ -30,6 +30,12 @@ foreach(TEST IN LISTS TESTS)
     COMMAND bash -x ${CMAKE_CURRENT_LIST_DIR}/${TEST}
     WORKING_DIRECTORY ${mold_BINARY_DIR})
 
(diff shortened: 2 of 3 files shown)


sicherha:
The reason I want this feature is because I would like to maximise test coverage during package builds.
For instance, if a test gets skipped because `llvm-readelf` is not available in the build environment, I want to know so I can add `llvm` to the list of build requirements.

rui314:
I hurts you haven't enabled qemu tests. If you do, one skipped test message is amplified to almost 10 lines of messages which in total produces overwhelming amount of messages at end. We don't want it on each test run.

sicherha:
Oh, I see - the QEMU tests are the same scripts but executed with different environment variables. I could remove the hunk in lines 16-21 so only 'native' tests are shown as skipped. Would that be okay for you?

rui314:
It depends on how many messages it will produce. If test environment has all the necessary tools, it should not produce any "skipped" messages because otherwise the message is not actionable.

sicherha:
Here's what I get on my x86_64 box after explicitly setting `-DMOLD_SHOW_SKIPPED_TESTS=ON`:
```
...
274/274 Test #274: x86_64-z-unknown.sh ......................   Passed    0.01 sec

100% tests passed, 0 tests failed out of 274

Total Test time (real) =  27.27 sec

The following tests did not run:
          5 - x86_64-arm32-thumb-interwork.sh (Skipped)
        154 - x86_64-range-extension-thunk-arm32.sh (Skipped)
        169 - x86_64-riscv-norvc.sh (Skipped)
        216 - x86_64-tls-module-base-i386.sh (Skipped)
        220 - x86_64-tlsdesc-arm.sh (Skipped)
```
If the user doesn't override `MOLD_SHOW_SKIPPED_TESTS` at configuration time, then it stays `OFF` by default and the output is uncluttered as before:
```
...
274/274 Test #274: x86_64-z-unknown.sh ......................   Passed    0.01 sec

100% tests passed, 0 tests failed out of 274

Total Test time (real) =  21.36 sec
```

rui314:
I commited 930964ab3899477701911f23291b84f314e92aa4 to reduce the number of skipped tests. I guess that change helps this patch a bit?

rui314:
I committed b36562fc943a8a2fbcb5e950c5e467a58cfc3934 to show skipped tests. Does this work for you?

sicherha:
Thanks, these commits do exactly what I need! Closing this PR.

Issue #748, opened 2022-09-30
mold: fatal: library not found: -l


erneestoc:
`mold` is working fine to compile an iOS app using Xcode, it's failing however to link dependent targets, in this case in particular a test target fails to link with the app giving the following error:

`mold: fatal: library not found: -l/Users/ernestocambuston/DerivedData/ExampleApp/Build/Products/Debug-iphonesimulator/ExampleApp.app/ExampleApp`

I'm looking at the command and all relevant flags, looks like everything in the following commands is supported but I may be wrong. I also made sure the `...Products/Debug-iphonesimulator/ExampleApp.app/ExampleApp` file exists.

> 
> InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
[...]

PR #749, opened 2022-09-30
Makes xcode tests run

erneestoc:
Fixes: https://github.com/rui314/mold/issues/748

It looks like reading a bundle was being treated as a library even though the bundle is emitted as a single object. I don't have technical knowledge of linkers, but I'm trying to learn here. Tests are passing, but I don't know if I should add more. My example Xcode project tests are working fine now too. 😁

For the main project I'm working on, individual test targets are working fine as well but I found another bug I'm gonna file separately.

### DIFF
--- a/macho/main.cc
+++ b/macho/main.cc
@@ -785,6 +785,13 @@ static void read_input_files(Context<E> &ctx, std::span<std::string> args) {
     read_file(ctx, mf);
   };
 
+  auto read_bundle = [&](std::string name) {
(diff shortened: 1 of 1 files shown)


erneestoc:
Also, I may need approval to run this workflow 😁

rui314:
Can you add a `signed-off-by` line as requested by CONTRIBUTING.md?

erneestoc:
Done, looks like forcing the push to correct the commit name stopped the workflow. 🧐

Issue #779, opened 2022-10-11
qtwebengine fails linking on aarch64


jpalus:
`qtwebengine` build fails with `mold 1.5.1`:
```
mold: error: undefined symbol: QtWebEngineCore::JavaScriptDialogController::defaultPrompt() const
>>> referenced by qquickwebenginedialogrequests.cpp
>>>               .obj/qquickwebenginedialogrequests.o:(QQuickWebEngineJavaScriptDialogRequest::QQuickWebEngineJavaScriptDialogRequest(QSharedPointer<QtWebEngineCore::JavaScriptDialogController>, QObject*))>>> referenced by ui_delegates_manager.cpp
>>>               .obj/ui_delegates_manager.o:(QtWebEngineCore::UIDelegatesManager::showDialog(QSharedPointer<QtWebEngineCore::JavaScriptDialogController>))
mold: error: undefined symbol: QtWebEngineCore::JavaScriptDialogController::type() const
>>> referenced by ui_delegates_manager.cpp
>>>               .obj/ui_delegates_manager.o:(QtWebEngineCore::UIDelegatesManager::showDialog(QSharedPointer<QtWebEngineCore::JavaScriptDialogController>))>>> referenced by qquickwebenginedialogrequests.cpp
>>>               .obj/qquickwebenginedialogrequests.o:(QQuickWebEngineJavaScriptDialogRequest::QQuickWebEngineJavaScriptDialogRequest(QSharedPointer<QtWebEngineCore::JavaScriptDialogController>, QObject*))>>> referenced by ui_delegates_manager.cpp
>>>               .obj/ui_delegates_manager.o:(QtWebEngineCore::UIDelegatesManager::showDialog(QSharedPointer<QtWebEngineCore::JavaScriptDialogController>))>>> referenced 1 more times
```

Links to sources for `JavaScriptDialogController`:
https://code.qt.io/cgit/qt/qtwebengine.git/tree/src/core/javascript_dialog_controller.cpp?h=v5.15.11-lts
https://code.qt.io/cgit/qt/qtwebengine.git/tree/src/core/javascript_dialog_controller.h?h=v5.15.11-lts
https://code.qt.io/cgit/qt/qtwebengine.git/tree/src/core/javascript_dialog_controller_p.h?h=v5.15.11-lts

Links to source for `UIDelegatesManager` using above:
[...]

PR #782, opened 2022-10-15
[ELF] Assign the correct SHNDX to symbols killed by ICF.

ishitatsuyuki:
We have been setting the symbol value and size correctly, but the SHNDX was
erroneously set to UND, potentially causing undefined symbol errors. Fix this by
forwarding the appropriate value from the `leader`.

Closes: #779

### DIFF
--- a/elf/mold.h
+++ b/elf/mold.h
@@ -329,6 +329,13 @@ public:
   std::atomic_bool is_visited = false;
 
   // For ICF
+  //
+  // `leader` is the section that this section has been merged with.
+  // Three kind of values are possible:
+  // - `leader == nullptr`: This section was not eligible for ICF.
+  // - `leader == this`: This section was retained.
+  // - `leader != this`: This section was merged with another idential section.
+  //                     Implies `killed_by_icf`.
   InputSection<E> *leader = nullptr;
   u32 icf_idx = -1;
   bool icf_eligible = false;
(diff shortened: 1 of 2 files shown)


ishitatsuyuki:
Draft until the fix is confirmed by #779's reporter.

elf/mold.h
```diff
@@ -329,6 +329,13 @@ class InputSection {
   std::atomic_bool is_visited = false;
 
   // For ICF
+  //
+  // `leader` is the section that this section has been merged with.
+  // Three kind of values are possible:
+  // - `leader == nullptr`: This section was not eligible for ICF.
+  // - `leader == this`: This section was retained.
+  // - `leader != this`: This section was merged with another idential section.
+  //                     Implies `killed_by_icf`.
```
rui314, line 338: Does this mean we can get rid of `killed_by_icf` flag and use this pointer instead?

rui314:
Great, thanks!

elf/mold.h
ishitatsuyuki: It’s assigned [here](https://github.com/rui314/mold/blob/b3ed06aa3956ed309ad2fe9ab757f6ee87f4f467/elf/icf.cc#L598), and it’s assigned from a single place, so we should be able to replace this with a helper function denoting the same condition.

rui314: Do you mind if I ask you to do this?

ishitatsuyuki: No problem, will do.

PR #807, opened 2022-10-22
[ELF] Fix uninitialized ElfSym paddings

ishitatsuyuki:
Designated initializer doesn't initialize the padding; make sure we memset them by ourselves.

Fixes readelf complaning about garbage after the visibility field.

Signed-off-by: Tatsuyuki Ishi <ishitatsuyuki@gmail.com>

### DIFF
--- a/elf/input-files.cc
+++ b/elf/input-files.cc
@@ -1535,6 +1535,7 @@ template <typename E>
 void SharedFile<E>::populate_symtab(Context<E> &ctx) {
   ElfSym<E> *symtab =
     (ElfSym<E> *)(ctx.buf + ctx.symtab->shdr.sh_offset) + this->global_symtab_idx;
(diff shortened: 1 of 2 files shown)


elf/input-files.cc
```diff
@@ -1535,6 +1535,7 @@ template <typename E>
 void SharedFile<E>::populate_symtab(Context<E> &ctx) {
   ElfSym<E> *symtab =
     (ElfSym<E> *)(ctx.buf + ctx.symtab->shdr.sh_offset) + this->global_symtab_idx;
+  memset(symtab, 0, sizeof(ElfSym<E>) * this->num_global_symtab);
```
rui314, line 1538: Not a big deal but I think it is slightly better if we move this memset inside the loop and clear each ElfSym individually just before setting values to the struct members. I believe compiler is smart enough to fuse the memset and struct member assignments. Also it should improve memory locality.

PR #810, opened 2022-10-22
[ELF] Run COMDAT elimination inside resolve_symbols.

ishitatsuyuki:
We've long had a hack to de-prioritize GNU_UNIQUE symbols, due to an edge case where these symbols could get higher resolution priority yet eliminated in the later COMDAT elimination stage. But since GNU_UNIQUE symbols were supposed to have a resolution as strong as GLOBAL symbols, it created unsoundness in the resolution and could have cause false duplicate symbol errors in later stages.

The root cause is that COMDAT elimination, unlike ICF and SHF_MERGE, removes symbols from the candidate pool and can actually affect resolution result. Hence we ought to do the elimination before the final resolution.

Doing so requires fully wiping the resolution instead of just the ones from eliminated archive members. This has a performance hit that we need to keep an eye on.

On another note this wiping of resolution is actually justifiable. The old algorithm did not handle the case of an archive member overriding a DSO symbol. So wiping is the correct way even in absence of COMDAT elimination.

One caveat is that currently the symbol resolution stage doesn't actually care about the is_alive flag which is how COMDAT symbols are killed. There are two other users of this flag: mergeable sections, and eh_frame. The eh_frame section is already handled with special logic, and we can treat it as if it had no symbols to resolve. Mergeable sections needs a bit more careful handling, but a good way is to just leave them unmerged and update the references in the same later pass where merge happens. This has been done in `Defer initialize_mergeable_sections`.

### DIFF
--- a/elf/input-files.cc
+++ b/elf/input-files.cc
@@ -818,22 +818,12 @@ static u64 get_rank(InputFile<E> *file, const ElfSym<E> &esym, bool is_lazy) {
     return (5 << 24) + file->priority;
   }
 
-  // GCC creates symbols in COMDATs with STB_GNU_UNIQUE instead of
-  // STB_WEAK if it was configured to do so at build time or the
-  // -fgnu-unique flag was given. In order to to not select a
-  // GNU_UNIQUE symbol in a discarded COMDAT section, we treat it as
-  // if it were weak.
-  //
-  // It looks like STB_GNU_UNIQUE is not a popular option anymore and
-  // often disabled by default though.
-  bool is_weak = (esym.st_bind == STB_WEAK || esym.st_bind == STB_GNU_UNIQUE);
-
   if (file->is_dso || is_lazy) {
-    if (is_weak)
+    if (esym.st_bind == STB_WEAK)
       return (4 << 24) + file->priority;
     return (3 << 24) + file->priority;
   }
-  if (is_weak)
+  if (esym.st_bind == STB_WEAK)
     return (2 << 24) + file->priority;
   return (1 << 24) + file->priority;
 }
@@ -897,7 +887,7 @@ void ObjectFile<E>::resolve_symbols(Context<E> &ctx) {
     InputSection<E> *isec = nullptr;
     if (!esym.is_abs() && !esym.is_common()) {
       isec = get_section(esym);
-      if (!isec)
+      if (!isec || !isec->is_alive)
         continue;
     }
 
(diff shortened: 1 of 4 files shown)


rui314:
I thought that there an edge case for handling comdats like this, but I can't think of any now. So let's give this a shot.

You want to revert my recent change with this: https://github.com/rui314/mold/commit/c4850273c79548b6bc46a094b0119c27434dfaa7#diff-8e029606ffb9f10045108f5aab4df54b48c555b7910e0025fdfe8097df12d31a

`.eh_frame` doesn't contain symbols except a few, and we have a special logic for the special symbols. See https://github.com/rui314/mold/blob/main/elf/mold.h#L2468. Essentially, we can treat `.eh_frame` as if they don't have any symbol at all.

ishitatsuyuki:
One caveat is that COMDAT elimination needs to be done after LTO, since the information only comes once object files are emitted. That said, the cost should be marginal compared to symbol resolution itself, so perf is not a big concern here. The justification is that COMDAT elimination can actually affect symbol resolution because they delete symbols, unlike ICF and mergeable sections where symbols are *updated* without changing there st_bind, etc. properties.

i686 and some other tests are wreaking havoc. I'll see what I did wrong.

ishitatsuyuki:
It looks like the "edge case" here is that COMDAT elimination also affects archive extraction. Symbols in an eliminated COMDAT group needs to be treated as an undefined symbol. Let's see if this strategy can work out.

ishitatsuyuki:
Here we go, this is an updated approach. Turns out we need to run COMDAT elimination after archive extraction but still before the final resolution. We used to re-use some results from the 1st resolution pass but now we completely throw it away for correctness. This has some performance impact, which we should keep an eye on.

ishitatsuyuki:
On the bright side, we should be resolving less symbols after we deduplicate COMDAT, so the improvement might balance things out.

ishitatsuyuki:
Tested on Chromium, it indeed has a bit of overhead. Although not too concerning since ICF and copying remains the larger chunk of time.

Before:
```
    1.221    0.099    0.042        resolve_symbols
(   0.592    0.197    0.032        register_section_pieces)
    0.034    0.000    0.004        eliminate_comdats
```

After:
```
    2.006    0.128    0.069        resolve_symbols
    0.077    0.000    0.004          eliminate_comdats
```

ishitatsuyuki:
Was this... merged? GitHub glitch?

rui314:
I pushed the "Merge" button and got an odd error, but it looks like it was merged.

PR #814, opened 2022-10-23
Add an optimized mark function for faster GC passes.

ishitatsuyuki:
seq_cst atomic operations can take hundreds of cycles and tends to be massively slower than a fence-free alternative.

Add a fast_mark function that optimistically tests if the node is already visited first, which allows avoiding the fence penalty if we can early exit.

When testing on Chromium, this roughly improved the associated pass's runtime by 20--30%.

Signed-off-by: Tatsuyuki Ishi <ishitatsuyuki@gmail.com>

### DIFF
--- a/mold.h
+++ b/mold.h
@@ -232,6 +232,16 @@ void update_maximum(std::atomic<T> &atomic, u64 new_val, Compare cmp = {}) {
          !atomic.compare_exchange_weak(old_val, new_val));
 }
 
+// An optimized "mark" operation for parallel mark-and-sweep algorithms.
+// Returns true if `visited` was false and updated to true.
+inline bool fast_mark(std::atomic<bool> &visited) {
+  // A relaxed load + branch (assuming miss) takes only around 20 cycles, while an atomic RMW can easily take hundreds
+  // on x86.
+  // We note that it's common that another thread beat us in marking, so doing an optimistic early test tends to improve
+  // performance in the ~20% ballpark.
+  return !visited.load(std::memory_order_relaxed) && !visited.exchange(true, std::memory_order_relaxed);
+}
+
 template <typename T, typename U>
 inline void append(std::vector<T> &vec1, std::vector<U> vec2) {
   vec1.insert(vec1.end(), vec2.begin(), vec2.end());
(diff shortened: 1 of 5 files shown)


mold.h
```diff
@@ -232,6 +232,16 @@ void update_maximum(std::atomic<T> &atomic, u64 new_val, Compare cmp = {}) {
...
 }
 
+// An optimized "mark" operation for parallel mark-and-sweep algorithms.
+// Returns true if `visited` was false and updated to true.
+inline bool fast_mark(std::atomic<bool> &visited) {
+  // A relaxed load + branch (assuming miss) takes only around 20 cycles, while a seq_cst operation can easily take
+  // hundreds.
+  // We note that it's common that another thread beat us in marking, so doing an optimistic early test tends to improve
+  // performance in the ~20% ballpark.
+  return !visited.load(std::memory_order_relaxed) && !visited.exchange(true, std::memory_order_seq_cst);
```
rui314, line 242: In the marking phase, we only mutate `visited` flag, so is sequential consistency really needed? As long as each individual `exchange` is atomic, actual order of marking doesn't matter. So I wonder if we can just do `return !visited.exchange(true, std::memory_order_relaxed);` here and add memory barriers before and after `visit` like this

```
  std::atomic_thread_fence(std::memory_order_acquire);
  tbb::parallel_for_each(rootset, [&](InputSection<E> *isec,
                                    tbb::feeder<InputSection<E> *> &feeder) {
    visit(ctx, isec, feeder, 0);
  });
  std::atomic_thread_fence(std::memory_order_release);
```

to guarantee that the relaxed write to `visited` is guaranteed to be visible from all threads after the `parallel_for_each`.

ishitatsuyuki: Yes, `relaxed` is sufficient. You don't need a fence either, that's handled by whatever logic `parallel_for_each` uses to wait for completion (if we see the completion, we're guaranteed to also see whatever happened before the completion).

I think relaxed should be good for all mark-and-sweep like algos, so I'll change it to use that.

That said, all RMW operations have the same cost on x86 (lock prefix). `relaxed` can be a bit faster for ARM (LL/SC), but it's still more costly than an relaxed load/store. So the early exit testing is still needed.

ishitatsuyuki: I updated it to use relaxed and the comments to mention RMW instead of SeqCst.

PR #818, opened 2022-10-24
Some trivial optimizations

ishitatsuyuki:
Explanation in commit message.

### DIFF
--- a/elf/input-files.cc
+++ b/elf/input-files.cc
@@ -926,7 +926,6 @@ ObjectFile<E>::mark_live_objects(Context<E> &ctx,
     if (esym.is_weak())
       continue;
 
-    std::scoped_lock lock(sym.mu);
     if (!sym.file)
       continue;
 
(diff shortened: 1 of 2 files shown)


rui314:
We set to `sym.flags` in arch-*.cc files. I believe that also enforces sequential consistency. Does changing the code improve `scan_relocations`?

ishitatsuyuki:
I don’t have an ARM testing environment, so no idea how much improvement we can expect here (the one in this PR is mostly done because it’s trivial and reasonably hot, e.g. update_minimum gets called in loop in HLL reductions).

For optimizing flags (assuming it’s actually hot), I’d suggest something different: put each flag into a 1-byte field so they can be individually set with an ordinary store. This is something that can help on x86, while relaxed RMWs don’t (again, all RMWs are equal on x86).

ishitatsuyuki:
Scan reloc only takes 24ms for Chromium so it doesn’t seem super hot. I wonder if there’s some example that could be more reloc (or more accurately, PLT/GOT) heavy?

rui314:
Yeah, maybe it doesn't worth to change it. It doesn't stand out in the perf logs.

Issue #813, opened 2022-10-23
Symbol resolution optimization ideas


ishitatsuyuki:
While working on #810 I've looked for some (low-level) opportunities to undo the performance regression so I'll put the ideas here.

**`clear_symbols` take a redundant lock.**
This lock is *mostly* unnecessary:
https://github.com/rui314/mold/blob/61f28d270c3c6b9b88392c4c07d54c3362d9762a/elf/input-files.cc#L55
Perf changes from 69ms to 51ms. WTF.

The caveat is that plain assignments don't enjoy as much guarantees as `atomic` with `mo_relaxed`, and notably the spec allows observing *tearing* writes. On practical platforms the atomicity is guaranteed in hardware at no cost. Boehm did a [proposal](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4013.html) before for a standard compliant way to do this but it seems the progress is stalled.

Using plain atomics have poor ergonomics, because the operators on them defaults to seq_cst and we need to manually call the `<atomic>` member functions with `mo_relaxed` to get the zero-cost relaxed ordering. Compilers tend to treat relaxed atomic operations as a blackbox which also leads to them being optimized poorly.

On the other hand if we decide to just not care about compliance we can just disable TSan for the block. It's more pragmatic but puts more assumption against the architecture. That said, typically it's safe up to pointer size (which means the only unsafe thing is 64bit stores on 32bit archs).

**Experimenting with multi-pass resolution**
"Multi-pass resolution" isn't a descriptive name, but what I want to do here is to do two pass: 1. Determine the min cost and 2. Once the min cost is determined, actually assign the content. Pass 2 can be done without locking and pass 1 can be done with atomic instructions, which is an improvement from mutexes. A similar idea is used inside COMDAT elimination.

Right now we take a mutex which adds to wait time.
https://github.com/rui314/mold/blob/61f28d270c3c6b9b88392c4c07d54c3362d9762a/elf/input-files.cc#L905

---

There are a few more usages of `sym.mu`. Would be cool if we can get rid of all of them.

ishitatsuyuki:
I tried iterating across the ctx symbol map, but it was over 10 times slower despite having better theoretical complexity. TBB’s concurrent hashmap is probably to blame.

If we don’t have a better option here, I think the easier way here would be to add an `atomic<T>` wrapper that turns reads and assignments into relaxed.

rui314:
[...]

PR #819, opened 2022-10-24
Remove lock from clear_symbols and add a justification for disabling TSan

ishitatsuyuki:
As discussed in #813, doing a mutex lock is too pessimistic here but having a correct fix would also be cumbersome. Hence we add a justification for it and prevent TSan from complaining about the harmless race.

### DIFF
--- a/elf/input-files.cc
+++ b/elf/input-files.cc
@@ -49,10 +49,15 @@ ElfShdr<E> *InputFile<E>::find_section(i64 type) {
   return nullptr;
 }
 
+// This function is actually racy, in that "tearing" might be observed during the read or write of `sym->file`.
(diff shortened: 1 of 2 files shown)


mold.h
```diff
@@ -40,6 +40,16 @@
 # define unreachable() assert(0 && "unreachable")
 #endif
 
+#if defined(__has_feature)
+# if __has_feature(thread_sanitizer)
+#  define MOLD_TSAN_DISABLE __attribute__((no_sanitize("thread")))
```
rui314, line 45: Does unrecognized `__attribute__` cause an error for gcc/clang? If not, we want to directly use this without macro.

PR #832, opened 2022-10-26
[Mach-O] Use binary search in find_subsection()

JamesWidman:
When using ld64.mold to link clang for Mach-O, most of the time was spent in the first call to find_subsection() from
parse_compact_unwind(). Changing its linear search to a binary search cut an 18-second link time down to about 3 seconds.

Signed-off-by: James Widman <james.widman@gmail.com>

### DIFF
--- a/macho/input-files.cc
+++ b/macho/input-files.cc
@@ -418,11 +418,18 @@ LoadCommand *ObjectFile<E>::find_load_command(Context<E> &ctx, u32 type) {
 template <typename E>
 Subsection<E> *
 ObjectFile<E>::find_subsection(Context<E> &ctx, u32 secidx, u32 addr) {
-  Subsection<E> *ret = nullptr;
-  for (Subsection<E> *subsec : subsections)
-    if (subsec->isec.secidx == secidx && subsec->input_addr <= addr)
-      ret = subsec;
-  return ret;
+  assert(subsections.size() > 0);
+  auto begin = subsections.begin();
+  auto end = subsections.end();
+  auto upper = std::upper_bound(begin, end, addr,
+      [](u32 addr, Subsection<E> *subsec) { return addr < subsec->input_addr; });
+
+  if (upper == begin)
+    return nullptr;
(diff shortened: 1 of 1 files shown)


macho/input-files.cc
```diff
@@ -418,11 +418,18 @@ LoadCommand *ObjectFile<E>::find_load_command(Context<E> &ctx, u32 type) {
...
-      ret = subsec;
-  return ret;
+  assert(subsections.size() > 0);
+  auto begin = subsections.begin();
+  auto end = subsections.end();
+  auto upper = std::upper_bound(begin, end, addr,
+      [](u32 addr, Subsection<E> *subsec) { return addr < subsec->input_addr; });
+
+  if (upper == begin)
+    return nullptr;
```
ishitatsuyuki, line 428:
```suggestion
  auto upper = std::ranges::upper_bound(subsections, addr, {}, &Subsection<E>::input_addr);

  if (upper == subsections.begin())
    return nullptr;
```

C++20 Ranges & Projections.

JamesWidman: I want to try this, but clang (with whatever headers are shipped with MacOSX12.3.sdk) replies: 

```
error: no member named 'upper_bound' in namespace 'std::ranges'; did you mean 'std::upper_bound'?
  auto upper = std::ranges::upper_bound(subsections, addr, {}, &Subsection<E>::input_addr);
               ^~~~~~~~~~~~~~~~~~~~~~~~
               std::upper_bound
```

ishitatsuyuki: Sigh, looks like it’s a moment where Apple clang is lagging far behind (per https://en.cppreference.com/w/cpp/compiler_support). Nevermind then.

rui314:
We actually used to do binary search in this function but it turned out that it caused a crash when the linker relaxation is turned on. The option was buggy so I turned it off by default (i.e. `-ignore_optimization_hints` is now default). We probably should remove the relaxation code completely until the linker becomes stable. I'm fine to merge this commit.

Issue #842, opened 2022-10-31
ICF optimization ideas


ishitatsuyuki:
Identical Code Folding (ICF) is a code size optimization. It is currently used by mainly Chromium, but there are also interests within the Rust community to [mitigate some of the code duplication](https://kobzol.github.io/rust/rustc/2022/10/27/speeding-rustc-without-changing-its-code.html#using-icf-for-rustc-99062) that comes from the current codegen design. Due to the inherent difficulty of the problem, ICF is the most compute intensive pass inside mold.

Below I outline some opportunities to make ICF faster.

### Faster skipping of converged elements

ICF turns code in to a DAG and hashes the tree of nodes up to depth N from each node, with N increasing each iteration. When the descendants of the node doesn't contain cycles, this hash will converge once it reaches the leaf, and further iterations can be skipped as it doesn't change result.

Right now, this skipping is implemented through a [comparison of digests](https://github.com/rui314/mold/blob/2942ba83d4b1f189f0be1103e8498a3b6154e6e5/elf/icf.cc#L416). If instead, the "convergence" is stored into a bitset, this skipping can be done through 1 bit of memory access per node, instead of `2*HASH_BITS`. This cuts quite a few instruction in the later phase and also improves cache utilization. This cuts around 50ms out of 310ms for the ICF pass, or out of 1.012s for total linking of Chromium.

### Faster cardinality estimates through hash-based approximation

The ICF is converged iff the cardinality of node hashes doesn't increase between depth N and N+1. To test for this cardinality, right now we perform a parallel sort which is rather costly (O(n log n) work and O(n) latency). The cost is somewhat amortized by only running it after 10 iterations of propagation, but O(n) latency is still not a nice number.

It might be possible to test for non-convergence through probabilistic method. Below is a non-exhaustive list of known options:
- HyperLogLog (HLL). HLL is currently implemented in mold to estimate cardinality before inserting into some concurrent hashmaps (e.g. string table). HLL's primary problem is that it has a rather high absolute error, making it too noisy for testing convergence. HLL operations are associative and as such is trivially parallelizable as a reduce.
- Bloom Filter. Bloom Filter is a famous algorithm to approximately answer the membership test, but it can also be used to estimate cardinality. A [survey](https://www.vldb.org/pvldb/vol11/p499-harmouch.pdf) shows that "the relative error of Bloom filter is inversely proportional to the exact cardinality of the dataset", making it suitable for our use case where even an off-by-one error could affect the effectiveness of estimation.
  
[...]

PR #845, opened 2022-11-01
[ELF] ICF optimizations and simplifications

ishitatsuyuki:
Implements the low-hanging fruit from #842, and a few additional simplifications while on it.

### DIFF
--- a/elf/icf.cc
+++ b/elf/icf.cc
@@ -402,34 +402,59 @@ static void gather_edges(Context<E> &ctx,
   });
 }
 
+static std::vector<u64> init_converged_mask(i64 num_digests) {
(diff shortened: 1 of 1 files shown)


rui314:
I wonder if you can use `std::vector<bool>` instead. IIUC, `std::vector` is specialized for `bool` so that it uses only one bit for each element.

ishitatsuyuki:
Hm actually it wasn't removed ([the proposal](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2007/n2204) was rejected), but it's widely considered a mistake anyway and I don't think it's worth using given it lacks ability to do a fast "count last zero" operation anyway.

PR #846, opened 2022-11-01
Respect prefix set by `cmake --install`

ZhongRuoyu:
It is noticed that the tarballs in 1.6.0 do not come with the symlinks (like `bin/ld.mold`). This is because the values of `CMAKE_INSTALL_FULL_{LIBEXEC,BIN,MAN}DIR` are already hardcoded in the rules defined in `cmake_install.cmake` during configuration. They are not updated when `cmake --install <build-dir> --prefix <prefix>` overrides the install prefix. So, despite the prefix override in `cmake --install . --prefix $dest --strip`, the symlinks went to `/usr/local` instead.

https://github.com/rui314/mold/blob/323ad30e25c2c81efdb07ab76601a119335a40c8/CMakeLists.txt#L365-L377

A trivial first attempt would be to make them evaluated during installation, like this:

```diff
      execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink \${RELPATH}
-       \$ENV{DESTDIR}/${CMAKE_INSTALL_FULL_LIBEXECDIR}/mold/ld)
+       \$ENV{DESTDIR}/\${CMAKE_INSTALL_FULL_LIBEXECDIR}/mold/ld)
```

But unfortunately that does not work -- `${CMAKE_INSTALL_FULL_LIBEXECDIR}` is not available during installation. From CMake's manpage (["INSTALL A PROJECT", cmake(1)](https://cmake.org/cmake/help/latest/manual/cmake.1.html#install-a-project)), it seems only `CMAKE_INSTALL_PREFIX` will reflect the override. So this is the only thing we can rely on.

My second attempt was the following:

```diff
      execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink \${RELPATH}
-       \$ENV{DESTDIR}/${CMAKE_INSTALL_FULL_LIBEXECDIR}/mold/ld)
+       \$ENV{DESTDIR}/\${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBEXECDIR}/mold/ld)
```

That, however, does not work when the `DESTDIR` environment variable is not set and the prefix is a relative path, because a bare slash `/` would be prepended.

Therefore, I am proposing this fix. The symlink creation logic is extracted into a function `mold_install_symlink(OLD NEW)`. It first sets a variable `PREFIX` depending on whether `DESTDIR` is defined or not. Specifically, if it is not, then `PREFIX` and all that follow are relative paths, which is fine. Then it creates a symlink that does the same thing as `ln -s ${PREFIX}${NEW} ${OLD}`, creating parent directories if they do not exist.

By the way, it seems okay to use the [`MAKE_DIRECTORY`](https://cmake.org/cmake/help/latest/command/file.html#make-directory) and [`CREATE_LINK`](https://cmake.org/cmake/help/latest/command/file.html#create-link) subcommands under command `file`. So, it shouldn't be needed to invoke a subprocess.

cc @jpalus, author of #738.

Signed-off-by: Zhong Ruoyu <zhongruoyu@outlook.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -376,17 +376,19 @@ if(NOT CMAKE_SKIP_INSTALL_RULES)
   install(TARGETS mold RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
   install(FILES LICENSE DESTINATION ${CMAKE_INSTALL_DOCDIR})
   install(FILES docs/mold.1 DESTINATION ${CMAKE_INSTALL_MANDIR}/man1/)
-  install(CODE "
-    file(RELATIVE_PATH RELPATH
-       /${CMAKE_INSTALL_FULL_LIBEXECDIR}/mold /${CMAKE_INSTALL_FULL_BINDIR}/mold)
-    execute_process(COMMAND ${CMAKE_COMMAND} -E make_directory
-      \$ENV{DESTDIR}/${CMAKE_INSTALL_FULL_LIBEXECDIR}/mold)
-    execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink \${RELPATH}
-      \$ENV{DESTDIR}/${CMAKE_INSTALL_FULL_LIBEXECDIR}/mold/ld)
-    execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink mold
-      \$ENV{DESTDIR}/${CMAKE_INSTALL_FULL_BINDIR}/ld.mold)
-    execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink mold
-      \$ENV{DESTDIR}/${CMAKE_INSTALL_FULL_BINDIR}/ld64.mold)
-    execute_process(COMMAND ${CMAKE_COMMAND} -E create_symlink mold.1
-      \$ENV{DESTDIR}/${CMAKE_INSTALL_FULL_MANDIR}/man1/ld.mold.1)")
+
+  function(mold_install_symlink OLD NEW)
+    install(CODE "
+      get_filename_component(PREFIX_ABS \${CMAKE_INSTALL_PREFIX} ABSOLUTE)
+      get_filename_component(NEW_DIR
+        \$ENV{DESTDIR}/\${PREFIX_ABS}/${NEW} DIRECTORY)
+      file(MAKE_DIRECTORY \${NEW_DIR})
+      file(CREATE_LINK ${OLD} \$ENV{DESTDIR}/\${PREFIX_ABS}/${NEW} SYMBOLIC)")
+  endfunction()
+  file(RELATIVE_PATH RELPATH
+       /${CMAKE_INSTALL_LIBEXECDIR}/mold /${CMAKE_INSTALL_BINDIR}/mold)
+  mold_install_symlink(${RELPATH} ${CMAKE_INSTALL_LIBEXECDIR}/mold/ld)
+  mold_install_symlink(mold ${CMAKE_INSTALL_BINDIR}/ld.mold)
+  mold_install_symlink(mold ${CMAKE_INSTALL_BINDIR}/ld64.mold)
+  mold_install_symlink(mold.1 ${CMAKE_INSTALL_MANDIR}/man1/ld.mold.1)
 endif()


ZhongRuoyu:
To see what would happen:

```diff
    function(mold_install_symlink OLD NEW)
      install(CODE "
        if(DEFINED ENV{DESTDIR})
          set(PREFIX \$ENV{DESTDIR}/\${CMAKE_INSTALL_PREFIX}/)
        else()
          set(PREFIX \${CMAKE_INSTALL_PREFIX}/)
        endif()
        get_filename_component(NEW_DIR \${PREFIX}${NEW} DIRECTORY)
+       message(STATUS \"Installing symlink: \${PREFIX}${NEW} -> ${OLD}\")
        file(MAKE_DIRECTORY \${NEW_DIR})
        file(CREATE_LINK ${OLD} \${PREFIX}${NEW} SYMBOLIC)")
    endfunction()
```

With no prefix override:

```bash
root@8fca0e0f1cc7:~/mold/build# cmake --install .
-- Install configuration: "Release"
-- Installing: /usr/local/lib/mold/mold-wrapper.so
-- Installing: /usr/local/bin/mold
-- Installing: /usr/local/share/doc/mold/LICENSE
-- Installing: /usr/local/share/man/man1/mold.1
-- Installing symlink: /usr/local/libexec/mold/ld -> ../../bin/mold
-- Installing symlink: /usr/local/bin/ld.mold -> mold
-- Installing symlink: /usr/local/bin/ld64.mold -> mold
-- Installing symlink: /usr/local/share/man/man1/ld.mold.1 -> mold.1
```

With prefix override, relative path:

```
root@8fca0e0f1cc7:~/mold/build# cmake --install . --prefix out
-- Install configuration: "Release"
-- Up-to-date: /root/mold/build/out/lib/mold/mold-wrapper.so
-- Up-to-date: /root/mold/build/out/bin/mold
-- Up-to-date: /root/mold/build/out/share/doc/mold/LICENSE
-- Up-to-date: /root/mold/build/out/share/man/man1/mold.1
-- Installing symlink: out/libexec/mold/ld -> ../../bin/mold
-- Installing symlink: out/bin/ld.mold -> mold
-- Installing symlink: out/bin/ld64.mold -> mold
-- Installing symlink: out/share/man/man1/ld.mold.1 -> mold.1
```

With prefix override, absolute path:

```
root@8fca0e0f1cc7:~/mold/build# cmake --install . --prefix /mold
-- Install configuration: "Release"
-- Installing: /mold/lib/mold/mold-wrapper.so
-- Installing: /mold/bin/mold
-- Installing: /mold/share/doc/mold/LICENSE
-- Installing: /mold/share/man/man1/mold.1
-- Installing symlink: /mold/libexec/mold/ld -> ../../bin/mold
-- Installing symlink: /mold/bin/ld.mold -> mold
-- Installing symlink: /mold/bin/ld64.mold -> mold
-- Installing symlink: /mold/share/man/man1/ld.mold.1 -> mold.1
```

With both prefix overridden and `DESTDIR` set:

```
root@8fca0e0f1cc7:~/mold/build# DESTDIR=/tmp cmake --install . --prefix out
-- Install configuration: "Release"
-- Installing: /tmp/root/mold/build/out/lib/mold/mold-wrapper.so
-- Installing: /tmp/root/mold/build/out/bin/mold
-- Installing: /tmp/root/mold/build/out/share/doc/mold/LICENSE
-- Installing: /tmp/root/mold/build/out/share/man/man1/mold.1
-- Installing symlink: /tmp/out/libexec/mold/ld -> ../../bin/mold
-- Installing symlink: /tmp/out/bin/ld.mold -> mold
-- Installing symlink: /tmp/out/bin/ld64.mold -> mold
-- Installing symlink: /tmp/out/share/man/man1/ld.mold.1 -> mold.1
```

This lets the user know what happens as well. Should we add this status message?

rui314:
Cool! Thank you for doing this!

jpalus:
Unfortunately this change breaks again case that #738 was trying to solve: absolute paths in `CMAKE_INSTALL_*` vars. I'll try to look at it later today.

ZhongRuoyu:
> Unfortunately this change breaks again case that #738 was trying to solve: absolute paths in `CMAKE_INSTALL_*` vars. I'll try to look at it later today.

Sorry! Can you have a look at #848?

hwti:
@ZhongRuoyu This change also breaks the support of `-DCMAKE_INSTALL_PREFIX=` and `-DCMAKE_INSTALL_PREFIX=/`.

They have a different behavior in `GNUInstallDirs`, but both result in a **empty** `CMAKE_INSTALL_PREFIX` when `mold_install_symlink` (and later `mold_install_relative_symlink`) are called.
So `get_filename_component(PREFIX_ABS \${CMAKE_INSTALL_PREFIX} ABSOLUTE)` fails.

ZhongRuoyu:
> @ZhongRuoyu This change also breaks the support of `-DCMAKE_INSTALL_PREFIX=` and `-DCMAKE_INSTALL_PREFIX=/`.
> 
> They have a different behavior in `GNUInstallDirs`, but both result in a **empty** `CMAKE_INSTALL_PREFIX` when `mold_install_symlink` (and later `mold_install_relative_symlink`) are called. So `get_filename_component(PREFIX_ABS \${CMAKE_INSTALL_PREFIX} ABSOLUTE)` fails.

@hwti Could you please have a look at #868 and see if it works for you?

PR #856, opened 2022-11-11
[Mach-O] Ignore some unsupported command line flags

keith:
By only warning about these flags users can more easily test mold without having to figure out how to remove these from their builds. Some may lead to issues, but for testing that should be ok.


rui314:
Honestly I don't think we should to do this because these flags cannot be ignored safely. I'd keep it as a private patch.

keith:
Fwiw this is how lld rolled out flags as well. The other option is people can't test at all

rui314:
But if we ignore these flags, when a linker output doesn't work, we can't tell whether it is because we ignore a necessary flag or not.

PR #877, opened 2022-11-19
Fix out-of-bounds error on aarch64 with `_GLIBCXX_ASSERTIONS` enabled…

sicherha:
… (again)

The square-bracket operator of `std::span` has undefined behaviour for out-of-bounds element accesses. Resorting to iterators yields defined behaviour.

This issue (originally described in #298) had been fixed in the past by commit 03ba8f93255055adaf69984c17bcfea3b22340db, but it has since resurfaced with commit 193c245781df18e3c7d96efc1ea03e9760175681.


ishitatsuyuki (approved):
Nice find!

Issue #849, opened 2022-11-06
CMake integration


rui314:
Currently, each project has to add some code to their `CMakeLists.txt` to use mold. To lower the bar to start using mold, we probably should provide our own cmake module. It should be usable just by including and calling a function like this:

```
include(UseLinker)
target_use_linker(<target-name> mold)
```

We might even want to ask CMake to include such module to their distribution.

jschueller:
the cmake way to do distribute cmake macros is to provide MoldConfig.cmake file containing with such a macro then downstream users might do:
```
find_package(Mold CONFIG)
mold_target_use_linker(...)
```
(this file should be installed in prefix/lib/cmake/mold for example, along with other mold files)

gruenich:
CMake does not offer a built-in mechanism to change the linker. You have to set a variable for every language in use, and its value depends on the actual compiler, e.g., passing `-DCMAKE_CXX_FLAGS="-fuse-ld=mold"` to the CMake call for Clang. This problem occurs with gold or lld, too.

Someone should [file a feature request for CMake](https://gitlab.kitware.com/cmake/cmake/-/issues). If nobody is willing, I can do this.

rui314:
@gruenich Please go ahead and file a feature request. Thanks!

gruenich:
Done, I opened [issue 24254](https://gitlab.kitware.com/cmake/cmake/-/issues/24254). Feel free to commend and upvote!

ADKaster:
It looks like CMake 3.29 will support a `CMAKE_LINKER_TYPE=mold` option, as the MR was just merged today. [gitlab.kitware.com/cmake/cmake/-/merge_requests/8861](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/8861)

That MR closes the issue opened in https://github.com/rui314/mold/issues/849#issuecomment-1356834692

rui314:
Great news! I think we can now mark this as "resolved".

PR #878, opened 2022-11-19
Create mold-config.cmake

ZhongRuoyu:
This creates a CMake config file that makes it easy for users to use mold in their CMake projects. Sample usage:

```
root@1b8ec34409d4:~/hello# cat CMakeLists.txt
add_executable(hello hello.cc)
find_package(mold CONFIG)
if(mold_FOUND)
  target_use_mold(hello)
endif()

root@1b8ec34409d4:~/hello# mkdir build && cd build

root@1b8ec34409d4:~/hello/build# cmake ..
<output truncated>
-- Build files have been written to: /root/hello/build

root@1b8ec34409d4:~/hello/build# cmake --build .
[ 50%] Building CXX object CMakeFiles/hello.dir/hello.o
[100%] Linking CXX executable hello
[100%] Built target hello

root@1b8ec34409d4:~/hello/build# readelf -p .comment hello
String dump of section '.comment':
  [     0]  GCC: (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0
  [    2c]  mold 1.7.0 (b63bb4aecf7afd23858d8d743eee734ca9f3868d; compatible with GNU ld)
```

Thanks to [`configure_package_config_file`](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#command:configure_package_config_file) it is not difficult to create a relocatable config file that can be installed anywhere, because no absolute paths are hardcoded in the generated file. Instead there's this line in the generated file which computes the prefix relative to its own location:

```cmake
get_filename_component(PACKAGE_PREFIX_DIR "${CMAKE_CURRENT_LIST_DIR}/../../../" ABSOLUTE)
```

`target_use_mold` is implemented by simply adding a linker flag. It assumes that `CMAKE_LINKER` is not manually set to a linker executable, or something else that's not compatible (if that's the case, the user won't need this anyway). `-B${LIBEXECDIR}/mold/ld` is used on non-macOS targets. For macOS targets, `-fuse-ld=${BINDIR}/ld64.mold` is used, because the `ld` in mold's `LIBEXECDIR` will invoke the ELF linker.

I created this with the intention of resolving #849, but it ended up being a little different from what #849 asked for, as this `target_use_mold` function does not allow the user to choose their own linker. But here's what I think:

1. Just as mentioned in the readme, it seems that the compatible way to tell both Clang and GCC (pre-12.1) to use a specific linker is to use the `-B` flag. But that requires the linker to have its own `ld` somewhere. To my knowledge, there's no such thing from gold and lld. [^1]
2. I personally think that maybe `target_use_mold` is good enough. The config file is distributed with mold, so it would be a little peculiar to do something like:
```cmake
find_package(mold CONFIG)
if(mold_FOUND)
  mold_target_use_linker(<target> lld)
endif()
```

What do you think, @rui314?

[^1]: That's not impossible to implement, though -- a proxy as simple as `ld.$LINKER "$@"` will do, I think.


ZhongRuoyu:
I made a stupid mistake that I didn't realise until I rebased -- there's no reason to restrict that to executable targets.

rui314:
I think we need to do two things for CMake integration:

**Short term solution**: We want to provide a package like this, and let the user to use the mold linker if it is available. So the typical usage pattern would be:

```
find_package(mold CONFIG)
if(mold_FOUND)
  target_use_mold(<target-name>)
endif()
```

**Long term solution***: I'd generalize the fetature so that the function takes a linker name like `target_use_linker(<target_name> [ bfd | gold | lld | mold])`. And then contribute it to CMake rather than to this project.

This patch should work as a short-term solution, so I'm fine with that. I'm no CMake expert, so I'm not sure if it is good enough though, so let me allow more time to review it. But it's general looking good.

Do you want to work on the long-term solution by the way?

ZhongRuoyu:
> **Short term solution**: We want to provide a package like this, and let the user to use the mold linker if it is available.

Agreed. So that's what's being done here, on mold's side.

> ```
> find_package(mold CONFIG)
> if(mold_FOUND)
>   target_use_mold(<target-name>)
> endif()
> ```

Oops, I missed the `mold_FOUND` part -- will update the usage example later.

> This patch should work as a short-term solution, so I'm fine with that. I'm no CMake expert, so I'm not sure if it is good enough though, so let me allow more time to review it. But it's general looking good.

Sure, thanks, Rui-san.

> Do you want to work on the long-term solution by the way?

Understood. I'd be genuinely interested! But speaking off the top of my head, that'd be quite some work, and from where I'm standing, I may not be familiar enough with CMake to qualify :( . While I'll definitely have a look at it, I think it's best to leave that opportunity open to those who are more capable.

gruenich:
This pull request is a bad solution for the short-term problem. `*-config.cmake` are used to pass include directories, libraries, compile options, and CMake target. None is required for using mold. All you need to do is find the executalbe mold or mold.exe.

Given CMake 3.29 (not yet released) will [support an easy way to set the linker](https://github.com/rui314/mold/issues/849), I suggest closing this pull request.

Issue #884, opened 2022-11-20
Segmentation fault in libuuid


SebastianSpeitel:
Executing the following script in python produces a segmentation fault in libuuid when it was built with mold:

```bash
python -c 'import _uuid; _uuid.generate_time_safe()'
```
```
python[14667] general protection fault ip:7f3d8eee792a sp:7fff6d9a86d0 error:0 in libuuid.so.1.3.0[7f3d8eee7000+5000]
```

`_uuid` is the internal uuid c-wrapper from python used in the builtin `uuid` module. It's dynamically linked to libuuid, which in my case is produced by [util-linux](https://github.com/util-linux/util-linux).

When building without mold no segmentation fault occurs.

My environment:

 * mold 1.7.1 (compatible with GNU ld)
 * gcc version 12.2.0 (Gentoo 12.2.0 p1)
 * Python 3.10.8
 * util-linux: 2.37.4

ishitatsuyuki:
Can you try running with `gdb` and get a backtrace? That might give some more idea about what's going wrong.

SebastianSpeitel:
That's what i get:

```gdb
(gdb) bt
#0  0x00007ffff7fb092a in __uuid_generate_time () from /lib64/libuuid.so.1
#1  0x00007ffff7fb89a0 in ?? () from /usr/lib/python3.10/lib-dynload/_uuid.cpython-310-x86_64-linux-gnu.so
#2  0x00007ffff7dfd526 in ?? () from /usr/lib64/libpython3.10.so.1.0
#3  0x00007ffff7e8892d in _PyEval_EvalFrameDefault () from /usr/lib64/libpython3.10.so.1.0
#4  0x00007ffff7e864f0 in ?? () from /usr/lib64/libpython3.10.so.1.0
#5  0x00007ffff7e8647b in PyEval_EvalCode () from /usr/lib64/libpython3.10.so.1.0
#6  0x00007ffff7ecb853 in ?? () from /usr/lib64/libpython3.10.so.1.0
#7  0x00007ffff7ecb7ea in ?? () from /usr/lib64/libpython3.10.so.1.0
#8  0x00007ffff7ecc7c1 in ?? () from /usr/lib64/libpython3.10.so.1.0
#9  0x00007ffff7ecc584 in _PyRun_SimpleFileObject () from /usr/lib64/libpython3.10.so.1.0
[...]

PR #893, opened 2022-11-25
[ELF] TLS alignment fix

ishitatsuyuki:
This fixes a TLS alignment issue discovered in #884.

See commit message for more details.

Closes: #884

### DIFF
--- a/elf/passes.cc
+++ b/elf/passes.cc
@@ -1772,6 +1772,29 @@ static void set_virtual_addresses_regular(Context<E> &ctx) {
   std::vector<Chunk<E> *> &chunks = ctx.chunks;
   u64 addr = ctx.arg.image_base;
 
+  // TLS chunks alignments are special: in addition to having their virtual
+  // addresses aligned, they also have to be aligned when the region of
+  // tls_begin is copied to a new thread's storage area. In other words, their
+  // offset against tls_begin also has to be aligned.
+  //
+  // A good way to achieve this is to take the largest alignment requirement
+  // of all TLS sections and make tls_begin also aligned to that.
+  Chunk<E> *first_tls_chunk = nullptr;
+  u64 first_tls_chunk_alignment = 1;
+  for (Chunk<E> *chunk : chunks) {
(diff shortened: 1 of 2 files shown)


ishitatsuyuki:
@rui314 Could you take a look at [`[ELF] Re-introduce extra_addralign.`](https://github.com/rui314/mold/pull/893/commits/125f2c1bc9c185c879b431c0477748660918ab87) and let me know what you think? I would like to add another refactor so that the RELRO also uses `extra_addralign` but it would be good to have a confirmation that this is a good direction before going for that.

One thing to consider is that `extra_addralign` will affect the behavior even when `--section-order` is used, however it seems reasonable to also apply RELRO alignment adjustments for those cases (as RELRO wouldn't be possible otherwise).

rui314:
Do you think you can write a test to trigger the existing issue?

Currently, `set_virtual_addresses` completely controls how sections are laid out in memory, and `create_phdr` makes its best effort to create fewer numbers of program headers that covers a given section layout. There's no information flow from `create_phdr` to `set_virtual_addresses`. I think I like this design. Previously, `create_phdr` and `set_virtual_addresses` depend on each other, and it was not easy to understand how the two functions worked.

I think you can move the logic to `set_virtual_addresses`. In that function, you can all chunks to find the maximum tdata/tbss alignment and use that information for alignment.

ishitatsuyuki:
> Do you think you can write a test to trigger the existing issue?

Should be possible, the one I have in mind is to declare a 4-byte initialized variable (that goes into `.tdata`) then a 16-byte *aligned* zero variable (that goes into `.tbss`), then assert the latter have the right alignment during execution.

> I think you can move the logic to set_virtual_addresses. In that function, you can all chunks to find the maximum tdata/tbss alignment and use that information for alignment.

Ack.

ishitatsuyuki:
Updated with a new test and the logic consolidated in `set_virtual_addresses_regular`. Testing turned out to be quite tricky, as the crash actually doesn't produce if one of the optimized TLS model is used (initial-exec etc.). The test uses dlopen to force the global-dynamic behavior.

PR #902, opened 2022-12-03
Prefer --ld-path= to specify mold full path

TerrorJack:
See https://github.com/llvm/llvm-project/blob/4178671b2ed3d8c1ffec6f723808e22572601242/clang/include/clang/Basic/DiagnosticDriverKinds.td#L567, using `-fuse-ld=` to specify linker full path is deprecated in clang, `--ld-path` should be preferred in that case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ following flags to use `mold` instead of `/usr/bin/ld`:
   `-B/usr/libexec/mold` (or `-B/usr/local/libexec/mold`) to GCC.
 
 If you haven't installed `mold` to any `$PATH`, you can still pass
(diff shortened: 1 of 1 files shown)


rui314:
I believe `--ld-path` is relatively new. Since which version of clang does it accept `--ld-path`?

Alcaro:
Some quick checking at Godbolt says that it exists in Clang 12 but not 11.

rui314:
Clang/LLVM 11.0.0 was released in December 2020, so it's too early to make this change. We should wait until like 2026 to make this change.

PR #907, opened 2022-12-08
[ELF] Keep a reference to the defsym target. 

### LINKED ISSUE #829: collect2: fatal error: ld terminated with signal 11 [Segmentation fault], core dumped
The following crashes:

```a.c
extern void g(void);
void real_g(void) {}

int main()
{
  g();
  return 0;
}
```

```
$ gcc-12 -flto a.c -Wl,--defsym,g=real_g -B ~/Programming/mold/objdir
collect2: fatal error: ld terminated with signal 11 [Segmentation fault], core dumped
compilation terminated.
$ gcc-12 -flto a.c -Wl,--defsym,g=real_g
```

ishitatsuyuki:
Add an undefined symbol to keep a reference to the defsym target. This prevents
elimination by e.g. LTO or gc-sections. A test is added for the case of LTO
dead-code elimination.

Closes #829

### DIFF
--- a/elf/passes.cc
+++ b/elf/passes.cc
@@ -704,7 +704,8 @@ void add_synthetic_symbols(Context<E> &ctx) {
   // Make all synthetic symbols relative ones by associating them to
   // a dummy output section.
   for (Symbol<E> *sym : obj.symbols)
-    sym->set_output_section(ctx.symtab);
+    if (sym->file == &obj)
+      sym->set_output_section(ctx.symtab);
 
   // Handle --defsym symbols.
   for (i64 i = 0; i < ctx.arg.defsyms.size(); i++) {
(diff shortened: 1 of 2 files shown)


rui314:
We don't add undefined symbols for `--undefined` but instead handle arguments for `--undefined` as GC root symbols. I believe we can do the same for `--defsym`'ed symbols.

ishitatsuyuki:
I see what you mean. I feel that synthesizing undef symbols is a cleaner approach though.

The `--undefined` code doesn't handle the case of LTO right now, which is what the linked issue use. In an approach similar to `--undefined`, we need to manually mark defsym symbols as `PREVAILING_DEF`s. On the other hand, with synthesized undef symbols, the symbol resolver just do the "right thing".

What do you think? I wonder if there's any concerns with explicitly synthesizing undefined symbols?

PR #913, opened 2022-12-12
Dockerfile: fix OpenSSL libdir

ZhongRuoyu:
In the current version of Docker image, OpenSSL libraries are installed to `/usr/local/lib64`. It is not present in the default dynamic linker search path, and thus results in the following error when libcrypto is not statically-linked:

`mold: error while loading shared libraries: libcrypto.so.3: cannot open shared object file: No such file or directory`

Fix that by overriding `libdir`, as suggested in https://github.com/openssl/openssl/blob/master/INSTALL.md#libdir. Also, run `ldconfig` to update dynamic linker cache.

Signed-off-by: Ruoyu Zhong <zhongruoyu@outlook.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,9 +13,10 @@ RUN apt-get update && \
   \
   mkdir /openssl && cd /openssl && \
   wget -O- -q https://www.openssl.org/source/openssl-3.0.7.tar.gz | tar --strip-components=1 -xzf - && \
-  ./Configure --prefix=/usr/local && \
+  ./Configure --prefix=/usr/local --libdir=lib && \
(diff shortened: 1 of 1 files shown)


ZhongRuoyu:
> The docker image is to statically link most libraries. I wonder what is your motivation to dynamically link libssl using that docker image.

Actually, I didn't really mean to dynamically link that either. I was using this image to debug #911 (using default configs, so `MOLD_MOSTLY_STATIC` is off), but got the "error while loading shared libraries" while trying to run the executable installed. So it's more of a "now that we have this image, let's make it also work under default config" thing, I think?

PR #938, opened 2022-12-30
Add explanation for `-fuse-ld-mold` in README

arbipher:
This may be useful for new users (e.g. me) who tried `mold` from github [releases](https://github.com/rui314/mold/releases) rather than building and installing from the source.

`-fuse-ld=mold` can only be used when both `mold` and `ld.mold` are in your PATH (See Clang's [source](https://github.com/llvm/llvm-project/blob/a63b7247299ce6edfbbf47c4a2773e5ca7eb7f11/clang/lib/Driver/ToolChain.cpp#L673) on finding the linker). GCC 12.1.0 has the same behavior.

To use an absolute path (rather than adding them to PATH), use `--ld-path=` which is introduced in Clang 12.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ following flags to use `mold` instead of `/usr/bin/ld`. When using
   `-B/usr/libexec/mold` (or `-B/usr/local/libexec/mold`) to GCC.
 
 GCC does not take an absolute path as an argument for `-fuse-ld` though.
-GCC also does not support `--ld-path=`.
+GCC also does not support `--ld-path`.
 
 </details>
 
--- a/README.md
+++ b/README.md
@@ -109,9 +109,12 @@ invoked indirectly by the compiler driver (which is usually `cc`,
 
 If you can specify an additional command line option to your compiler
 driver by modifying build system's config files, add one of the
-following flags to use `mold` instead of `/usr/bin/ld`:
+following flags to use `mold` instead of `/usr/bin/ld`. When using
+`-fuse-ld=mold`, both `mold` and `ld.mold` should be in your path: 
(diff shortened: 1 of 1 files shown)


rui314:
Is that correct? I thought that only `ld.mold` is required to use `-fuse-ld=mold`.

arbipher:
Sorry, you are correct. Only `ld.mold` is needed. I removed `mold` from the PATH, and both `clang` and `gcc` are still satisfied as long as `ld.mold` are there.

Now the following sentence on README may also need an update a bit

```
If you haven't installed mold to any $PATH, you can still pass -fuse-ld=/absolute/path/to/mold to clang to use mold. GCC does not take an absolute path as an argument for -fuse-ld though.
```

arbipher:
The section for rust `"link-arg=-fuse-ld=/path/to/mold"` also looks dubious to me 
but I don't have a rust environment at hand to check it.

rui314:
Actually I don't think we want to mention `--ld-path` because `-fuse-ld` just works both for gcc and clang. Is there any reason to use `--ld-path`?

arbipher:
[`--ld-path` ](https://clang.llvm.org/docs/DiagnosticsReference.html#wfuse-ld-path)

rui314:
`-Wufse-ld-path` seems to be enabled only for `-Wextra`, which enables lots of other warnings. I guess those who deliberately  enable `-Wextra` can fix the warning for `-fuse-ld` by themselves.

arbipher:
I see. I am done with my questions this time.
`mold` is an excellent tool and runs very fast. 
Thank you!

PR #948, opened 2023-01-07
Fix ""s and "sv" cannot find issues

trcrsired:


elf/main.cc
```diff
@@ -373,7 +373,7 @@ static int redo_main(int argc, char **argv, std::string_view target) {
 template <typename E>
 int elf_main(int argc, char **argv) {
   Context<E> ctx;
-
+  using namespace std::string_view_literals;
```
rui314, line 376: I don't think this is necessary because we have it in `mold.h` in the top directory

rui314:
Actually I don't think we need both, as these files include the top directory's `mold.h`.

PR #963, opened 2023-01-13
Add: build deps when compiling mold on Void Linux

rilysh:
Install build dependencies when compiling mold on Void Linux

Signed-off-by: rilysh <luhamailbox@gmail.com>


rui314:
It looks odd that this PR contains two commits. I'll merge anyways, but in general most projects prefer contributors to squash commits into a single commit before sending a PR unless presenting it as a series of commit makes more sense.

rilysh:
Yeah, my bad, actually I shouldn't synchronize the forked one, and because of that the PR got extra another commit.

PR #972, opened 2023-01-22
Update output-file-win32.h

3rav:
Fix include


rui314:
This file has actually been renamed to `common.h`.

rui314:
I committed a fix with a citation to this PR.

PR #980, opened 2023-02-09
[RISCV] Fix R_RISCV_PCREL_HI20

ksco:
An undefined weak symbol should have the value 0, we don't need special handling here. Please correct me if I'm wrong.

### DIFF
--- a/elf/arch-riscv.cc
+++ b/elf/arch-riscv.cc
@@ -353,6 +353,9 @@ void InputSection<E>::apply_reloc_alloc(Context<E> &ctx, u8 *base) {
         write_cjtype(loc, S + A - P);
       } else {
         assert(removed_bytes == 0);
+        // Calling an undefined weak symbol does not make sense.
+        // We make such call into an infinite loop. This should
+        // help debugging of a faulty program.
         u64 val = sym.esym().is_undef_weak() ? 0 : S + A - P;
         check(val, -(1LL << 31), 1LL << 31);
         write_utype(loc, val);
@@ -370,14 +373,7 @@ void InputSection<E>::apply_reloc_alloc(Context<E> &ctx, u8 *base) {
       *(ul32 *)loc = sym.get_tlsgd_addr(ctx) + A - P;
       break;
     case R_RISCV_PCREL_HI20:
-      if (sym.esym().is_undef_weak()) {
-        // Calling an undefined weak symbol does not make sense.
-        // We make such call into an infinite loop. This should
-        // help debugging of a faulty program.
-        *(ul32 *)loc = 0;
-      } else {
-        *(ul32 *)loc = S + A - P;
-      }
+      *(ul32 *)loc = S + A - P;
       break;
     case R_RISCV_PCREL_LO12_I:
     case R_RISCV_PCREL_LO12_S:


ksco:
I found this issue when trying to use `riscv64-unknown-elf-gcc` (Newlib) with `mold`.
The issue is in the `frame_dummy` routine of `crtbegin.o`, the following is the result of `objdump`:

<details>

```
000000000000003c <frame_dummy>:
  3c:	00000797          	auipc	a5,0x0
			3c: R_RISCV_PCREL_HI20	__register_frame_info
			3c: R_RISCV_RELAX	*ABS*
  40:	00078793          	mv	a5,a5
			40: R_RISCV_PCREL_LO12_I	.L0 
			40: R_RISCV_RELAX	*ABS*
  44:	cf89                	beqz	a5,5e <.L8>
			44: R_RISCV_RVC_BRANCH	.L8
  46:	00000597          	auipc	a1,0x0
			46: R_RISCV_PCREL_HI20	object.0
			46: R_RISCV_RELAX	*ABS*
  4a:	00058593          	mv	a1,a1
			4a: R_RISCV_PCREL_LO12_I	.L0 
			4a: R_RISCV_RELAX	*ABS*
  4e:	00000517          	auipc	a0,0x0
			4e: R_RISCV_PCREL_HI20	__EH_FRAME_BEGIN__
			4e: R_RISCV_RELAX	*ABS*
  52:	00050513          	mv	a0,a0
			52: R_RISCV_PCREL_LO12_I	.L0 
			52: R_RISCV_RELAX	*ABS*
  56:	00000317          	auipc	t1,0x0
			56: R_RISCV_CALL	__register_frame_info
			56: R_RISCV_RELAX	*ABS*
  5a:	00030067          	jr	t1 # 56 <frame_dummy+0x1a>

000000000000005e <.L8>:
  5e:	8082                	ret
```

</details>

The `__register_frame_info` is an undefined weak symbol, so its value should be 0.

Without this patch:

<details>

```
00000000002012ca <frame_dummy>:
  2012ca:	00000797          	auipc	a5,0x0
  2012ce:	00078793          	mv	a5,a5
  2012d2:	cf89                	beqz	a5,2012ec <frame_dummy+0x22>
  2012d4:	00005597          	auipc	a1,0x5
  2012d8:	55458593          	addi	a1,a1,1364 # 206828 <object.0>
  2012dc:	ffdff517          	auipc	a0,0xffdff
  2012e0:	d2450513          	addi	a0,a0,-732 # 0 <_DYNAMIC>
  2012e4:	00000317          	auipc	t1,0x0
  2012e8:	00030067          	jr	t1 # 2012e4 <frame_dummy+0x1a>
  2012ec:	8082                	ret
```

</details>

With this patch:

<details>

```
00000000002012ca <frame_dummy>:
  2012ca:	ffdff797          	auipc	a5,0xffdff
  2012ce:	d3678793          	addi	a5,a5,-714 # 0 <_DYNAMIC>
  2012d2:	cf89                	beqz	a5,2012ec <frame_dummy+0x22>
  2012d4:	00005597          	auipc	a1,0x5
  2012d8:	55458593          	addi	a1,a1,1364 # 206828 <object.0>
  2012dc:	ffdff517          	auipc	a0,0xffdff
  2012e0:	d2450513          	addi	a0,a0,-732 # 0 <_DYNAMIC>
  2012e4:	00000317          	auipc	t1,0x0
  2012e8:	00030067          	jr	t1 # 2012e4 <frame_dummy+0x1a>
  2012ec:	8082                	ret
```

</details>

rui314:
I believe this is correct, though it is not easy for me to write a testcase with gcc. gcc seems to always emit a GOT-loading code to obtain an address of a weak undef symbol even with `-fPIE`. That's probably why we haven't noticed this issue before.

ksco:
https://github.com/rui314/mold/commit/56f11e5af5cfadb4c8dc3aa28faec9e96243b141
We could use inline assembly to write the test, I can submit a PR if this is an acceptable way.

Also, I noticed that the spec[1] says: "Any references to possibly-undefined weak symbols should be made indirectly through the GOT as is used for position-independent code. Not doing so is deprecated and a future version of this specification will require using the GOT, not just advise." Maybe this is why gcc always generates GOT-loaded code.

[1] https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#medium-any-code-model

rui314:
If that use is deprecated, you might want to send a patch to glibc so that crtbegin.o doesn't contain a direct reference to a possibly-weakly-defined symbol. I believe crtbegin.o is written in a hand-coded assembly file.

ksco:
It's actually C code[1], and this only happens with gcc targeting Newlib (i.e. `riscv64-unknown-elf-gcc`).

[1] https://github.com/gcc-mirror/gcc/blob/master/libgcc/crtstuff.c#L474

PR #993, opened 2023-02-18
[RISCV] Check object files' compatibility

ksco:


ksco:
ping @rui314

test/elf/riscv64_obj-compatible.sh
```diff
@@ -0,0 +1,22 @@
...
+. $(dirname $0)/common.inc
+
+set -x
+
+cat <<EOF | riscv64-linux-gnu-gcc -o $t/a.o -c -xc -
+void foo() {}
+EOF
+
+# Set `e_flags` to zero to simulate ABI incompatibility
+dd if=/dev/zero of=$t/a.o bs=4 oflag=seek_bytes seek=48 count=1 conv=notrunc
```
rui314, line 11: This is wild; is there a compiler flag or something to specify a float ABI?

ksco: Seems one needs to recompile the whole toolchain to get a different ABI support.

PR #999, opened 2023-02-23
Fix default versioned symbol

### LINKED ISSUE #998: Binary segfaults when using `.symver` on CentOS 7
The following test program crashes (segmentation fault) when building/running on a CentOS 7 x86_64 box (with glibc 2.17, and building with gcc 12 and mold 1.10.1):

```
#include <cmath>

__asm__(".symver pow,pow@GLIBC_2.2.5");

int main(int argc, char** argv)
{
    return std::pow(10, size_t(argc));
}
```

Built with `g++ -Wall -fuse-ld=mold -o prog prog.cpp`

The binary works as expected when linking with any other linker (I tried ld.bfd, lld, and gold).

The issue disappears if I remove the `__asm__` statement.

No issue as well when building and running with mold on a Ubuntu 20.04 box.

ksco:
A default versioned symbol like `foo@@version` should produce `foo` and `foo@version`.

Fix #998

### DIFF
--- a/elf/input-files.cc
+++ b/elf/input-files.cc
@@ -1262,11 +1262,13 @@ void SharedFile<E>::parse(Context<E> &ctx) {
     this->elf_syms2.push_back(esyms[i]);
     this->versyms.push_back(ver);
 
-    if (is_hidden) {
-      std::string_view mangled_name = save_string(
-        ctx, std::string(name) + "@" + std::string(version_strings[ver]));
-      this->symbols.push_back(get_symbol(ctx, mangled_name, name));
-    } else {
+    std::string_view mangled_name = save_string(
+      ctx, std::string(name) + "@" + std::string(version_strings[ver]));
+    this->symbols.push_back(get_symbol(ctx, mangled_name, name));
+
+    if (!is_hidden) {
+      this->elf_syms2.push_back(esyms[i]);
+      this->versyms.push_back(ver);
       this->symbols.push_back(get_symbol(ctx, name));
     }
   }
--- a/elf/passes.cc
+++ b/elf/passes.cc
@@ -1226,23 +1226,6 @@ void claim_unresolved_symbols(Context<E> &ctx) {
         if (!sym.esym().is_undef() || sym.file->priority <= file->priority)
           continue;
 
-      // If a symbol name is in the form of "foo@version", search for
-      // symbol "foo" and check if the symbol has version "version".
-      if (file->has_symver.get(i - file->first_global)) {
-        std::string_view str = file->symbol_strtab.data() + esym.st_name;
-        i64 pos = str.find('@');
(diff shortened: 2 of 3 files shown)


ksco:
I'll submit a test in a separate PR later.

ksco:
ping @rui314

rui314:
The mapping provided by `get_symbol` is one-to-one, meaning that the there's one-to-one correspondence between symbol names passed to `get_symbol` to the returned `Symbol<E>` instance. We don't create two `Symbol<E>` instances for a single symbol name.

The DSO symbol name with the default version is an exception. It may be resolved either as `foo` or `foo@@VERSION`.

I didn't want to break the one-to-one correspondence, so the default version name is handled specially here: https://github.com/rui314/mold/blob/main/elf/passes.cc#L1229-L1244

Do you know why that code didn't work for this particular case?

moncefmechri:
For the record, I tried the proposed patch both on the toy example in the linked issue, as well as on our large binary, and it does seem to solve the issue.

moncefmechri:
> For the record, I tried the proposed patch both on the toy example in the linked issue, as well as on our large binary, and it does seem to solve the issue.

Let me know if there are more patches you would like me to test, happy to help.

ksco:
> Do you know why that code didn't work for this particular case?

https://github.com/rui314/mold/blob/main/elf/input-files.cc#L1270
I think it's because a default versioned symbol will not have any version info according to the above line.

ksco:
If we don't want to break the 1-to-1 correspondence, maybe we can add a "default version" field to `Symbol`.

ksco:
On second thought, I think the original implementation is supposed to work, there must be something wrong. I should look into that and find the root cause.

ksco:
In conclusion, IIUC, the original solution (i.e. [passes.cc#L1229-L1244](https://github.com/rui314/mold/blob/main/elf/passes.cc#L1229-L1244)) is not right. 

For example, `foo@v1` is declared in `a.o`, and `foo@@v1` is defined in `b.so`. 
According to the original solution, `foo@v1` in `a.o` will be inserted into the symbol map with `key = foo@v1, name = foo`, and `foo@@v1` in `b.so` will be inserted into the symbol map with `key = foo, name = foo`. This causes `foo@@v1` to be cleaned during the `resolve_symbols` phase.
So in [passes.cc#L1229-L1244](https://github.com/rui314/mold/blob/main/elf/passes.cc#L1229-L1244), as `foo` has been cleaned, it cannot match `foo@v1` correctly.

rui314:
This isn't an easy problem, allow me a time to think and experiment.

moncefmechri:
Any luck on this? Can I help in any way?

ksco:
Closed as this is already fixed.

Issue #977, opened 2023-02-01
Link failure with LTOed toolchain binaries


dhewg:
When using a gcc (v12.2.0) cross compiler which was build with:
CFLAGS_FOR_TARGET="-flto=auto -ffat-lto-objects"
LDFLAGS_FOR_TARGET="-flto=auto -fuse-linker-plugin"

(There's https://gcc.gnu.org/bugzilla/show_bug.cgi?id=108630, so add `--enable-symvers=no` for now too if attempting that).

I see link errors when building various programs that link libgcc/libatomic/libc/etc:
`mold: error: undefined symbol: __EH_FRAME_BEGIN__.lto_priv.0`

I haven't digged further, so didn't find out what combinations trigger this.

I don't know if the existence of symbols with a name of `__EH_FRAME_BEGIN__.lto_priv.0` is a toolchain bug, or how useful LTOing the cross toolchain's target libraries even is, but the fact is that bfd links the programs in question while mold errors out.

This is the only hit for `lto_priv` in the gcc source: https://github.com/gcc-mirror/gcc/blob/master/gcc/lto/lto-partition.cc#L950
It sounds like mold may need support for "LTO privatized symbols"?

dhewg:
With this gcc patch:

```
diff -ur a/libgcc/Makefile.in b/libgcc/Makefile.in
--- a/libgcc/Makefile.in	2022-08-19 10:09:54.664689148 +0200
+++ b/libgcc/Makefile.in	2023-02-01 21:56:50.561528520 +0100
@@ -301,7 +301,7 @@
 CRTSTUFF_CFLAGS = -O2 $(GCC_CFLAGS) $(INCLUDES) $(MULTILIB_CFLAGS) -g0 \
   $(NO_PIE_CFLAGS) -finhibit-size-directive -fno-inline -fno-exceptions \
   -fno-zero-initialized-in-bss -fno-toplevel-reorder -fno-tree-vectorize \
-  -fbuilding-libgcc -fno-stack-protector $(FORCE_EXPLICIT_EH_REGISTRY) \
+  -fbuilding-libgcc -fno-stack-protector -fno-lto $(FORCE_EXPLICIT_EH_REGISTRY) \
   $(INHIBIT_LIBC_CFLAGS) $(USE_TM_CLONE_REGISTRY)
 
 # Extra flags to use when compiling crt{begin,end}.o.
```

(disables LTO on crt files)
Those mold errors are gone.

I'm not sure if that's the right thing to do though

[...]

PR #1004, opened 2023-03-02
[ELF] Resolve .eh_frame symbols renamed by GCC LTO.

ishitatsuyuki:
Closes: #977.

See commit descriptions.


elf/input-files.cc
```diff
@@ -324,7 +324,6 @@ void ObjectFile<E>::initialize_ehframe_sections(Context<E> &ctx) {
     std::unique_ptr<InputSection<E>> &isec = sections[i];
     if (isec && isec->is_alive && isec->name() == ".eh_frame") {
       read_ehframe(ctx, *isec);
-      isec->is_alive = false;
```
rui314, line 327: If a section is not copied from input files to an output file as-is, I believe we mark it as dead. So we want to keep this code.

dhewg:
This fixes the build error from the linked issue for me, thanks!
Not runtime tested yet though

ishitatsuyuki:
(Currently a bit busy with other stuff, will try to address the review comment soon)

elf/input-files.cc
ishitatsuyuki: I switched to an approach that is similar to what is used for mergeable sections, i.e. killing the sections after symbol resolution.

ishitatsuyuki:
@dhewg Would appreciate retesting the new version (and also testing the linked binary if possible).

dhewg:
I gave it a quick spin, and it builds and works just fine!
Thanks!

ishitatsuyuki:
Thanks for testing. I've moved the PR out of draft.

PR #1018, opened 2023-03-16
[ELF] Skip tests on target mismatch

ZhongRuoyu:
This makes sure the non-matching, target-specific tests are skipped when the shell scripts are directly executed.

### DIFF
--- a/test/elf/riscv64_obj-compatible.sh
+++ b/test/elf/riscv64_obj-compatible.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 . $(dirname $0)/common.inc
 
+[ $MACHINE = riscv64 ] || skip
+
 set -x
 
 cat <<EOF | $CC -o $t/a.o -c -xc -
--- a/test/elf/riscv64_weak-undef.sh
+++ b/test/elf/riscv64_weak-undef.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 . $(dirname $0)/common.inc
 
+[ $MACHINE = riscv64 ] || skip
+
 cat <<EOF | $CC -c -o $t/a.o -xassembler -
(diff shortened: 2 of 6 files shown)


rui314:
`ctest` automatically skips incompatible tests, so adding such checks is redundant. Could you explain a little bit more about the motivation to make this change?

ZhongRuoyu:
> `ctest` automatically skips incompatible tests, so adding such checks is redundant. Could you explain a little bit more about the motivation to make this change?

Yes. I'm packaging `mold` 1.11.0 for Homebrew at https://github.com/Homebrew/homebrew-core/pull/125809. [Here](https://github.com/Homebrew/homebrew-core/blob/master/Formula/mold.rb)'s the formula (Homebrew terminology for package definition). Build and test are in separate steps, each in its own sandboxed environment. It would be more straightforward to run the test scripts in that case (and perhaps trickier if `ctest` is to be used).

ZhongRuoyu:
Speaking of that, is this (direct execution of shell scripts) still considered to be supported now? Direct execution was a story prior to v1.5.0, when Makefile was used. But CMake (and I'd assume `ctest` too) has been the supported way since then.

rui314:
Not really. Direclty running test scripts mostly works, but that's for those who know what they are doing. The official way to run tests is to run `ctest` as explained in the README.

If you want to take the alternative route, you may want to conservatively skip all tests that contains `_` in their filenames.

ZhongRuoyu:
Thanks @rui314. In that case, I'll try adjusting the test for now, because even if `ctest` is used, the same tweaks may still be required due to the separation of build and test. But I'll think about it.

Regarding this PR, please feel free to close it or merge it at your choice. Either would be fine.

PR #1041, opened 2023-05-25
RISC-V TLSDESC support

ishitatsuyuki:
TODO:
- [x] Support relaxation
- [x] Get spec approval (https://github.com/riscv-non-isa/riscv-elf-psabi-doc/pull/373)


elf/elf.h
```diff
@@ -759,6 +760,10 @@ enum : u32 {
   R_RISCV_PLT32 = 59,
   R_RISCV_SET_ULEB128 = 60,
   R_RISCV_SUB_ULEB128 = 61,
+  R_RISCV_TLSDESC_HI20 = 62,
+  R_RISCV_TLSDESC_LOAD_LO12 = 63,
+  R_RISCV_TLSDESC_ADD_LO12 = 64,
+  R_RISCV_TLSDESC_CALL = 65,
```
rui314, line 766: You want to add these relocations to `elf/elf.cc` file too.

ishitatsuyuki: Done.

ishitatsuyuki:
Added relaxation support. Tested with basic assembly, with WIP gas tree (should upload in a few minutes) at https://github.com/ishitatsuyuki/binutils/tree/rv-tlsdesc.

elf/arch-riscv.cc
```diff
@@ -458,6 +462,49 @@ void InputSection<E>::apply_reloc_alloc(Context<E> &ctx, u8 *base) {
...
+    }
+    case R_RISCV_TLSDESC_CALL: {
+      i64 idx2 = find_paired_reloc();
+      const ElfRel<E> &rel2 = rels[idx2];
+      Symbol<E> &sym2 = *file.symbols[rel2.r_sym];
+      u64 S = sym2.get_addr(ctx);
+      u64 A = rel2.r_addend;
+      if (!sym2.has_tlsdesc(ctx)) {
+        u64 val = S + A - ctx.tp_addr;
+        if (sign_extend(val, 11) == val)
```
ishitatsuyuki, line 500: Not satisfied with how we're computing this twice (here and in shrink_section), so I'm open to suggestions for how to deduplicate this logic.

elf/elf.h
```diff
@@ -1960,6 +1971,7 @@ struct RV64BE {
   static constexpr u32 plt_hdr_size = 32;
   static constexpr u32 plt_size = 16;
   static constexpr u32 pltgot_size = 16;
+  static constexpr u32 tlsdesc_words = 4;
```
rui314, line 1974: I wonder if it is still a good idea to inline four words into `.got` instead of reserving two words like for other psABIs. With this optimization, the RISC-V psABI's TLSDESC would diverge from others. My understanding is that, as a return, we can eliminate one indirection for each TLV reference for dlopen()'ed DSOs, which seems like a relatively rare use case. I'm not sure if this micro-optimization was a good trade-off. What do you think?

ishitatsuyuki: I don't have a strong opinion either. For the glibc implementation there was no complications, I just had to assign the values directly instead of calling `_dl_make_tlsdesc_dynamic` which allocates and fills in the structure.

The inline design is in fact simpler to implement, since we don't need to worry about allocation or keeping track of the list of allocations, but this doesn't matter much for cross-arch libcs since they will have a common helper to do that anyway.

Perhaps we could ask on the spec PR what others think of this.

rui314:
> The inline design is in fact simpler to implement, since we don't need to worry about allocation or keeping track of the list of allocations, but this doesn't matter much for cross-arch libcs since they will have a common helper to do that anyway.

That's my feeling too. If it had four slots from the beginning, it would have been simpler. But we already have code for two-slots design, so it doesn't make much difference anyway. And if it doesn't make much difference, I'd make it consist with others than diverging from them.

ishitatsuyuki:
We have spec approved now. I'll rebase and undraft this soon.

ishitatsuyuki:
TODO: We should probably get tests on this (right now I do it manually with my PoC). I'd expect portable C ones to work once we have compiler support shipped.

ishitatsuyuki:
Actually, I think I need to look a bit more how to deal with IE relaxation (now that it's in the common framework).

rui314:
Yup. But as long as it works with your wip patches, I don't care too much about writing tests in assembly. The feature will eventually be tested once your patches are merged.

rui314:
As to relaxation, you can just mimic what we are doing for ARM64, can't you?

ishitatsuyuki:
> As to relaxation, you can just mimic what we are doing for ARM64, can't you?

Yeah, it should be doable.

ishitatsuyuki:
IE is now working, I've also got the testsuite to pass with a few hacks to use locally built toolchain.

elf/arch-riscv.cc
```diff
@@ -459,6 +463,70 @@ void InputSection<E>::apply_reloc_alloc(Context<E> &ctx, u8 *base) {
...
+      if (sym2.has_tlsdesc(ctx)) {
+        u64 desc = sym2.get_tlsdesc_addr(ctx);
+        write_itype(loc, desc + A - P);
+      } else if (sym2.has_gottp(ctx)) {
+        u64 desc = sym2.get_gottp_addr(ctx);
+        *(ul32 *)loc = 0x517; // auipc a0,<hi20>
+        write_utype(loc, desc + A - P);
+      } else {
+        if (removed_bytes == 0) {
+          *(ul32 *)loc = 0x537; // lui a0,<hi20>
```
rui314, line 501: Not directly related to this change, but I think there's one more relaxation opportunity here. If the hi20 bits are small enough, you can use C.LUI instead of LUI if the C extension is available, which should save two bytes.

PR #1053, opened 2023-06-19
Use correct location of global config for Nim

ire4ever1190:
Correct path is listed [here](https://nim-lang.org/docs/nimc.html#compiler-usage-configuration-files) (See number 2).

While it says `nim.cfg` in the docs, `config.nims` is the correct file for this purpose

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,7 +167,7 @@ example, `gcc` is used as the linker driver. Use the `-fuse-ld` option if your
 GCC is recent enough to recognize this option.
 
 If you want to use mold for all projects, add the above snippet to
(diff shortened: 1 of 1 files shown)


rui314:
> While it says nim.cfg in the docs, config.nims is the correct file for this purpose

OK, but that confused me. Is that the official documentation plainly wrong and needs fixing?

ire4ever1190:
Oh sorry should've been more specific.

Basically `.cfg` is a basic config type that doesn't let you run Nim code (which the example shows).
But both are valid config file types

PR #1058, opened 2023-07-04
Fix support for R_RISCV_SET_ULEB128 and R_RISCV_SUB_ULEB128

andreas-schwab:
Handle allocated sections, write the full value into the location and only
write the final value.

### DIFF
--- a/elf/arch-riscv.cc
+++ b/elf/arch-riscv.cc
@@ -127,6 +127,7 @@ static void overwrite_uleb(u8 *loc, u64 val) {
     *loc++ = 0b1000'0000 | (val & 0b0111'1111);
     val >>= 7;
   }
+  *loc = val & 0b0111'1111;
 }
 
 // Returns the rd register of an R/I/U/J-type instruction.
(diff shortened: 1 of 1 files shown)


rui314:
How did you verify this fix?

andreas-schwab:
https://build.opensuse.org/package/live_build_log/home:Andreas_Schwab:riscv:mold/mold/standard/riscv64

andreas-schwab:
You cannot write a full address into the location, if it only has enough
room for the difference.

rui314:
I mean, is there any program that you found didn't work without this change?

andreas-schwab:
https://build.opensuse.org/package/live_build_log/devel:gcc:next/mold/openSUSE_Factory_RISCV/riscv64

andreas-schwab:
I was mislead by the brokenness of overwrite_uleb.  The difference will
work out right even without being able to write the full address.

rui314:
Can you read https://github.com/rui314/mold/blob/main/CONTRIBUTING.md and sign your commit?

andreas-schwab:
It is confusing that almost none of the commits are signed.

rui314:
I don't need to sign the contributor agreement that I worded because "I" and "YOU" are the same person, but I want third party contributors to sign.

PR #1072, opened 2023-07-29
Fix typo in mold.h

eltociear:
addres -> address


rui314:
We don't usually care too much about typos in comments. I'll fix them all at once later.

PR #1119, opened 2023-09-30
Enable strict compiler warnings by default

sicherha:
I'm a fan of leveraging compiler warnings as a lightweight means of static code analysis. For instance, the bug fixed by commit 49f38828d1eec8cd37507792b83254f2c4d64b90 could have been detected and reported by GCC.

Obviously, there is a trade-off between the uncovering of real issues on the one hand and unhelpful warning spam on the other hand. This PR is meant to serve as a ground for discussions.

My personal preference is to go all the way and set `-Wall -Wextra` by default, then deal with the fallout in the following ways:
1. For all legitimate warnings, fix the code.
2. For dubious warnings that are sufficiently rare, massage the code to avoid them.
3. For excessive spam of a certain warning category, selectively disable that category (`-Wno-...`).

For the mold code base, setting `-Wall -Wextra` has had the following consequences (which are addressed by the three additional commits in this PR):
* Several unused parameters being identified. *- I have remove these parameters.*
* A bunch of additional 'unused parameter' warnings for methods whose interface cannot be changed. *- I have suppressed them by commenting-out their names.*
* Two warnings related to differing types in a ternary operator. *- I have suppressed them by adding explicit casts; the warnings per se seem uncritical.*

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,9 +64,11 @@ if(NOT "${CMAKE_CXX_COMPILER_FRONTEND_VARIANT}" STREQUAL "MSVC")
     -fno-exceptions
     -fno-unwind-tables
     -fno-asynchronous-unwind-tables
+    -Wall
+    -Wextra
     -Wno-sign-compare
     -Wno-unused-function
-    -Wunused-variable
+    -Wno-unused-parameter
     -ggnu-pubnames)
 endif()
 
--- a/elf/arch-ppc64v1.cc
+++ b/elf/arch-ppc64v1.cc
@@ -521,7 +521,7 @@ struct OpdSymbol {
 };
 
 static Symbol<E> *
-get_opd_sym_at(Context<E> &ctx, std::span<OpdSymbol> syms, u64 offset) {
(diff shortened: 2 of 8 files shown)


rui314:
Enabling more compiler warnings are I think in a good direction, but I don't like to enable the unused parameter warning. It's too noisy for our code base and doesn't seem to provide a large value for us.

sicherha:
I have dropped the commit that out-comments unused parameters, and added `-Wno-unused-parameter` to the compile options instead.

rui314:
I also don't like to remove `ctx` parameters from functions even if it is necessary. `ctx` is essentially a global variable passed through all functions.

sicherha:
> I also don't like to remove `ctx` parameters from functions even if it is necessary. `ctx` is essentially a global variable passed through all functions.

All right. Do you also want to keep `ctx` for module-local (i.e. `static`) helper functions?

rui314:
It's a personal preference, and I think I'm okay in either way. I'll merge this. Thank you for doing this!

PR #1121, opened 2023-10-03
[ELF] Move main() and redo_main() into a separate compile unit.

ishitatsuyuki:
Despite the extern template items Clang 14 seems to insist on instantiating elf_main(), with the hilarious consequence of the function being duplicated quadratically to the number of archs.

Move this to a different CU so that the compiler cannot possibly instantiate it more than once. This significantly speeds up compilation related to main.cc.


ishitatsuyuki:
v2: Removed extern template items that are now redundant (the compiler cannot instantiate them anyway).

rui314:
Thanks!

PR #1136, opened 2023-10-26
Fix `aarch64_variant-pcs.sh` test on Linux

sicherha:
The test expects the output format given by `llvm-readelf`:
```
5: 0000000000010698     0 FUNC    GLOBAL DEFAULT [VARIANT_PCS]   13 foo
```

However, in GNU's `readelf` implementation the symbol name comes first:
```
5: 0000000000010698     0 FUNC    GLOBAL DEFAULT   13 foo  [VARIANT_PCS]
```

Replace the regular expression with two `grep` calls piped together to support both output formats.

### DIFF
--- a/test/elf/aarch64_variant-pcs.sh
+++ b/test/elf/aarch64_variant-pcs.sh
@@ -12,7 +12,7 @@ foo:
 EOF
 
 $CC -B. -shared -o $t/b.so $t/a.o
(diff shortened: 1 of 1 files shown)


rui314:
My readelf is `GNU readelf (GNU Binutils for Ubuntu) 2.38` so maybe they have changed the format at some point? Anyways LGTM.

PR #1147, opened 2023-11-14
Add `--trace-symbol` to the help message

Saldivarcher:
Noticed that this command line option was missing from the help message.


rui314:
Thanks!

PR #1154, opened 2023-11-22
mold-wrapper: also wrap posix_spawnp

andreas-schwab:
Fixes issue 1153.

Signed-off-by: Andreas Schwab <schwab@suse.de>


rui314:
Thanks!

PR #1161, opened 2023-12-05
Use fallocate when creating a new output

ishitatsuyuki:
Using ftruncate() creates holes inside files, that on first (write) page fault gets allocated an actual disk block.  There are several disadvantages with this:
- Allocating block one-by-one is slower than doing it all at once.
- As we write to the output in parallel the write order is not exactly sequential.  The allocated blocks can also be out-of-order and make the on-disk representation fragmented.

Instead, use posix_fallocate() which allocates a contiguous on-disk block (if possible).  It's arguably a better fit for linker output, since holes are unlikely to be present after writing all the output (any contiguous zeroes should just go into .bss).

I benchmarked this by linking mold with debug symbols.  On an 8-core desktop, it is 5% (1.0s -> 0.95s) faster, but on a 64-core workstation it is actually 40% slower (0.37s -> 0.52s).  The total amount of System time is reduced (6.7s -> 5.2s), however.  I suspect that the reduced page fault work ends up causing more lock contention.

Despite potential regression on high-core systems I still think this change is justified because of the better on-disk layout.

### DIFF
--- a/common/output-file-unix.h
+++ b/common/output-file-unix.h
@@ -14,6 +14,23 @@ inline u32 get_umask() {
   return orig_umask;
 }
 
+// Resize and allocate space for a file.
+// File size should be 0 before calling, as this function does not handle shrinking.
+template <typename Context>
+static bool
+allocate_file(Context &ctx, int fd, i64 filesize) {
+#if defined(__linux__)
+  if (fallocate(fd, 0, 0, filesize) == 0)
+    return true;
+#endif
+
+  if (ftruncate(fd, filesize) == -1) {
+    Fatal(ctx) << "ftruncate failed: " << errno_string();
+    return false;
+  }
+  return true;
+}
+
 template <typename Context>
 static std::pair<i64, char *>
 open_or_create_file(Context &ctx, std::string path, i64 filesize, i64 perm) {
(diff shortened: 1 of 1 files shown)


rui314:
The performance regression on the high core count machine is concerning as our focus is to speed up build time on such powerful machine. What if we consider an output file as 16 shards and call fallocate on each shard simultaneously from multiple threads?

Besides that, I believe we should use fallocate rather than posix_fallocate. As far as I know, posix_fallocate write to the entire file if fallocate is not supported by the underlying filesystem. If that happens, it'd be very slow.

ishitatsuyuki:
Actually, it seems the slowdown is from some regression between 2.3.3 and 2.4.0 (I benched against the system mold which turns out to be outdated). The `User` column has increased significantly for `.debug_*` sections. I'll debug this issue separately.

Interesting note about the fallback behavior.  Since `posix_fallocate` seems to be absent for macOS anyway, I'll make it so that we try `fallocate` first if it exists and succeed, and otherwise call `ftruncate`.

ishitatsuyuki:
Welp, after a clean rebuild I can't reproduce the slowness anymore. I probably messed up my build CFLAGS or something.

The good news is that performance is neutral after doing the benchmarking again (both around 0.36s, well within the margin of error).  I've removed the performance claims from the commit message.

rui314:
I believe we still need to call `ftruncate` before or after `fallocate` in case the existing file is larger than the new size. I'll make the change after merging this commit.

ishitatsuyuki:
The path for reusing an existing file is unchanged and still uses ftruncate directly.  But I guess using fallocate can make sense for it too.

rui314:
Ah, you are right. I missed that one.

PR #1170, opened 2023-12-31
Update README.md

mominul:
GCC doesn't accept a path as an argument to `-fuse-ld`. It errors out by
```
cc: error: unrecognized command-line option '-fuse-ld=/usr/bin/mold'
```

Passing `-fuse-ld=mold` is enough.

Tested with
```
gcc (GCC) 13.2.1 20231205 (Red Hat 13.2.1-6)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ may be able to remove the `linker = "clang"` line.
 
 ```toml
 [target.x86_64-unknown-linux-gnu]
-rustflags = ["-C", "link-arg=-fuse-ld=/path/to/mold"]
+rustflags = ["-C", "link-arg=-fuse-ld=mold"]
 ```
 
 If you want to use mold for all projects, add the above snippet to


README.md
```diff
@@ -143,7 +143,7 @@ may be able to remove the `linker = "clang"` line.
 
 ```toml
 [target.x86_64-unknown-linux-gnu]
-rustflags = ["-C", "link-arg=-fuse-ld=/path/to/mold"]
+rustflags = ["-C", "link-arg=-fuse-ld=mold"]
```
pothos, line 146: It seems that gcc doesn't support the path, hence I would change `may be able to remove the `linker = "clang"` line.` to `may be able to remove the `linker = "clang"` line and change the '-fuse-ld=' setting to 'mold'.`

Ordoviz (changes requested):
Please replace `/path/to/mold` here with `mold` as well. See https://searchfox.org/llvm/source/clang/include/clang/Basic/DiagnosticDriverKinds.td#640

PR #1184, opened 2024-02-01
Revert "Handle R_RISCV_PCREL_HI20 as a PLT-generating reloc"

### LINKED ISSUE #1182: mold refused to link GP-setup code for RISC-V
### Description

While building musl for riscv64 with mold as linker, mold refused to link the GP-setup code.

### A minimal reproduce

This assembly snippet is created from [`arch/riscv64/crt_arch.h`](https://git.musl-libc.org/cgit/musl/tree/arch/riscv64/crt_arch.h?h=v1.2.4&id=f5f55d6589940fd2c2188d76686efe3a530e64e0), which is included by [`ldso/dlstart.c`](https://git.musl-libc.org/cgit/musl/tree/ldso/dlstart.c?h=v1.2.4&id=f5f55d6589940fd2c2188d76686efe3a530e64e0). Save this snippet as `dlstart.s`.

```asm
.section .sdata,"aw"

.text
.global _dlstart
.type _dlstart,%function
_dlstart:
.weak __global_pointer$
.hidden __global_pointer$
.option push
.option norelax
        lla     gp, __global_pointer$
.option pop
        mv      a0, sp
.weak _DYNAMIC
.hidden _DYNAMIC
        lla     a1, _DYNAMIC
        andi    sp, sp, -16
        tail    _dlstart_c

.type _dlstart_c,%function
_dlstart_c:
        # The actual code here does not matter.
        ret
```

Then compile and link.

```
$ clang --target=riscv64 -march=rv64g dlstart.s -c -o dlstart.o
$ ld.lld -e _dlstart -shared dlstart.o -o fake_libc.so # no error
$ mold -shared -e _dlstart dlstart.o -o fake_libc.so
mold: error: dlstart.o:(.text): R_RISCV_PCREL_HI20 relocation at offset 0x0 against symbol `__global_pointer$' can not be used; recompile with -fno-PIC
```

Note that the GP-setup code in `_dlstart` do follows [the recommended way to initialize the gp register in RISC-V ELF specification](https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/v1.0/riscv-elf.adoc#user-content-gp-relax), and the PIC option for compiler doesn't affect the content of `dlstart.o`.

### More information

With `git bisect` I found mold refuses to link such input after 5fb6091fb7242a5654729cb52295324166d693dc.

piggynl:
This reverts the functional change of commit 5fb6091fb7242a5654729cb52295324166d693dc, and adds `test/elf/riscv64_gp-setup.sh`.

Fixes #1182 "mold refused to link GP-setup code for RISC-V".


rui314:
It looks like it would break CI. Can you take a look?

piggynl:
Hi, the CI should be fine now. I found we can't simply revert 5fb6091fb7242a5654729cb52295324166d693dc, because we still need to generate PLT for R_RISCV_PCREL_HI20. Changing the previous ERROR action to NONE is enough to fix the issue.

piggynl:
Hi, what do you think about this change? for fixing #1182.

rui314:
I think ERROR is logically the correct behavior here, so we need to fix it in a different way. I'm trying to figure out how.

piggynl:
Thanks! I'm closing this PR then.

PR #1186, opened 2024-02-01
illumos: Treat absolute symbols specially

### LINKED ISSUE #1183: Adding illumos support
Hi there. I'm floating a patch that enables mold-linked binaries to run on illumos, where I'm using it to link large Rust programs. I'd like to get this upstreamed, but am not sure the current patch I have fits the overall structure of the linker particularly well. What's the best way to proceed? Send a PR and work through the review process, or try to pre-arrange a path via an issue (perhaps this one?) or something else entirely? Thanks!

The specifics are that illumos appears to treat symbols with st_shndx=SHN_ABS and st_value=0 specially, so that they should be imported in e.g. `compute_import_export`, even though absolute symbols are ordinarily skipped. Right now, I address this by adding an inline function with an `#ifdef` for the illumos case; I think for non-illumos platforms dead-code elimination probably elides it entirely.

dancrossnyc:
illumos is a descendent of Solaris/OpenSolaris, and treats symbols with shndx=SHN_ABS and st_value=0 specially, such that they should be marked imported.  This change adds a special case for this behavior, wrapped up in a function that #ifdef's out on other platforms.  With this patch, executables linked with mold on illumos run; without it, they do not.

The original behavior was spotted by Luqman Aden (@luqmana). This also patches a small bogon in the third-party `mimalloc` code.  I've sent that same fix upstream, but so far the maintainer has not resonded, so replicate here.

Fixes #1183

### DIFF
--- a/elf/passes.cc
+++ b/elf/passes.cc
@@ -1783,6 +1783,20 @@ void parse_symbol_version(Context<E> &ctx) {
   });
 }
 
+namespace {
+template <typename E>
+inline bool should_import_absolute_sym(const Symbol<E> *sym) {
+#ifdef __sun
+  // On illumos, symbols with shndx=SHN_ABS and st_value=0 are special
+  // and should be imported.
+  return sym->value == 0;
+#else
+  (void)sym;
+  return false;
+#endif
+}
+} // anonymous namespace
+
 template <typename E>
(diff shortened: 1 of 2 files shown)


rui314:
Or maybe we can just remove `if (!sym->is_absolute())` completely?

Issue #456, opened 2022-04-21
mold (1.2) fails to run with mimalloc (2.0.6) with Mapsize overflow error on FreeBSD 14-CURRENT


koobs:
## Summary 
After upgrading the FreeBSD `devel/mimalloc` port version from `2.0.5` to `2.0.6` on a recent FreeBSD 14-CURRENT build, mold fails to run, outputting the following on invocation:

```
[koobs@140-CURRENT-amd64-564d:/usr/home/koobs/repos/freebsd/ports/devel/mimalloc] mold
Mapsize overflow
Mapsize overflow
zsh: exec format error: mold
```

This appears to be related to a recent [freebsd base commit bf83941638](https://cgit.freebsd.org/src/commit/?id=bf839416381cb9f63a8a82ea6e897a22830a8009) by @kostikbel at the end of last year via a review [1] that is not publicly available 

## Reproduction Environment / Details

* **FreeBSD** 14.0-CURRENT main-n254374-4fd141c7d7a GENERIC-NODEBUG amd64
* **mold** 1.2
* **mimalloc** 2.0.6

**Note**

 * The issue does **not** appear reproducible with mold `1.2` with mimalloc `2.0.5`
 * The issue does **not** appear reproducible with mold `1.1.1` with mimalloc `2.0.6`

## Steps to Reproduce

* Build mold `1.2` with mimalloc `2.0.6` on FreeBSD CURRENT
* Run the `mold` command

## References

[1] https://reviews.freebsd.org/D33359

rui314:
Thank you for your report. Since it's not easy to set up an environment to reproduce the issue, do you mind if I ask you to help me debug this?

1. Can you get a stacktrace of mold when it crashes?
2. If 1.1.1 is OK but 1.2.0 isn't, there might be a regression introduced at some point between the two. Can you find it by `git bisect`?

kostikbel:
The 'Mapsize overflow' error comes from the image activator.  Basically, for some binary, the total size of the segments to mmap is too large.

Find the binary that causes the problem and put it somewhere so that I can take a look at it.

rui314:
mold exclusively uses mmap for all file IO, and it can handle multi-gibibyte input files and output files in a few seconds. This usage pattern may be unique.

koobs:
@rui314 I'll do what I can this end. It may be worth you testing a branch with thirdparty updating mimalloc to 2.0.6 to see if anything interesting comes up in CI.

[...]

PR #1187, opened 2024-02-03
Feat: use accumulate address on sections with SHT_NOBITS

aokblast:
I fix the bug in #456.

The reason is caused by the offset of sections with SHT_NOBITS. 

As the following picture shows:

The left side is the output of mold and the right side is the output of lld. We can see that file offset of the sections with SHT_NOTIBS is all zero in mold. This cause   the sections enter this [`function`](https://github.com/freebsd/freebsd-src/blob/main/contrib/elftoolchain/elfcopy/segments.c#L54) in FreeBSD strip returned ``unloaded=0`` thus discard in the segment as the second picture shows.

20240203_22h23m53s_grim

This pictures .bss is discarded in left side.

20240203_22h28m32s_grim

### DIFF
--- a/elf/passes.cc
+++ b/elf/passes.cc
@@ -2457,6 +2457,8 @@ static i64 set_file_offsets(Context<E> &ctx) {
     }
 
     if (first.shdr.sh_type == SHT_NOBITS) {
+      fileoff = align_to(fileoff, first.shdr.sh_addralign);
+      first.shdr.sh_offset = fileoff;
       i++;
       continue;
     }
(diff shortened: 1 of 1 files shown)


aokblast:
After all, I can strip and run the binary after strip

rui314:
Isn't it more like a bug in bsd's strip than that in mold?

aokblast:
I don't know but I have tested on gold, lld, and bfd. They all have same behaviors. I think we should not call it a bug. It is an ambiguity on the design of ELF format.

PR #1198, opened 2024-02-20
Minor fixes for the Windows build

nga888:
Removes some Windows build warnings.

Fixes conversion of Windows FILETIME to nsec in get_usage() used by
TimerRecord.

### DIFF
--- a/common/main.cc
+++ b/common/main.cc
@@ -20,6 +20,11 @@
 # include <unistd.h>
 #endif
 
+#ifdef _WIN32
+# define unlink _unlink
+# define write _write
+#endif
+
 namespace mold {
 
 std::string mold_version_string = MOLD_VERSION;
--- a/elf/cmdline.cc
+++ b/elf/cmdline.cc
@@ -9,7 +9,7 @@
 #include <unordered_set>
 
 #ifdef _WIN32
-# define _isatty isatty
+# define isatty _isatty
 # define STDERR_FILENO (_fileno(stderr))
 #else
 # include <unistd.h>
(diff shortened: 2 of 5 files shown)


rui314:
Our windows builder is always able to build mold. What was broken in your build?

nga888:
> Our windows builder is always able to build mold. What was broken in your build?

Nothing was "broken" with the build itself but these changes reduce the warnings and fixes an actual code issue with the conversion in `get_usage()` from `common/perf.cc`.

nga888:
I've expanded the commit message and the above description to clarify the intent.

common/main.cc
```diff
@@ -20,6 +20,11 @@
 # include <unistd.h>
 #endif
 
+#ifdef _WIN32
+# define unlink _unlink
+# define write _write
```
rui314, line 25: What's the error are you trying to fix with these macros?

elf/cmdline.cc
```diff
@@ -9,7 +9,7 @@
 #include <unordered_set>
 
 #ifdef _WIN32
-# define _isatty isatty
+# define isatty _isatty
```
rui314, line 12: Ditto

nga888: These, and the above for `common/main.cc`, are to prevent some of the following types of warning:

> warning: 'write' is deprecated: The POSIX name for this item is deprecated. Instead, use the ISO C and C++ conformant name: _write. See online help for details. [-Wdeprecated-declarations]

I noticed the existing use of such definitions in some of the source, although they appeared to be the wrong way around in some instances, like the above.

PR #1199, opened 2024-02-21
Fixes and improvements to Windows file handling and memory mapping

nga888:
Prior to this change, links on Windows were failing with:

  mold: fatal: <input file>: CreateFileMapping failed: 5

"5" is the Windows error code for ERROR_ACCESS_DENIED.

Use native Windows file API in MappedFile. Create read only mapping with FILE_MAP_COPY view because the file mapping contents may be modified in process but should not be written back.

On Windows, it is important that files are opened in "binary" mode to avoid automatic "text" mode conversions which can result in "corrupted" input/output.

### DIFF
--- a/common/common.h
+++ b/common/common.h
@@ -756,10 +756,19 @@ public:
     if (this->path == "-") {
       fp = stdout;
     } else {
+#ifdef _WIN32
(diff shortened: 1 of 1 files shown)


nga888:
I've rebased and made a slight improvement to the MallocOutputFile handling of file permissions for Windows.

PR #1201, opened 2024-02-23
Add "basic" support for memory mapping to Windows OutputFile

nga888:
This is a "basic" implementation that does not make use of intermediate temporary files.


rui314:
How much performance difference did you observe with the change. Generally I want to keep Windows-specific code as small as possible, so unless this new code makes a big difference, I wouldn't make that change.

nga888:
Ping.

PR #1206, opened 2024-02-27
Do not try to run executables created with --section-order

andreas-schwab:
Executables with an unmapped ELF header crash when run in qemu.

Signed-off-by: Andreas Schwab <schwab@suse.de>

### DIFF
--- a/test/elf/section-order.sh
+++ b/test/elf/section-order.sh
@@ -1,9 +1,6 @@
 #!/bin/bash
 . $(dirname $0)/common.inc
 
-# qemu crashes if the ELF header is not mapped to memory
-[ -z "$QEMU" ] || skip
-
 cat <<EOF | $CC -o $t/a.o -c -xc -fno-PIC $flags -
 #include <stdio.h>
 
@@ -17,7 +14,6 @@ EOF
 
 $CC -B. -o $t/exe1 $t/a.o -no-pie \
   -Wl,--section-order='=0x100000 PHDR =0x200000 .fn2 TEXT =0x300000 .fn1 DATA BSS RODATA'
-$QEMU $t/exe1 | grep -q Hello
 
 readelf -SW $t/exe1 | grep -q '\.fn2 .*00200000'
 readelf -SW $t/exe1 | grep -q '\.fn1 .*00300000'


rui314:
It doesn't crash for me. What version of Qemu are you using?

andreas-schwab:
This is qemu 8.2.0.

andreas-schwab:
It's still crashing with both 8.2.2 and current master

test/elf/section-order.sh
```diff
@@ -1,9 +1,6 @@
 #!/bin/bash
 . $(dirname $0)/common.inc
 
-# qemu crashes if the ELF header is not mapped to memory
-[ -z "$QEMU" ] || skip
```
rui314, line 5: I'm confused with this change. This line prevents the rest of the script to be executed if $QEMU is non-empty. In other words, this test will be skipped if it is running with QEMU. Isn't it the opposite of what you are trying to do?

andreas-schwab:
That variable is irrelevant and not used.  The use of qemu is implicit.

rui314:
mold tests always explicitly use qemu for non-native processors. Did you mean that these tests don't pass in a QEMU full system emulation (as opposed to qemu-user)?

andreas-schwab:
The tests are run with implicit emulation, as if natively.

rui314:
Could you explain a little bit more about how to reproduce the issue? I don't want to disable executing the output file because that's an important step of this test.

mold tests are written so that all native tests are run natively while all cross-build tests are explicitly run using Qemu. We do not rely on implicit emulation.

andreas-schwab:
https://build.opensuse.org/package/live_build_log/devel:gcc:next/mold/openSUSE_Factory_RISCV/riscv64

rui314:
What is the environment in which this build and tests ran? I'm not a SuSE user, and I'm not familiar with this build bot as well.

andreas-schwab:
It's built as if native, but with binfmt_misc setup for qemu linux-user.

andreas-schwab:
https://en.opensuse.org/openSUSE:RISC-V

rui314:
Is there any signal such as an environment variable to detect if the process is running under qemu-user emulation? If there's a signal, I'll use it to disable the test.

andreas-schwab:
$ head -n 6 /proc/cpuinfo 
processor       : 0
hart            : 0
isa             : rv64imafdcvh_zicbom_zicboz_zicond_zicntr_zicsr_zifencei_zihintntl_zihintpause_zihpm_zmmul_zawrs_zfa_zfh_zfhmin_zca_zcb_zcd_zba_zbb_zbc_zbkb_zbkc_zbkx_zbs_zk_zkn_zknd_zkne_zknh_zkr_zks_zksed_zksh_zkt_zvbb_zvbc_zve32f_zve64f_zve64d_zvkb_zvkg_zvkn_zvknc_zvkned_zvkng_zvknha_zvknhb_zvks_zvksc_zvksed_zvksg_zvksh_zvkt_smepmp_smstateen_sscofpmf_sstc_svadu_svinval_svnapot_svpbmt
mmu             : sv48
uarch           : qemu

rui314:
Thanks. The above change should make this test to be skipped if it's running on qemu.

marxin:
I can confirm the tests are green now: https://build.opensuse.org/package/show/devel:gcc:next/mold

PR #1217, opened 2024-03-14
Fix search order in INPUT and GROUP of script

demin-han:
The scenario is from incremental cross compiling gcc. libgcc_s.so is a ld script in my OS.
  GROUP ( libgcc_s.so.1 -lgcc )
The path of libgcc_s.so.1 is not absolute.
When current directory contains libgcc_s.so.1 of other ISA, We will get target of other ISA and report "skipping incompatible file".

As ld manual shows:
If a sysroot prefix is not used then the linker will try to open the file in the directory containing the linker script. If it is not found the linker will then search the current directory. If it is still not found the linker will search through the archive library search path.

### DIFF
--- a/elf/linker-script.cc
+++ b/elf/linker-script.cc
@@ -170,6 +170,14 @@ static MappedFile<Context<E>> *resolve_path(Context<E> &ctx, std::string_view to
   if (str.starts_with("-l"))
     return find_library(ctx, str.substr(2));
 
+  // Try to open the file in the directory containing the linker script.
+  auto parent_path = [](std::string_view path) {
+    return filepath(path).lexically_normal().parent_path().string();
+  };
+  std::string path = parent_path(ctx.script_file->name) + "/" + str;
+  if (MappedFile<Context<E>> *mf = open_library(ctx, path))
+    return mf;
+
   if (MappedFile<Context<E>> *mf = open_library(ctx, str))
     return mf;
 
@@ -252,8 +260,7 @@ get_script_output_type(Context<E> &ctx, MappedFile<Context<E>> *mf) {
 
   if (tok.size() >= 3 && (tok[0] == "INPUT" || tok[0] == "GROUP") &&
       tok[1] == "(")
-    if (MappedFile<Context<E>> *mf =
-        MappedFile<Context<E>>::open(ctx, std::string(unquote(tok[2]))))
+    if (MappedFile<Context<E>> *mf = resolve_path(ctx, tok[2]))
       return get_machine_type(ctx, mf);
 
   return "";


demin-han:
CI failed because of "No space left on device"

demin-han:
@rui314 
linker-script-relocatable.sh failed in all other ISAs. 
The reason:
open_library checks machine_type. 
resolve_path calls open_library.
get_script_output_type reuses resolve_path in this pr. 
deduce_machine_type failed first.

which is the best way to solve this?
1. add check_machine parameter to functions or
2. add check_machine flag to Context

demin-han:
> Does the above commit fix the problem for you?
   Yes, the PR can fix the issue mention in comments.
   But, CI failed with one testcase, the reason had explained in above reply.

   I think it's necessary to reuse "resolve_path" in get_script_output_type function to
   meet search semantic of INPUT or GROUP.
  
   Now , I can't determine the better way to solved the deduce_machine_type issue.
   At that point, we have not determined the machine type, but "open_library" checks machine type.

demin-han:
> Does the above commit fix the problem for you?

No, the commit not fix the problem.

get_script_output_type still use "open" function.

`  if (tok.size() >= 3 && (tok[0] == "INPUT" || tok[0] == "GROUP") &&
      tok[1] == "(")
    if (MappedFile<Context<E>> *mf =
        MappedFile<Context<E>>::open(ctx, std::string(unquote(tok[2]))))
      return get_machine_type(ctx, mf);`

rui314:
Can you write a test to demonstrate the issue?

demin-han:
> Can you write a test to demonstrate the issue?

My OS is Manjaro. The minimal requirements as following:
1. the /usr/lib/libgcc_s.so should contains following content:
   /* GNU ld script
   Use the shared library, but some functions are only in
   the static library.  */
   GROUP ( libgcc_s.so.1 -lgcc )
2. mkdir test
3. write a test.c in test directory
`int main() {}`
5. copy libgcc_s.so.1 of riscv64(or other ISA) to test directory
6. gcc -fuse-lld=mold test.c
7.  following error reported:
    mold: warning: /usr/lib/libgcc_s.so: skipping incompatible file riscv64 62
    mold: warning: /lib/libgcc_s.so: skipping incompatible file riscv64 62
    mold: warning: /usr/lib/libgcc_s.so: skipping incompatible file riscv64 62
    mold: warning: /usr/lib//libgcc_s.so: skipping incompatible file riscv64 62
    mold: fatal: library not found: gcc_s

The call trace :
mold::MappedFile<mold::elf::Context<mold::elf::X86_64> >::open(mold::elf::Context<mold::elf::X86_64>&, 
                                                                     std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) 
mold::elf::get_script_output_type<mold::elf::X86_64>(mold::elf::Context<mold::elf::X86_64>&, 
                                                             mold::MappedFile<mold::elf::Context<mold::elf::X86_64> >*) 
mold::elf::get_machine_type<mold::elf::X86_64>(mold::elf::Context<mold::elf::X86_64>&, 
                                                             mold::MappedFile<mold::elf::Context<mold::elf::X86_64> >*) 
mold::elf::open_library<mold::elf::X86_64>(mold::elf::Context<mold::elf::X86_64>&, std::__cxx11::basic_string<char, 
                                                                             std::char_traits<char>, std::allocator<char> >)
mold::elf::find_library<mold::elf::X86_64>(mold::elf::Context<mold::elf::X86_64>&, std::__cxx11::basic_string<char, 
                                                                           std::char_traits<char>, std::allocator<char> >) 
mold::elf::read_input_files<mold::elf::X86_64>(mold::elf::Context<mold::elf::X86_64>&, 
        std::span<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, 18446744073709551615ul>) 
mold::elf::elf_main<mold::elf::X86_64>(int, char**)
mold::elf::main(int, char**) 
main(int, char**)

demin-han:
Can you reproduce the problem with the minimal requirement?

rui314:
I think I understand the issue. I'm preparing a fix.

rui314:
Can you try again with the above change?

demin-han:
> Can you try again with the above change?

It's OK now.

rui314:
Cool, thank you for confirming!

PR #1229, opened 2024-04-02
Added a linker flag to favor static libraries over dynamic ones.

BrootPhorse:
Hi there!

(For my usage) This addition is primarily to support the generation of AppImages.  For an AppImage, it's better to statically link and optimize as much of the executable as possible, but not every library the executable depends upon is available as an archive (.a) file.

The change adds a flag tentatively called ```--Bstatic_preferred``` which reverses the standard behavior when searching for libraries.  When the flag is set, the ```main.cc::find_library()``` function looks for .a files first, but will return a .so if no .a is found.

* This behavior is opposite from the default which searches for .so files first, but returns .a if no .so is found.
* This behavior is different from the ```--Bstatic``` flag that ignores all .so and declares an error if the expected .a is not found.

For testing I linked my local application, and also mold itself and checked results using the dependency listing from ldd.
For initial testing I added debug to be sure that command line variables resulted in the expected library types being returned from ```main.cc::find_library()```.  This pull request version has the test debug removed.
* ```--Bstatic```  result: mold ignores .so files
* ```--Bstatic_preferred```  result: mold looks for .a then .so in each folder
* ```no argument```  result: mold looks for .so then .a in each folder

### DIFF
--- a/elf/cmdline.cc
+++ b/elf/cmdline.cc
@@ -55,6 +55,7 @@ Options:
                               Trace references to SYMBOL
   --Bdynamic, --dy            Link against shared libraries (default)
   --Bstatic, --dn, --static   Do not link against shared libraries
+  --Bstatic_preferred         Link against static if available, and shared otherwise
   --Bsymbolic                 Bind global symbols locally
   --Bsymbolic-functions       Bind global functions locally
   --Bno-symbolic              Cancel --Bsymbolic and --Bsymbolic-functions
@@ -696,9 +697,14 @@ std::vector<std::string> parse_nonpositional_args(Context<E> &ctx) {
       ctx.arg.print_map = true;
     } else if (read_flag("Bstatic") || read_flag("dn") || read_flag("static")) {
       ctx.arg.is_static = true;
+      ctx.arg.is_static_preferred = false;
       remaining.push_back("--Bstatic");
+    } else if (read_flag("Bstatic_preferred")) {
+      ctx.arg.is_static_preferred = true;
+      remaining.push_back("--Bstatic_preferred");
     } else if (read_flag("Bdynamic") || read_flag("dy")) {
       ctx.arg.is_static = false;
+      ctx.arg.is_static_preferred = false;
       remaining.push_back("--Bdynamic");
     } else if (read_flag("shared") || read_flag("Bshareable")) {
       ctx.arg.shared = true;
--- a/elf/cmdline.cc
+++ b/elf/cmdline.cc
@@ -55,7 +55,7 @@ Options:
                               Trace references to SYMBOL
   --Bdynamic, --dy            Link against shared libraries (default)
   --Bstatic, --dn, --static   Do not link against shared libraries
-  --Bstatic_preferred         Link against static if available, and shared otherwise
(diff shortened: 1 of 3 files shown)


rui314:
I'll loop in @MaskRay as this is more like a proposal to all Unix linkers than to mold. It probably should be discussed at other place like GNU ld mailing list or something like that.

MaskRay:
Thanks for looping in me. I feel nervous about the semantics. Since the application knows `-L` and `-l` options passed to the linker, I feel better it the application probes `.a`/`.so` by itself and specifies absolute/relative paths instead of `-l`. Then we don't need a new linker option at all.

BrootPhorse:
Hi Folks,

Thanks for considering.  I have a couple of points to share in defense of the concept.  Sorry for the wall of text.

1. This conversation has come up in various forums from time to time.  The suggestion from the opposition is usually to build using -l: with an explicit filename, or to switch back and forth on the command line between --Bstatic and --Bdynamic.  Those solutions technically work, but they place the burden on these types of applications to hard-code their linking and/or build environments in their makefiles.  My development team maintains over 200 applications developed over 30 years with an 8 person team.  We might be an extreme, but hard-coding the linking lines of each on those is extra effort we can't afford when we modify the build environment.  Others would have different stories, but similar problems.  Hard-coding something because a tool forces it is painful.  Avoiding that pain is why linker developers put in the ".so then .a" behavior in the first place.  Otherwise we would all hardcode the library type on the linking line.
2. Many of the default behaviors of the compiler and linker tools have a -no-\<feature\> option so that the default behavior can be turned off.  My suggestion is offered in this spirit.  All of the linkers default to .so then .a, and the only alternative is to force .a.  A developer should probably be able to select any of the 4 implied combinations: .a only, .so only, .so then .a, .a then .so.
3. "Back in the day" favoring .so libraries made sense because space and upgradability were major concerns.  These days, space is not usually a concern, so only upgradability remains.  However, in some deployments, an equal or greater concern is failed backward compatibility (DLL hell), cyber security, or qualified system continuity.  So is it really the case anymore that .so should be preferred in every application linking?

Take care,
Broot

rui314:
I actually think that this is a reasonable feature request. For whatever reason, you may want to link against static libraries as much as possible, but if some libraries are not available in the .a form, you have no choice other than linking against .so. There's no easy way to do that at this moment.

A new feature will always add maintenance burden, but in this case that would be minimal, so I'm not worried too much about it.

My main concern is that I don't want mold to be the only linker that will have this feature, as there is nothing in this feature that is specific to mold. So I want a wider consensus before proceeding.

rui314:
Honestly speaking, this pull request itself gave a slightly negative impression because the obvious inconsistency in the option name (using `_` instead of `-`) suggested that this request might not have been considered carefully. That said, the feature request itself seems fairly reasonable.

BrootPhorse:
> If you want to propose this option, you may ask on binutils AT sourceware DOT org

Okay - I'll give that a try.  Thanks.

rui314:
You don't have to prepare a patch for GNU ld. I'd just start a discussion there.

BrootPhorse:
I'm going to take this pull request back for the moment.  The interface and implementation that I discussed with the binutils folks will require changes to this.

PR #1246, opened 2024-04-30
test: use `command -v`, not non-portable `which`

thesamesam:
`which` isn't in POSIX and several Linux distributions are trying to remove it from their base system, see e.g. https://lwn.net/Articles/874049/.

Just use `command -v` which is POSIX.


rui314:
Thanks!

PR #1276, opened 2024-06-10
Fix corrupted .riscv.attributes ISA string

### LINKED ISSUE #1275: corrupted .riscv.attributes ISA string
When compiling with Android NDK 27-rc1 with riscv support, mold raises an error:

```console
C/C++:     mold: error: CMakeFiles/cmTC_fecfb.dir/testCCompiler.c.o: corrupted .riscv.attributes ISA string: rv64i2p1_m2p0_a2p1_f2p2_d2p2_c2p0_v1p0_zicsr2p0_zba1p0_zbb1p0_zbs1p0_zve32f1p0_zve32x1p0_zve64d1p0_zve64f1p0_zve64x1p0_zvl128b1p0_zvl32b1p0_zvl64b1p0
```

Version: mold 2.32.0 (compatible with GNU ld)

yujincheng08:
extension name can contain digits

Fix #1275

### DIFF
--- a/elf/arch-riscv.cc
+++ b/elf/arch-riscv.cc
@@ -1250,7 +1250,7 @@ static bool extn_version_less(const Extn &e1, const Extn &e2) {
 
 static std::optional<Extn> read_extn_string(std::string_view &str) {
   auto flags = std::regex_constants::optimize | std::regex_constants::ECMAScript;
-  static std::regex re(R"(^([a-z0-9]+)(\d+)p(\d+))", flags);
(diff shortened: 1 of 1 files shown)


rui314:
Can it start with a number? I mean I wonder if `^([a-z][a-z0-9]*)(\d+)p(\d+)` is correct.

yujincheng08:
@rui314 I dont think it can. Your solution looks better.

rui314:
And it appears it cannot end with a number, as such digit cannot be distinguished with a following major version number. So I guess it should be `^([a-z]|[a-z][a-z0-9]*[a-z])(\d+)p(\d+)`

PR #1277, opened 2024-06-13
Update README adding Conan usage instructions

perseoGI:
Hi @rui314, thanks for your amazing work with `mold`!
I've been using it personally for a while now.

As a member of Conan team I've recently reviewed your projects recipe on Conan Center Index and came with the idea of adding some explanation on how to use `conan` to make `mold` the linker.

It is pretty simple as you can see and avoids users to directly download `mold` by themselves. It will all be managed by `conan`.

If you have any other doubt or think it may be needed more explanation, feel free to edit the PR or leave a comment!
Thank you again and happy coding!

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,26 @@ GNU linker, especially when linking large programs.
 
 </details>
 
+<details><summary>If you are using Conan package manager</summary>
+
+You can configure [Conan](https://github.com/conan-io) to download the latest
+version of `mold` and use it as the linker when building your dependencies and
+projects from source.
+
+Add the following section to your _host_ profile targetting Linux:
+
+```
+[tool_requires]
+*:mold/[*]
+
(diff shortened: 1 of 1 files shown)


rui314:
It looks like it is more manageable for both of us to write the instructions on your site (maybe https://conan.io/center/recipes/mold, but I'm not familiar with Conan) and add just a link in the README. What do you think?

perseoGI:
Hi @rui314, thank you for your suggestion. We have implemented a new tab in our web https://conan.io/center/recipes/mold?version=2.33.0
which will contain a README (if added) for detailed information on how to use the packages.
It is still not deployed to production but it will be soon!
Thank you again for your incredible work with mold, I personally use it in my system!

rui314:
Thanks!

PR #1281, opened 2024-06-16
Use $CXX to detect cxxflags

ziyao233:
Use $CXX instead of $CC to detect whether C++ programs could be built statically. Add function test_cxxflags for convenience.

Fix test error in an environment where C++ programs cannot be built statically, with error like:

  mold: fatal: library not found: c++


rui314:
Thanks!

PR #1283, opened 2024-06-19
Fix `get_self_path` on windows

yujincheng08:
On windows, calling `read_symlink` will throw an exception.

### DIFF
--- a/common/filepath.cc
+++ b/common/filepath.cc
@@ -60,6 +60,13 @@ std::string get_self_path() {
   path.resize(size);
   sysctl(mib, 4, path.data(), &size, NULL, 0);
   return path;
+#elif _WIN32
+  char path[32768];
+  if (!GetModuleFileName(nullptr, path, sizeof(path))) {
+    std::cerr << "GetModuleFileName failed\n";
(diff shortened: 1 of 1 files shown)


rui314:
We use this fiction for `mold -run`, but the first of not supported on Windows, so we don't need the code for it.

yujincheng08:
well, i am trying to make mold support lto on msys: https://github.com/yujincheng08/MINGW-packages/commit/8ff0e682a1596db1fbb3df638dbcc49fdfaf7450 and this is one of the patch so I wonder if it can be somehow merged by upstream.

rui314:
Then can you send me PR once the feature is ready?

rui314:
So, are you trying to cross build Linux binaries on msys on Windows?

yujincheng08:
@rui314 sure I am more than happy to PR this. And yes I am cross-compiling Linux binaries on msys on windows.

yujincheng08:
You are right. This is because the `is_llvm` check has failed so lto tries to `restart_process` which involves this. So this PR is not needed.

PR #1285, opened 2024-06-19
Only define `STDERR_FILENO` when not defined

yujincheng08:


rui314:
What is this fixing?

yujincheng08:
when trying to compile mold on msys, there's a warning of redefining STDERR_FILENO.

PR #1287, opened 2024-06-21
Support lto on mingw

yujincheng08:


rui314:
Some buildbots are broken. Please fix.

common/signal-win32.cc
```diff
@@ -38,6 +38,10 @@ static LONG WINAPI vectored_handler(_EXCEPTION_POINTERS *exception_info) {
     static const char msg[] =
       "mold: failed to write to an output file. Disk full?\n";
     (void)!_write(_fileno(stderr), msg, sizeof(msg) - 1);
+  } else if (rec->ExceptionCode == EXCEPTION_STACK_OVERFLOW) {
+    static const char msg[] =
+      "mold: stackoverflow.\n";
+    (void)!_write(_fileno(stderr), msg, sizeof(msg) - 1);
```
rui314, line 44: Unrelated change?

yujincheng08: Yes. I encountered this problem during my testing and it makes me confused that `mold` exits with 1 without any message.

yujincheng08:
@rui314 CI fixed

CMakeLists.txt
```diff
@@ -56,6 +56,10 @@ if(MOLD_USE_MOLD)
   add_link_options(-fuse-ld=mold -Wl,--gdb-index)
 endif()
 
+if(MINGW)
+  set(CMAKE_DL_LIBS -ldl)
```
rui314, line 60: If this is equivalent to `target_link_libraries(mold PRIVATE dl)`, please do so instead of overriding the system-provided variable.

yujincheng08: Done

PR #1304, opened 2024-07-14
Align Chuck, InputSection, and SectionFragment to at least 4 bytes

chewi:
There are already static_asserts to check this, so it is almost always true anyway, but not on m68k, which aligns to 2 bytes by default.

Perhaps the static_asserts are now redundant? I can remove them if wanted.

The tests pass on m68k.

### DIFF
--- a/elf/mold.h
+++ b/elf/mold.h
@@ -60,7 +60,7 @@ std::string get_mold_version();
 //
 
 template <typename E>
-struct SectionFragment {
+struct alignas(4) SectionFragment {
   SectionFragment(MergedSection<E> *sec, bool is_alive)
     : output_section(*sec), is_alive(is_alive) {}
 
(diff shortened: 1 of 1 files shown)


chewi:
Hmmm, GCC's `__attribute__((alignment))` sets the minimum alignment, but C++'s `alignas` sets it absolutely. I believe Clang support's GCC's method, but would you be okay with that?

rui314:
We need to make it compile with clang, so I think we need to use the good old C macro. I'll merge this and fix the build breakage in a follow up patch.

I'm honestly very surprised that all tests pass on m68k only with this change. It seems Linux/m68k has been maintained very well!

rui314:
Does this work for you?

chewi:
I was still in the middle of trying `__attribute__((aligned(4)))` with Clang. I wasn't able to try it for m68k, but it does otherwise work and should do so for all platforms, so you might want to apply that instead.

chewi:
> I'm honestly very surprised that all tests pass on m68k only with this change. It seems Linux/m68k has been maintained very well!

Actually, I had tested with qemu-user, but only m68k-range-extension-thunk fails under qemu-system. It says to skip 32-bit arches (specifically i686 and riscv32) at the top of range-extension-thunk.sh though, so you could add m68k here.

PR #1308, opened 2024-07-17
Add --encoded-package-metadata

bdrung:
Specifying the compiler flag `-Wl,--package-metadata=<JSON>` might not work, because the shells might eat the quotation marks and the compiler splits the JSON at the commas.

Ubuntu tried to using a specs file to set `--package-metadata` but that turned out to be too fragile. autopkgtests might use the compiler flags but the needed environment variables are not set in the test environment. Debugging a crash of an application build with the -spec parameter lacks the environment variables. People like to iteratively continue building the software in the build directory while hacking on the package and then have no environment variable set.

So introduce a `--encoded-package-metadata` linker flag that takes a percent-encoded JSON. Percent-encoding is used because it is a standard and simple to implement.

Bug-Ubutru: https://bugs.launchpad.net/bugs/2071468

### DIFF
--- a/elf/cmdline.cc
+++ b/elf/cmdline.cc
@@ -122,6 +122,8 @@ Options:
   --oformat=binary            Omit ELF, section, and program headers
   --pack-dyn-relocs=[relr,none]
                               Pack dynamic relocations
+  --encoded-package-metadata=PERCENT_ENCODED_STRING
+                              Set a given string to .note.package
   --package-metadata=STRING   Set a given string to .note.package
   --perf                      Print performance statistics
   --pie, --pic-executable     Create a position-independent executable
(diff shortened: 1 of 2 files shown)


rui314:
Ah, it's so nice to see yet another example of how hard it is to escape characters properly. I'm fine with this change, as it's as simple as it is, and it's not going to be a maintenance burden. But I wonder if you are going to add the same flag to other linkers.

bdrung:
Yes, I will add this flag to the other compilers.
* binutils ld: https://sourceware.org/pipermail/binutils/2024-July/135733.html
* gold: https://sourceware.org/pipermail/binutils/2024-July/135734.html
* lld: TBD

rui314:
So it looks like the discussion to add the same flag to the GNU linker is still ongoing, and it's not entirely clear whether the GNU linker will receive the flag in this form* I'll remove the flag from our documentation, making it effectively a hidden option, to avoid making it an "official" feature. Let me know when the feature is added to the GNU ld.

By the way, one thing I'm wondering is *why* the issue with `--package-metadata` wasn't noticed before adding the flag to GNU ld, mold, and lld? In hindsight, it is an obvious problem to pass a raw JSON string as a command line parameter while `,` is not escapable with `-Wl,`. Was that due to the lack of testing?

I'm asking because if we add another flag without testing it, it may also have an unnoticed issue again. For example, is `%` a safe escape character everywhere? Maybe we should use MIME's Q-encoding because `=` is a safer choice than `%`? Or BASE64? Or should we pass a JSON string as a file instead of a command line argument? All in all, is there any chance that we might need one more flag to work around an unknown `--encoded-package-metadata` issue?

bdrung:
> So it looks like the discussion to add the same flag to the GNU linker is still ongoing, and it's not entirely clear whether the GNU linker will receive the flag in this form I'll remove the flag from our documentation, making it effectively a hidden option, to avoid making it an "official" feature. Let me know when the feature is added to the GNU ld.

Thanks for hiding the option for now. Let's see which direction the discussion goes. My current suggestions: https://sourceware.org/bugzilla/show_bug.cgi?id=32003#c27

> By the way, one thing I'm wondering is why the issue with `--package-metadata` wasn't noticed before adding the flag to GNU ld, mold, and lld? In hindsight, it is an obvious problem to pass a raw JSON string as a command line parameter while `,` is not escapable with `-Wl,`. Was that due to the lack of testing?

No idea why that wasn't noticed before. We noticed it quite quickly while trying to enable ELF package metadata in the Ubuntu archive. 

> I'm asking because if we add another flag without testing it, it may also have an unnoticed issue again. For example, is `%` a safe escape character everywhere?

Shell and Makefiles have no problems with it. % is used for escaping in other cases as well (e. g. escaping version numbers in git tags).

> Maybe we should use MIME's Q-encoding because `=` is a safer choice than `%`? Or BASE64?

Both mentioned in https://sourceware.org/bugzilla/show_bug.cgi?id=32003#c27 and decided against.

> Or should we pass a JSON string as a file instead of a command line argument?

That would be fragile as well, because such file would be needed to be generated dynamically and could be gone later.

rui314:
I don't have a strong opinion on it. This option is not for ordinary users. As long as it works for you guys, it is fine for me. I honestly thought that this is not a big deal and not that controversial.

One thing to note though is that the Q encoding can encode anything, just like the percent encoding, right? By default, your encoder may not encode `,`, but it is legal to encode it as `=2C`.

bdrung:
> One thing to note though is that the Q encoding can encode anything, just like the percent encoding, right? By default, your encoder may not encode `,`, but it is legal to encode it as `=2C`.

Yes, that would work.

```
$ ipython3
In [1]: import quopri

In [2]: quopri.decodestring(b"foo=2Cbar")
Out[2]: b'foo,bar'

In [3]: quopri.encodestring(b"foo,bar")
Out[3]: b'foo,bar'
```

PR #1316, opened 2024-07-25
[LoongArch] adjust the range of R_LARCH_B26 relocation

ywgrit:
Hi, Rui.
Thanks for enhancing R_LARCH_B26 relocation of Loongarch.
The immediate number field of the bl/b instruction occupies 26 bits, and the jump target value is the immediate number followed by two bit 0s, so I think the range of value-pc should be (0xf8000000~0x07fffffc).

### DIFF
--- a/elf/arch-loongarch.cc
+++ b/elf/arch-loongarch.cc
@@ -305,7 +305,7 @@ void InputSection<E>::apply_reloc_alloc(Context<E> &ctx, u8 *base) {
       break;
     case R_LARCH_B26: {
       i64 val = S + A - P;
-      if (val < -(1 << 27) || (1 << 27) <= val)
+      if (val < (i32)0xf8000000 || val > (i32)0x07fffffc)
         val = get_thunk_addr(i) + A - P;
       write_d10k16(loc, val >> 2);
       break;


rui314:
I think your commit doesn't change the behavior of the program at all because these values are the same. See this: https://godbolt.org/z/Wvf1Wrch3

rui314:
I took a quick look at your fork of this repo (https://github.com/ywgrit/mold/blob/main/elf/arch-loongarch.cc), and it looks like you are making a lot of changes to this file. I don't think your file is ready for code review, but I have to say that you are doing too many things at once, also in wrong ways. You will waste of your time if you do not fix some fundamental issues first.

I'd recommend you do this incrementally. You probably want to implement just one relaxation, perhaps R_LARCH_RELAX, and submit it as a pull request to see if your way of doing it aligns with the way that we do in mold. Then you can gradually implement one or a few relaxations at a time.

ywgrit:
> I think your commit doesn't change the behavior of the program at all because these values are the same. See this: https://godbolt.org/z/Wvf1Wrch3

Sorry, I didn't double check before launching pr

ywgrit:
> I took a quick look at your fork of this repo (https://github.com/ywgrit/mold/blob/main/elf/arch-loongarch.cc), and it looks like you are making a lot of changes to this file. I don't think your file is ready for code review, but I have to say that you are doing too many things at once, also in wrong ways. You will waste of your time if you do not fix some fundamental issues first.
> 
> I'd recommend you do this incrementally. You probably want to implement just one relaxation, perhaps R_LARCH_RELAX, and submit it as a pull request to see if your way of doing it aligns with the way that we do in mold. Then you can gradually implement one or a few relaxations at a time.

Thanks for the advice.
The testing isn't done yet, the code is just stacked together, and we'll split this change once the local testing is passed.

rui314:
It looks like there are a few fundamental issues in your code, and some parts of them are likely to be rewritten completely. So don't spend too much time on them and instead get an early feedback from me by sending the first one.

ywgrit:
> It looks like there are a few fundamental issues in your code, and some parts of it are likely to need to be rewritten from the ground up. So don't spend too much time on it, and instead get early feedback from me by sending the first pull request. I don't want to ask you to rewrite them completely after you think you've finished them, but that seems like the likely trajectory at this point.

Okay, I will send first pull request soon.

PR #1318, opened 2024-07-30
[LoongArch] Support relax of pcalau12i/addi.d->pcaddi

ywgrit:
Similar to riscv, doing relaxation on LoongArch for some instruction pairs can reduce the number of instructions, a few cases have been supported now.

### DIFF
--- a/elf/arch-loongarch.cc
+++ b/elf/arch-loongarch.cc
@@ -336,8 +336,6 @@ void InputSection<E>::apply_reloc_alloc(Context<E> &ctx, u8 *base) {
       write_k12(loc, (S + A) >> 52);
       break;
     case R_LARCH_PCALA_LO12:
-      if (removed_bytes == 4)
-        break;
       // It looks like R_LARCH_PCALA_LO12 is sometimes used for JIRL even
       // though the instruction takes a 16 bit immediate rather than 12 bits.
       // It is contrary to the psABI document, but GNU ld has special
@@ -348,15 +346,13 @@ void InputSection<E>::apply_reloc_alloc(Context<E> &ctx, u8 *base) {
         write_k12(loc, S + A);
       break;
     case R_LARCH_PCALA_HI20:
-      if (i + 4 <= rels.size()
-          && rels[i+2].r_type == R_LARCH_PCALA_LO12
-          && rels[i+1].r_type == R_LARCH_RELAX
-          && rels[i+3].r_type == R_LARCH_RELAX
-          && get_r_delta(i+3) - get_r_delta(i+2) == 4) {
-        // relax pcalau12i/addi.d to pcaddi
+      if (removed_bytes == 4) {
+	// pcalau12i/addi.d has been relaxed to pcaddi
         i64 rd = bits(*(ul32 *)(contents.data() + rel.r_offset), 4, 0);
         *(ul32 *)loc = rd | 0x18000000;
         write_j20(loc, (S + A - P) >> 2);
+	// skip the next R_LARCH_RELAX, R_LARCH_PCALA_LO12, R_LARCH_RELAX
+	i += 3;
       } else {
         write_j20(loc, hi20(S + A, P));
       }
@@ -769,7 +765,7 @@ void shrink_section(Context<E> &ctx, InputSection<E> &isec, bool use_rvc) {
         delta += 4;
       break;
     case R_LARCH_PCALA_HI20:
-      if ((i+4) > rels.size())
+      if (i + 4 > rels.size())
         continue;
 
       ul32 pcala = *(ul32 *)(isec.contents.data() + rels[i].r_offset);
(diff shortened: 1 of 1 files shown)


ywgrit:
@Jinyang He

ywgrit:
@MQ-menging

rui314:
Thank you for doing this. It's looking much better.

elf/arch-loongarch.cc
```diff
@@ -346,7 +348,18 @@ void InputSection<E>::apply_reloc_alloc(Context<E> &ctx, u8 *base) {
         write_k12(loc, S + A);
       break;
     case R_LARCH_PCALA_HI20:
-      write_j20(loc, hi20(S + A, P));
+      if (i + 4 <= rels.size()
```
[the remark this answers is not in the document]
ywgrit: When we traverse to the first reloc of (pcalau12i/addi.d) pair, i.e., R_LARCH_PCALA_HI20, the removed_bytes are 0. So we need more information to check if we have applied relaxation on this two instructions before.
The following is a more detailed explanation
                                                       reloc                             delta
current reloc ->          R_LARCH_PCALA_HI20         n                     removed_bytes == 0
                                          R_LARCH_RELAX                     n                     removed_bytes == 0
                                          R_LARCH_LO12                        n                     removed_bytes == 4
                                          R_LARCH_RELAX                     n+4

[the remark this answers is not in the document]
ywgrit: > You should be able to check only `removed_bytes` to see if you need to rewrite the instruction.

When we traverse to the first reloc of (pcalau12i/addi.d) pair, i.e., R_LARCH_PCALA_HI20, the removed_bytes are 0. So we need more information to check if we have applied relaxation on this two instructions before.
The following is a more detailed explanation
```
                    reloc                           delta
current reloc ->    R_LARCH_PCALA_HI20               n                     removed_bytes == 0
                    R_LARCH_RELAX                    n                     removed_bytes == 0
                    R_LARCH_PCALA_LO12               n                     removed_bytes == 4
                    R_LARCH_RELAX                    n+4
```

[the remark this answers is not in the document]
rui314: If you take a look at arch-riscv.cc, you would notice that we remove an instruction pointed to by a HI20 relocation and keep a LO12. You are trying to keep HI20 and remove LO12. Try to follow what we do in arch-riscv64.cc.

elf/arch-loongarch.cc
```diff
@@ -346,7 +346,16 @@ void InputSection<E>::apply_reloc_alloc(Context<E> &ctx, u8 *base) {
...
       break;
     case R_LARCH_PCALA_HI20:
-      write_j20(loc, hi20(S + A, P));
+      if (removed_bytes == 4) {
+	// pcalau12i/addi.d has been relaxed to pcaddi
+        i64 rd = bits(*(ul32 *)(contents.data() + rel.r_offset), 4, 0);
+        *(ul32 *)loc = rd | 0x18000000;
+        write_j20(loc, (S + A - P) >> 2);
+	// skip the next R_LARCH_RELAX, R_LARCH_PCALA_LO12, R_LARCH_RELAX
+	i += 3;
```
rui314, line 355: Do not rewrite the instruction if it is removed. Instead, do this in LO12. See how we handle R_RISCV_TPREL_LO12_I in arch-riscv.cc, for example.

ywgrit: > Do not rewrite the instruction if it is removed. Instead, do this in LO12. See how we handle R_RISCV_TPREL_LO12_I in arch-riscv.cc, for example.

Sorry for this oversight.

rui314:
It looks like it's more complicated than I thought because of the improper use of `R_LARCH_PCALA_LO12` for the `JIRL` instruction, so let me merge this as-is and fix the issues in a follow-up change. It should be much faster than back-and-forth code reviews.

ywgrit:
Is there anything else I should do?

elf/arch-loongarch.cc
```diff
@@ -755,6 +762,31 @@ void shrink_section(Context<E> &ctx, InputSection<E> &isec, bool use_rvc) {
...
+
+      i64 rd = pcala & 0x1f;
+      const ul32 addi_d = 0x02c00000;
+      const u64 loc = isec.get_addr() + r.r_offset - delta;
+
+      u64 symval = sym.get_addr(ctx) + r.r_addend;
+      /* Is pcalau12i + addi.d insns?  */
+      if (rels[i+2].r_type != R_LARCH_PCALA_LO12
+          || rels[i+3].r_type != R_LARCH_RELAX
+          || (addi & addi_d) != addi_d
```
rui314, line 780: By the way, this line is wrong as a test if the instruction is `addi` right? I took a look at the LoongArch instruction format, and it says that the opcode always starts at the most significant bit. So it looks like it should be `addi & 0xffc0'0000) != addi`.

ywgrit: > By the way, this line is wrong as a test if the instruction is `addi` right? I took a look at the LoongArch instruction format, and it says that the opcode always starts at the most significant bit. So it looks like it should be `(addi & 0xffc0'0000) != addi`.

You are right, this is a mistake.

PR #1322, opened 2024-08-03
LoongArch: support relaxation of pcalau12i/ld.d to pcalau12i/addi.d or pcaddi

ywgrit:
If we get symval by accessing to local got, then we can omit one memory loading.
The first phase: pcalau12i/ld.d to pcala12i/addi.d
The second phase: pcalau12i/addi.d to pcaddi

### DIFF
--- a/elf/arch-loongarch.cc
+++ b/elf/arch-loongarch.cc
@@ -377,11 +377,12 @@ void InputSection<E>::apply_reloc_alloc(Context<E> &ctx, u8 *base) {
           // loc stores 'ld.d', rewrite ld.d with pcaddi
           *(ul32 *)(loc) = 0x1800'0000 | get_rd(*(ul32 *)loc);
           write_j20(loc, (S + A - P) >> 2);
+          i += 3;
           break;
         case 0:
-          if (i + 3 < rels.size() &&
-            sym.is_pcrel_linktime_const(ctx);
-            ctx.arg.relax &&
+          if (ctx.arg.relax &&
+            sym.is_pcrel_linktime_const(ctx) &&
+            i + 3 < rels.size() &&
             rels[i + 1].r_type == R_LARCH_RELAX &&
             rels[i + 3].r_type == R_LARCH_RELAX &&
             rels[i + 2].r_type == R_LARCH_GOT_PC_LO12 &&
@@ -409,6 +410,7 @@ void InputSection<E>::apply_reloc_alloc(Context<E> &ctx, u8 *base) {
         default:
           unreachable();
       }
+      break;
     case R_LARCH_GOT64_PC_LO20:
       write_j20(loc, higher20(GOT + G + A, P));
       break;
@@ -870,7 +872,7 @@ void shrink_section(Context<E> &ctx, InputSection<E> &isec, bool use_rvc) {
       //   pcalau12i $t0, 0
       //   addi.d    $t0, $t0, 0
       if (ctx.arg.relax &&
-          sym.is_pcrel_linktime_const(ctx);
+          sym.is_pcrel_linktime_const(ctx) &&
           i + 3 < rels.size() &&
           rels[i + 2].r_type == R_LARCH_GOT_PC_LO12 &&
           rels[i + 2].r_offset == rels[i].r_offset + 4 &&
--- a/elf/arch-loongarch.cc
+++ b/elf/arch-loongarch.cc
@@ -118,6 +118,10 @@ static u32 get_rd(u32 insn) {
   return insn & 0x1f;
 }
 
+static u32 get_rj(u32 insn) {
+  return (insn >> 5) & 0x1f;
+}
+
 static void set_rj(u8 *loc, u32 rj) {
   assert(rj < 32);
   *(ul32 *)loc &= 0b111111'1111111111111111'00000'11111;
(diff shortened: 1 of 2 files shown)


ywgrit:
@MQ-mengqing

elf/arch-loongarch.cc
```diff
@@ -816,6 +878,39 @@ void shrink_section(Context<E> &ctx, InputSection<E> &isec, bool use_rvc) {
...
+          ctx.arg.relax &&
+          i + 3 < rels.size() &&
+          rels[i + 2].r_type == R_LARCH_GOT_PC_LO12 &&
+          rels[i + 2].r_offset == rels[i].r_offset + 4 &&
+          rels[i + 3].r_type == R_LARCH_RELAX) {
+          u32 insn1 = *(ul32 *)(isec.contents.data() + rels[i].r_offset);
+          u32 insn2 = *(ul32 *)(isec.contents.data() + rels[i].r_offset + 4);
+
+          // relax pcalau12i/ld.d to pcalau12i/addi.d
+          if (get_rd(insn1) != get_rd(insn2))
```
MQ-mengqing, line 905: For more reliable, check rj is also necessary.

elf/arch-loongarch.cc
```diff
@@ -816,6 +878,39 @@ void shrink_section(Context<E> &ctx, InputSection<E> &isec, bool use_rvc) {
...
+            continue;
+
+          i64 dist = compute_distance(ctx, sym, isec, r);
+          // the second phase: relax pcalau12i/addi.d to pcaddi
+          if (dist % 4 == 0 && -(1 << 21) < dist && dist < (1 << 21))
```
MQ-mengqing, line 910: There should be an eage case "-(1<<21) == dist".

elf/arch-loongarch.cc
```diff
@@ -364,11 +364,73 @@ void InputSection<E>::apply_reloc_alloc(Context<E> &ctx, u8 *base) {
...
+            !sym.is_absolute() &&
+            !sym.is_ifunc() &&
+            ctx.arg.relax &&
+            rels[i+1].r_type == R_LARCH_RELAX &&
+            rels[i+3].r_type == R_LARCH_RELAX &&
+            rels[i+2].r_type == R_LARCH_GOT_PC_LO12 &&
+            rels[i+2].r_offset == rel.r_offset + 4) {
+            u32 insn1 = *(ul32 *)(contents.data() + rel.r_offset);
+            u32 insn2 = *(ul32 *)(contents.data() + rels[i+2].r_offset);
+            u32 rd = get_rd(insn1);
```
MQ-mengqing, line 418: I think there should be a range check for pcalau12i.

elf/arch-loongarch.cc
```diff
@@ -364,11 +364,73 @@ void InputSection<E>::apply_reloc_alloc(Context<E> &ctx, u8 *base) {
...
+
+            if (rd == get_rd(insn2)) {
+              // pcalau12i/ld.d has been relaxed to pcalau12i/addi.d
+              // reloc the pcalau12i as R_LARCH_PLACA_HI20
+              write_j20(loc, hi20(S + A, P));
+              break;
```
rui314, line 424: You should directly write to `loc + 4` and then `i += 3` here, instead of doing the same check in `case R_LARCH_GOT_PC_LO12`.

elf/arch-loongarch.cc
```diff
@@ -364,11 +364,73 @@ void InputSection<E>::apply_reloc_alloc(Context<E> &ctx, u8 *base) {
...
-      break;
+      switch (removed_bytes) {
+        // pcalau12i/ld.d has been relaxed to pcaddi, the first insn has been removed.
+        case 4:
+          break;
+        case 0:
+          if (i + 3 < rels.size() &&
+            sym.is_local(ctx) &&
+            !sym.is_absolute() &&
+            !sym.is_ifunc() &&
```
rui314, line 410: It looks like you want to do the same as we do for `R_RISCV_GOT_HI20`. Did you check that in arch-riscv.cc? We have `sym.is_pcrel_linktime_const(ctx)` to check if the symbol's PC-relative address is link-time constant.

elf/arch-loongarch.cc
```diff
@@ -364,11 +364,73 @@ void InputSection<E>::apply_reloc_alloc(Context<E> &ctx, u8 *base) {
...
+            ctx.arg.relax &&
+            rels[i+1].r_type == R_LARCH_RELAX &&
```
rui314, line 412: You want to pay more attention to the details -- add spaces around `+`. This is how we format it in other places.

elf/arch-loongarch.cc
```diff
@@ -364,11 +364,73 @@ void InputSection<E>::apply_reloc_alloc(Context<E> &ctx, u8 *base) {
...
+            rels[i+3].r_type == R_LARCH_RELAX &&
+            rels[i+2].r_type == R_LARCH_GOT_PC_LO12 &&
+            rels[i+2].r_offset == rel.r_offset + 4) {
+            u32 insn1 = *(ul32 *)(contents.data() + rel.r_offset);
```
rui314, line 416: and also indentation.

test/elf/loongarch64_relax-got.sh
```diff
@@ -0,0 +1,64 @@
...
+  ld.w $a0, $a0, 0
+  ret
+get_sym4:
+  la.global $a0, sym4
+  ld.w $a0, $a0, 0
+  ret
+get_sym5:
+  la.global $a0, sym5
+  ld.w $a0, $a0, 0
+  ret
```
rui314, line 25: I don't think you need five symbols for this test, as they are testing exactly the same relaxation.

elf/arch-loongarch.cc
```diff
@@ -364,11 +364,73 @@ void InputSection<E>::apply_reloc_alloc(Context<E> &ctx, u8 *base) {
...
+
+          if (rd == get_rd(insn2)) {
+            // pcalau12i/ld.d has been relaxed to pcalau12i/addi.d
+            // rewrite the ld.d with addi.d
+            *(ul32 *)loc = 0x02c00000 | rd | (rd << 5);
+            write_k12(loc, S + A);
+            break;
+          }
+        }
+
```
rui314, line 396: I think you can revert this change.

ywgrit:
Do not care about the last two commits now, there are still some issues, I'll commit the new code once I've tested it out.

rui314:
I'll fix a few minor issues after merging this change. Next time, please pay more attention to indentation. Thank you!

ywgrit:
> I'll fix a few minor issues after merging this change. Next time, please pay more attention to indentation. Thank you!

Yeah, Sorry for that.

ywgrit:
> Thank you for doing this anyway. Did you have anything other than this, or is this the last one?


TLS LD/GD relaxation is the next plan.

rui314:
I actually don't think you can relax TLS GD/LD on LoongArch, because calls to `__tls_get_addr` is not annotated by any relocation, so you cannot eliminate that function call. You might still be able to relax a R_LARCH_TLS_GD_PC_HI20 and R_LARCH_GOT_PC_LO12 pair into a single instruction, but that optimization is marginal and ineffective. `__tls_get_addr` is expensive anyways, so I wouldn't try to do that relaxation. What were you planning to do?

ywgrit:
> I actually don't think you can relax TLS GD/LD on LoongArch, because calls to `__tls_get_addr` are not annotated by any relocation, so you cannot eliminate that function call. You might still be able to relax a R_LARCH_TLS_GD_PC_HI20 and R_LARCH_GOT_PC_LO12 pair into a single instruction, but that optimization is marginal and probably meaningless. `__tls_get_addr` is expensive anyways, so I wouldn't try to do that relaxation. What were you planning to do?

You're right, actually tls_desc is currently supported by loongarch, though it hasn't been introduced on mold yet. Considering the negligible performance gain from saving an instruction, it seems there is no need to apply relaxation on tls.

rui314:
Then maybe you want to add TLSDESC support?

ywgrit:
> Then maybe you want to add TLSDESC support?

Yeah, I want have a try.

PR #1333, opened 2024-08-21
LoongArch: apply relaxation to tls desc

ywgrit:
The optimaztion on tls access model could be divided into two parts.
1) Trandform TLS desc to Initial-Exec/Local-Exec. When transform from TLS desc
to Local-Exec, we could shrink the instruction sequences to one instruction(ori) to get the TP-relative address if it is small.
2) Relax the instruction sequences. If we could not transform TLS access model, we may try relax the instruction sequences.


ywgrit:
@MQ-mengqing

PR #1335, opened 2024-08-26
LoongArch: fix error offset for relaxation of R_LARCH_CALL36

ywgrit:
LoongArch: fix error offset for relaxation of R_LARCH_CALL36


rui314:
Can you add a test?

ywgrit:
> Can you add a test?

Yeah, I changed the branch offset to check this in the second commit.

rui314:
Your test didn't fail even without this change. I'll fix myself, but next time, please verify that not only your test passes with the new change but also it fails without it.

PR #1337, opened 2024-08-29
alloca.h is missing on NetBSD

leleliu008:
alloca(3) is defined in stdlib.h on NetBSD

Reference: https://man.netbsd.org/alloca.3


rui314:
Out of curiosity, does mold work on NetBSD?

leleliu008:
It should works on NetBSD. It has already included in Official Repository https://github.com/NetBSD/pkgsrc/tree/trunk/devel/mold

Issue #851, opened 2022-11-07
Error: mold-1.6.0 - Too many open files 


Cerberios:
Hello there!
I got error while compile MESA driver with mold.
Environment - Gentoo
---
mold: fatal: cannot open src/gallium/drivers/nouveau/libnouveau.a.p/nvc0_nvc0_video_vp.c.o: Too many open files
clang-15: error: linker command failed with exit code 1 (use -v to see invocation)
---
Please, instruct me what kind of information need provide additionally in order to help devs fix it.
Thanks!

rui314:
Are you running it on a VM, in a Docker, or on a special filesystem?

Cerberios:
I use Laptop PC with real HDD with disk model: ST1000LM048-2E71
Separated partition for /root with LUKS encryption and EXT4 filesystem.
clang version 15.0.4
I tried to switch from stable 1.6.0 version to git-based version of mold - the same error.
mold 1.6.0 (bd555b33abad4abe23e755a2c1e0f773672c997f; compatible with GNU ld)
---My flags for compiler---
SPECFLAGS="-O3 -march=native -mtune=native -pipe -fopenmp
-mllvm -polly
-mllvm -polly-parallel
-mllvm -polly-vectorizer=polly
-mllvm -polly-omp-backend=LLVM
-mllvm -polly-num-threads=4
-mllvm -polly-scheduling=dynamic
-mllvm -polly-scheduling-chunksize=1
-mllvm -polly-ast-use-context
-mllvm -polly-invariant-load-hoisting
-mllvm -polly-loopfusion-greedy
-mllvm -polly-run-inliner
-mllvm -polly-run-dce
-fno-math-errno -fno-trapping-math -falign-functions=32 -fno-semantic-interposition -flto=thin"

CFLAGS="${SPECFLAGS}"
CXXFLAGS="${SPECFLAGS}"

LDFLAGS="${LDFLAGS} -Wl,--hash-style=gnu -Wl,-O1,--sort-common,-z,relro,-z,now -Wl,--enable-new-dtags -Wl,-O3,-Bsymbolic-functions,--as-needed -Wl,--thinlto-jobs=all -lomp -fuse-ld=mold"
---

rui314:
Please run `ulimit -a` and paste its output.

Cerberios:
**_cerberos /tmp # ulimit -a_**
real-time non-blocking time  (microseconds, -R) unlimited
core file size              (blocks, -c) 0
[...]

PR #1363, opened 2024-10-21
increase file limit when hit the ulimit

### LINKED ISSUE #1362: Potenial fd leak with clang lto and cause too many open files
My Environment
====
```
$ ld --version     
mold 2.34.1 (compatible with GNU ld)

$ clang++ --version                                                                                                                   
clang version 20.0.0pre20241012.gb1746894deebe3 (Fedora 20.0.0~pre20241012.gb1746894deebe3-2.fc41)
Target: x86_64-redhat-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
Configuration file: /etc/clang/x86_64-redhat-linux-gnu-clang++.cfg

$ ulimit -a 
-t: cpu time (seconds)              unlimited
-f: file size (blocks)              unlimited
-d: data seg size (kbytes)          unlimited
-s: stack size (kbytes)             8192
-c: core file size (blocks)         unlimited
-m: resident set size (kbytes)      unlimited
-u: processes                       511419
-n: file descriptors                1024
-l: locked-in-memory size (kbytes)  8192
-v: address space (kbytes)          unlimited
-x: file locks                      unlimited
-i: pending signals                 511419
-q: bytes in POSIX msg queues       819200
-e: max nice                        0
-r: max rt priority                 0
-N 15: rt cpu time (microseconds)   unlimited
```
***

- Project being built: https://github.com/Geant4/geant4/releases/tag/v11.2.2
- Error message: 
  ```
  /usr/bin/clang++ -fPIC -W -Wall -pedantic -Wno-non-virtual-dtor -Wno-long-long -Wwrite-strings -Wpointer-arith -Woverloaded-virtual -Wno-variadic-macros -Wshadow -pipe -Qunused-arguments -DGL_SILENCE_DEPRECATION -pthread  -O2 -g -DNDEBUG -flto=thin   -shared -Wl,-soname,libG4processes.so -o BuildProducts/lib64/libG4processes.so @CMakeFiles/G4processes.rsp 
  mold: fatal: opening source/CMakeFiles/G4processes.dir/processes/hadronic/models/im_r_matrix/src/G4CollisionNNToDeltaDelta.cc.o failed: Too many open files
  clang++: error: linker command failed with exit code 1 (use -v to see invocation)
  ```
- related input to clang (it is one object file per line)
  ```
  $ cat CMakeFiles/G4processes.rsp |wc -l              
  1710
  ```

***
Only reproducable with:
 - LTO enabled
 - mold is used
So not reporting to llvm project as lld or bfd can do the link with same ulimit.
 - Can be workarounded by raising `ulimit  -n`
***
`strace -f`:  (part of, [full strace result (gzipped)](https://github.com/user-attachments/files/17460184/strace-err.out.gz))
```
[pid 320263] openat(AT_FDCWD, "source/CMakeFiles/G4processes.dir/processes/hadronic/models/im_r_matrix/src/G4CollisionMesonBaryon.cc.o", O_RDONLY) = 1014
[pid 320263] openat(AT_FDCWD, "source/CMakeFiles/G4processes.dir/processes/hadronic/models/im_r_matrix/src/G4XNDeltaTable.cc.o", O_RDONLY) = 1015
[pid 320263] fstat(1015, {st_mode=S_IFREG|0644, st_size=54628, ...}) = 0
[pid 320263] mmap(NULL, 54628, PROT_READ|PROT_WRITE, MAP_PRIVATE, 1015, 0) = 0x7f08aec9b000
[pid 320263] close(1015)                = 0
[pid 320263] openat(AT_FDCWD, "source/CMakeFiles/G4processes.dir/processes/hadronic/models/im_r_matrix/src/G4XNDeltaTable.cc.o", O_RDONLY) = 1015
[pid 320263] openat(AT_FDCWD, "source/CMakeFiles/G4processes.dir/processes/hadronic/models/im_r_matrix/src/G4CollisionMesonBaryonElastic.cc.o", O_RDONLY) = 1016
[pid 320263] fstat(1016, {st_mode=S_IFREG|0644, st_size=174592, ...}) = 0
[pid 320263] mmap(NULL, 174592, PROT_READ|PROT_WRITE, MAP_PRIVATE, 1016, 0) = 0x7f08aec70000
[pid 320263] close(1016)                = 0
[pid 320263] openat(AT_FDCWD, "source/CMakeFiles/G4processes.dir/processes/hadronic/models/im_r_matrix/src/G4CollisionMesonBaryonElastic.cc.o", O_RDONLY) = 1016
[pid 320263] openat(AT_FDCWD, "source/CMakeFiles/G4processes.dir/processes/hadronic/models/im_r_matrix/src/G4XNNElastic.cc.o", O_RDONLY) = 1017
[pid 320263] fstat(1017, {st_mode=S_IFREG|0644, st_size=161368, ...}) = 0
[pid 320263] mmap(NULL, 161368, PROT_READ|PROT_WRITE, MAP_PRIVATE, 1017, 0) = 0x7f08aec48000
[pid 320263] close(1017)                = 0
```
This follows a pattern like: `[open a object file] -> [fstat a file] -> [mmap from the file] -> [close the file] (expect to start proceeding new file) -> [open the file again] (leak here)`

karuboniru:
Fixes #1362 #851

It seems that with llvm lto we don't have much way to fix the issue than increasing the ulimit. (I strace'd bfd and it is calling setprlimit when it gets any `EMFILE`)

### DIFF
--- a/lib/mapped-file-unix.cc
+++ b/lib/mapped-file-unix.cc
@@ -1,9 +1,21 @@
 #include "common.h"
 
+// for getrlimit, setrlimit
+#include <sys/resource.h>
+
 namespace mold {
 
 MappedFile *open_file_impl(const std::string &path, std::string &error) {
   i64 fd = ::open(path.c_str(), O_RDONLY);
+  // increase rlimit when EMFILE. This is required for llvm lto, since llvm
+  // plugin requires keeping input files open
+  if (fd == -1 && errno == EMFILE) {
+    if (struct rlimit rlim{}; getrlimit(RLIMIT_NOFILE, &rlim) == 0) {
+      rlim.rlim_cur = rlim.rlim_max;
+      setrlimit(RLIMIT_NOFILE, &rlim);
+      fd = ::open(path.c_str(), O_RDONLY);
+    }
+  }
   if (fd == -1) {
     if (errno != ENOENT)
       error = "opening " + path + " failed: " + errno_string();


rui314:
Well I don't think any program should automagically increase the ulimit. Can you file a bug to LLVM instead? If GCC can handle the same thing without hitting the ulimit, LLVM should be able to do the same thing without hitting it, too.

karuboniru:
> If GCC can handle the same thing without hitting the ulimit, LLVM should be able to do the same thing without hitting it, too.

Stracing `ld.bfd` with llvm lto shows it is calling `setprlimit`. Maybe I can try to see what happens with lld.

karuboniru:
Hi @rui314, I tried to remove the if statement at https://github.com/rui314/mold/blob/0841ffcb20b238677699728ae79457196d5203fb/src/lto-unix.cc#L640-L646 to also clean the fd for llvm lto, and the link succeed without any issue (and without any fd leak). 

Is the behavior of "LLVM needs it and takes the ownership of fd." specific to some version of llvm not the version I am currently using (git snapshot), in this case could we proceed with closing the fd anyway?

rui314:
I just tried LTO with Clang, and it seems it works just fine without closing the FDs. Can you try again with the following change?

```
diff --git a/src/lto-unix.cc b/src/lto-unix.cc
index 7d839c87..f009478e 100644
--- a/src/lto-unix.cc
+++ b/src/lto-unix.cc
@@ -637,13 +637,8 @@ ObjectFile<E> *read_lto_object(Context<E> &ctx, MappedFile *mf) {
                << " please make sure you are using the same compiler of the"
                << " same version for all object files";

-  // It looks like GCC doesn't need fd after claim_file_hook() while
-  // LLVM needs it and takes the ownership of fd. To prevent "too many
-  // open files" issue, we close fd only for GCC. This is ugly, though.
-  if (!is_llvm(ctx)) {
-    MappedFile *mf2 = mf->parent ? mf->parent : mf;
-    mf2->close_fd();
-  }
+  MappedFile *mf2 = mf->parent ? mf->parent : mf;
+  mf2->close_fd();

   // Create a symbol strtab
   i64 strtab_size = 1;
```

karuboniru:
I tried the same and the link succeed, however the code for LLVMGOLD suggested that it is using the fd information to track the orgin of file if multiple "handle" is from the same file. I believe it will cause the backend treat all object file as they have same source. 

Don't know if this will case any potenial issue.

https://github.com/llvm/llvm-project/blob/6e0b0038cd65ce726ce404305a06e1cf33e36cca/llvm/tools/gold/gold-plugin.cpp#L580-L585

rui314:
Do you mind if I ask you to try to fix the issue on the LLVM plugin side rather than on the linker side? If you don't have enough bandwidth, I can take a look later. But it seems you have already started reading the source code of the plugin.

karuboniru:
I've no idea how should I test a change in llvm, so I'd appreciate it if you can take time fix the issue in LLVM side. 

But from my understanding of the code in llvmgold I could tell the leading handle thing only wants uniqueness and this could be done by using file path instead of FD. (since if we close the file the fd will no longer be unique)

With the change the FD could be closed after https://github.com/llvm/llvm-project/blob/6e0b0038cd65ce726ce404305a06e1cf33e36cca/llvm/tools/gold/gold-plugin.cpp#L546

umbrageodotus:
@karuboniru updates on reporting to llvm?

PR #1376, opened 2024-12-07
Create cmake-single-platform.yml

Xlabjun:
.github/workflows/generator-generic-ossf-slsa3-publish.yml.github/workflows/generator-generic-ossf-slsa3-publish.yml1581gh pr checkout 1581docs/design.mdXlabjun:master


rui314:
I don't get the point of this file. What is this for?

PR #1383, opened 2024-12-21
Support percent-encoded JSON in --package-metadata

bdrung:
Support for percent-encoded JSON in --package-metadata landed in the ld linker:
https://sourceware.org/git/?p=binutils-gdb.git;a=commitdiff;h=b0cc81e87087bb8a6b12dc1e4fd7f2591927977b
So here comes the the pull request to implement the same in mold. The implementation has changed a lot since https://github.com/rui314/mold/pull/1308.

Specifying the compiler flag `-Wl,--package-metadata=<JSON>` will not work in case the JSON contains a comma, because compiler drivers eat commas. Example:

```
$ echo "void main() { }" > test.c
$ gcc -fuse-ld=mold '-Wl,--package-metadata={"type":"deb","os":"ubuntu"}' test.c
mold: fatal: cannot open "os":"ubuntu"}: No such file or directory
```

The quotation marks in the JSON value do not work well with shell nor make. Specifying the `--package-metadata` linker flag in a `LDFLAGS` environment variable might loose its quotation marks when it hits the final compiler call.

So support percent-encoded and %[string] encoded JSON data in the `--package-metadata` linker flag. Percent-encoding is used because it is a standard, simple to implement, and does take too many additional characters. %[string] encoding is supported for having a more readable encoding.

Bug-binutils: https://sourceware.org/bugzilla/show_bug.cgi?id=32003
Bug-Ubutru: https://bugs.launchpad.net/bugs/2071468

### DIFF
--- a/src/cmdline.cc
+++ b/src/cmdline.cc
@@ -411,28 +411,6 @@ static std::vector<u8> parse_hex_build_id(Context<E> &ctx, std::string_view arg)
   return vec;
 }
 
-template <typename E>
-static std::string
-parse_encoded_package_metadata(Context<E> &ctx, std::string_view arg) {
-  auto flags = std::regex_constants::optimize | std::regex_constants::ECMAScript;
-  static std::regex re(R"(([^%]|%[0-9a-fA-F][0-9a-fA-F])*)", flags);
-
-  if (!std::regex_match(arg.begin(), arg.end(), re))
-    Fatal(ctx) << "--encoded-package-metadata: invalid string: " << arg;
-
-  std::ostringstream out;
-  while (!arg.empty()) {
-    if (arg[0] == '%') {
-      out << (char)((from_hex(arg[1]) << 4) | from_hex(arg[2]));
-      arg = arg.substr(3);
-    } else {
-      out << arg[0];
-      arg = arg.substr(1);
-    }
-  }
-  return out.str();
-}
-
 // Decode a percent and/or %[string] encoded string.
 // Following %[string] encodings are supported:
 //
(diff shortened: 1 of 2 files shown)


doko42:
fyi, this has been accepted for the upcoming binutils 2.44

rui314:
Shouldn’t we report an error if a percent character is not followed by a valid sequence of percent-encoded characters? It looks like it currently silently accepts such malformed percent characters and passes through them without decoding. It feels like it would make finding a usage error harder.

rui314:
Instead of adding all these features, I just renamed `--encoded-package-metadata` `--package-metadata` in 7ddc8f46540ba8fe1202d67706e70dbdc885848d. Let me know if this suffices for your use cases.

bdrung:
The `--package-metadata` parameter does still "support" taking non percent-encoded input data (unless those percent symbols are followed by valid percent-encoding).

rui314:
Is that for backward compatibility with a command line argument that happens to contain '%'? But that's obviously fragile because there's no guarantee that '%' is not followed by a valid sequence of percent-encoded character.

Once we make `--package-metatdata` to take a percent-encoded string, the only safe way to use it is to always percent-encode an argument before passing it to the linker. And I believe that's what you guys are going to do, right?

rui314:
I'm preparing a new release with my change. So starting with the next version, mold will accept `--package-metadata` with a URL-encoded string. You will always need to percent encode an argument before passing it to the linker.

I'll close this PR. If you don't like my decision, please reopen.

doko42:
do you have an estimate for this release?

bdrung:
> Is that for backward compatibility with a command line argument that happens to contain '%'?

Yes.

There should be no '%' in the package metadata anyway. So dropping this backward compatibility should have no negative consequences. So I am fine with your implementation.

rui314:
@bdrung Thank you for confirming!

PR #1386, opened 2024-12-25
Restore executable overwriting behavior on Linux 6.13 or later

jinghao-jia:
Linux upstream has reverted commit `2a010c412853 ("fs: don't block i_writecount during exec")` and the change is merged in the 6.13 cycle. This means the executable overwriting behavior can be restored on Linux kernel version 6.13 or later.

Note that there is probably not much we can do for 6.11 / 6.12 kernels. Although the reversion is backported, many distros do not use upstream sub-version numbers which means we cannot determine whether the kernel has this reversion based on its version number.

All of the builtin tests passed (via `ninja test -C build`):
```
100% tests passed, 0 tests failed out of 386

Total Test time (real) =  16.50 sec

The following tests did not run:
         49 - x86_64-arch-x86_64-z-rewrite-endbr3 (Skipped)
         76 - x86_64-compressed-debug-info (Skipped)
         84 - x86_64-dead-debug-sections (Skipped)
```

I also tried building `ninja` with `mold` on Linux `6.13-rc4` as this is the original problem in #1361. The build finished successfully.

### DIFF
--- a/src/cmdline.cc
+++ b/src/cmdline.cc
@@ -556,15 +556,19 @@ parse_kernel_version(std::string str) {
   return {std::stoi(m[1]), std::stoi(m[2]), std::stoi(m[3])};
 }
 
-// Since 6.11.0, the Linux kernel no longer returns ETXTBSY for open(2)
-// on an executable file that is currently running. This function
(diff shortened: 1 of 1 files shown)


rui314:
Thanks!

PR #1388, opened 2024-12-25
Miscellaneous small cleanups

sicherha:


sicherha:
**issue (blocking):** Commit c045233cf9c42470a2882994e5b3e0691bfb77cf breaks the build on macOS. I'll look into it.

rui314:
It's a matter of taste in most places. I'm fine with most of it, except that I don't like to add `explicit` everywhere.

sicherha:
After way too much trial & error, the macOS build is now fixed.

Personally, I am a proponent of `explicit` not just because [it is recommended by the C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-explicit) but also because I have been bitten by unexpected implicit conversions in the past. In my view, `explicit` is another instance where the C++ standard got its defaults wrong - just like `const`.

Anyway, I have dropped the pertaining commit.

rui314:
I'll merge it if you keep `` instead of replacing them with $()

sicherha:
Commit dropped.

PR #1389, opened 2024-12-26
Small improvements: lambdas and ranges

sicherha:
In principle, compilers ought to be smart enough to detect lambdas that are needlessly declared as capturing, and to generate efficient code for them. Still, I personally like being able to spot non-capturing lambdas right away when reading through code.

Moreover, I propose making use of the C++20 ranges library for greater conciseness and readability.

As always, feel free to reject any changes that run counter to your preferences.

### DIFF
--- a/src/input-files.cc
+++ b/src/input-files.cc
@@ -656,7 +656,7 @@ void ObjectFile<E>::initialize_symbols(Context<E> &ctx) {
 template <typename E>
 void ObjectFile<E>::sort_relocations(Context<E> &ctx) {
   if constexpr (is_riscv<E> || is_loongarch<E>) {
-    auto less = [&](const ElfRel<E> &a, const ElfRel<E> &b) {
+    auto less = [](const ElfRel<E> &a, const ElfRel<E> &b) {
       return a.r_offset < b.r_offset;
     };
 
@@ -1106,7 +1106,7 @@ template <typename E>
 void ObjectFile<E>::compute_symtab_size(Context<E> &ctx) {
   this->output_sym_indices.resize(this->elf_syms.size(), -1);
 
-  auto is_alive = [&](Symbol<E> &sym) -> bool {
+  auto is_alive = [](Symbol<E> &sym) -> bool {
     if (SectionFragment<E> *frag = sym.get_frag())
       return frag->is_alive;
     if (InputSection<E> *isec = sym.get_input_section())
@@ -1399,7 +1399,7 @@ std::span<Symbol<E> *> SharedFile<E>::get_symbols_at(Symbol<E> *sym) {
   });
 
   auto [begin, end] = std::equal_range(sorted_syms.begin(), sorted_syms.end(),
-                                       sym, [&](Symbol<E> *a, Symbol<E> *b) {
+                                       sym, [](Symbol<E> *a, Symbol<E> *b) {
     return a->esym().st_value < b->esym().st_value;
   });
 
--- a/src/gdb-index.cc
+++ b/src/gdb-index.cc
@@ -619,7 +619,7 @@ static std::vector<Compunit> read_compunits(Context<E> &ctx) {
 
   // Uniquify elements because GCC 11 seems to emit one record for each
   // comdat group which results in having a lot of duplicate records.
-  tbb::parallel_for_each(cus, [&](Compunit &cu) {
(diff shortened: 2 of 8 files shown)


src/input-files.cc
```diff
@@ -666,7 +666,7 @@ void ObjectFile<E>::sort_relocations(Context<E> &ctx) {
         continue;
 
       std::span<ElfRel<E>> rels = isec->get_rels(ctx);
-      if (!std::is_sorted(rels.begin(), rels.end(), less))
+      if (!std::ranges::is_sorted(rels, less))
```
rui314, line 669: I avoided std::ranges because it tends to not work with older compilers or the standard C++ libraries.

src/mold.h
```diff
@@ -3097,11 +3097,11 @@ inline bool is_c_identifier(std::string_view s) {
   if (s.empty())
     return false;
 
-  auto is_alpha = [](char c) {
+  static constexpr auto is_alpha = [](char c) {
```
rui314, line 3100: I wouldn't add `constexpr` to a function unless it is necessary.

rui314 (commented):
Overall, these kind of changes are matter of taste. Since this is my personal project, I'd like to keep them as-is unless necessary. I'm happy to adopt `[&]` -> `[]` though.

src/mold.h
sicherha: Shall I change it to `static const` or drop it entirely?

rui314: I'd drop all but `[&]` -> `[]` changes.

src/input-files.cc
sicherha: I'm happy to drop these commits, but just out of curiosity I'll go double-check with various Docker images whether this is still an issue.

My first testing candidate, Ubuntu 20.04, fails for an unrelated reason:
```
CMake Error at third-party/mimalloc/CMakeLists.txt:1 (cmake_minimum_required):
  CMake 3.18 or higher is required.  You are running version 3.16.3
```

src/mold.h
sicherha: I could have explained this better. The thing here is that making `is_alpha` static allows us to make `is_alnum` non-capturing. If we want that, then my instinct is to make `is_alpha` a (compile-time or run-time) constant.

If we drop the `static`, we have to re-add the `[&]` to `is_alnum`.

src/input-files.cc
sicherha: Since the bundled mimalloc library requires CMake 3.18, building on **Ubuntu 20.04** has been broken since commit d2870ac57428be6f5f342b1f75b63d50cc1c9828 in July.

Building on **Debian 11** (which ships CMake 3.18) works fine with GCC 10, even with the range algorithms.
With Clang 11 the build fails for various reasons, among them the range algorithms (my fault) and the use of concepts (not my fault).

Building on **Ubuntu 22.04** with Clang 12, 13 and 14 fails for the same reasons.

Building on **Debian 12** with Clang 13 and 14 fails for the same reasons.
With Clang 15 it fails only due to the range algorithms (my fault).
With Clang 16 everything works fine.

sicherha: Bottom line:
* The minimum required versions are currently **CMake 3.18** and either **GCC 10** or **Clang 15**.
* Definitely no range algorithms for now; they would raise the minimum required Clang version to 16.

src/mold.h
rui314: It's not important, so let's keep it as-is. I believe the compiler is smart enough to optimize these lambdas out.

sicherha: Okay, done. Do you wish to keep the commit that converts the `for`-loop into a call to `std::any_of()` (following the example in `passes.cc:877`)?

rui314: Not really. These are a matter of taste, and I'm honestly not interested in accepting changes like that. I might make a change like that to improve code based on my preferences, but I don't want contributors to make similar non-functional changes because everyone has a different taste.

In fact, since I personally care about details like that and fix them, you don't have to care about them. If you look closely at this project, you might notice that I just accept incoming pull requests without too much code review and then edit the changes after merging. I do this to not waste time on both sides.

sicherha: Dropped; sorry about misunderstanding you there.

PR #1390, opened 2024-12-26
Fix the build with Podman

sicherha:
`podman build` does not support the `--push` shorthand. Replace it with the semantically identical long form `--output=type=registry`.


sicherha:
Scratch that; [it's not implemented yet in Podman](https://github.com/containers/buildah/issues/4671).

PR #1398, opened 2025-01-02
Improve performance by skipping char copies

Slackadays:
All of the lambdas that check if a character is a certain value get their char by value. This creates a lot of unnecessary copies, and if you're doing these in a hot loop, this creates a huge slowdown. I had just that happen in my RISC-V assembler (https://github.com/Slackadays/Chata/commit/727838c2db0dd0255b23bc880a67d4ac59c4f31f) and switching `char c` (actually `int c` for the standard library functions) to `const char& c` to get the value by const reference yielded an instant 15% speedup for free. Best of all, this change does not impact the lambas' usage semantics.

### DIFF
--- a/src/arch-riscv.cc
+++ b/src/arch-riscv.cc
@@ -1025,7 +1025,7 @@ struct Extn {
 // This function returns true if the first extension name should precede
 // the second one as per the rule.
 static bool extn_name_less(std::string_view x, std::string_view y) {
-  auto get_single_letter_rank = [](char c) -> i64 {
+  auto get_single_letter_rank = [](const char& c) -> i64 {
     std::string_view exts = "iemafdqlcbkjtpvnh";
     size_t pos = exts.find_first_of(c);
     if (pos != exts.npos)
--- a/src/cmdline.cc
+++ b/src/cmdline.cc
@@ -378,7 +378,7 @@ static i64 parse_number(Context<E> &ctx, std::string opt,
   return ret;
 }
 
-static char from_hex(char c) {
+static char from_hex(const char& c) {
   if ('0' <= c && c <= '9')
     return c - '0';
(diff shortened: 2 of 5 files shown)


rui314:
All these lambdas can be inlined, and I believe compilers are generally smart enough to inline them, or at least can figure out that the char values passed to these lambdas are not mutated inside the function and apply optimization based on that analysis.

Also copying a char by value is faster than passing a char as a reference, no? char is 1 byte, so copying it is as cheap as or cheaper than passing a pointer to a char.

Slackadays:
> All these lambdas can be inlined, and I believe compilers are generally smart enough to inline them, or at least can figure out that the char values passed to these lambdas are not mutated inside the function and apply optimization based on that analysis.

You'd think so and I did too, but this wasn't the case as of GCC 14 which is where I got the 15% performance boost.
 
> Also copying a char by value is faster than passing a char as a reference, no? char is 1 byte, so copying it is as cheap as or cheaper than passing a pointer to a char.

Actually not, since if you pass it by reference, you can reuse existing registers/memory to make comparisons to the character, but if you pass it by value, the compiler has to allocate a new register just for the comparison.

rui314:
I applied this patch, and it looks like only the change to `filetype.cc` makes a difference in the compiled code. And that change doesn't seem that different anyway. Even if it does, that function is not in a performance-critical pass, so we don't need to optimize that function.

Slackadays:
Then that's weird how most of them made no difference at all, but maybe it has something to do with the standard library versus lamba optimization.

rui314:
You may want to try that yourself. Anyway, we optimize only the code that matters and leave the rest alone. I appreciate your interest in our code. However, I'd accept PRs like this only when they make a measurable performance change with a real-world benchmark using the mold's `--perf` option.

PR #1406, opened 2025-01-24
[ELF][S390X] Skip tests that fail due to buggy code produced by GCC

sicherha:
This pull request addresses testsuite failures we're seeing on S390X Fedora Rawhide with GCC 15 (https://bugzilla.redhat.com/show_bug.cgi?id=2340876).

`bno-symbolic` and `canonical-plt` still fail with GCC 15; adjust the regex accordingly.

Also skip `rodata-name`, which started failing with GCC 14. Whether this has the same root cause as the other two is speculation on my part.


rui314:
Let me take a look `rodata-name` is failing really due to the same issue.

PR #1409, opened 2025-01-29
Honor --compress-debug-sections with --separate-debug-file

0x2Adr1:
Fix a bug when `--compress-debug-sections` is used along with `--separate-debug-file`. The problem was that if both are used together, the debug info file does NOT have its debug sections compressed !!

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -504,9 +504,6 @@ int mold_main(int argc, char **argv) {
   // create as few segments as possible.
   sort_output_sections(ctx);
 
-  if (!ctx.arg.separate_debug_file.empty())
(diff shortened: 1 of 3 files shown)


rui314:
It looks like your test fails on GitHub Actions runners. Can you take a look?

0x2Adr1:
Hmmm maybe we should add `-Wl,--no-detach` in the test ... even though the test passes on my machine, I guess it's possible in the CI that without `--no-detach` when readelf starts reading the file, the debug info file is still being written, could explain the error: `readelf: Warning: [41]: Unexpected value (92) in info field.`

I updated the test, let's see!

PR #1411, opened 2025-02-03
Re-enable the rodata-name.sh test on s390x

iii-i:
The symbol alignment requirement can be lifted using the -munaligned-symbols option, which is available on GCC >=14 and Clang >=19.


rui314:
Thanks, but the logic this test is testing is essentially target-agnostic, so I'm not worried too much about skipping it on s390x. My understanding is that the compiler never creates `.rodata.str1.1` on s390x (instead creates `.rodata.str1.2`.)

PR #1422, opened 2025-03-03
Remove temporary lto files from dependency-file

LarsGullik:
Mark lto temporary files to not be a dependency
    
Removes the temporary file created by the compiler
 from the dependency list created with --dependency-file.
    
 Without this build systems will never report "Nothing to do",
 but always do a re-link of the final object.


rui314:
Thanks!

PR #1423, opened 2025-03-04
Discard section

christycylee:
This is equivalent to discarding a section via `/DISCARD/` in linker scripts.


rui314:
I'd hide this option from the help message as this is an experimental option. I don't want to commit for maintaining it forever at this moment until all the features you need are ready and proved to work nicely.

Do you really need `--no-discard-section`? We have `--foo` and `--no-foo` for boolean options, but options that take an argument rarely have counterpart `--no-` options.

christycylee:
> Do you really need --no-discard-section? We have --foo and --no-foo for boolean options, but options that take an argument rarely have counterpart --no- options.

Unfortunately yes, we need positional overrides here :(

rui314:
OK, then let's give it a shot.

PR #1425, opened 2025-03-07
Remove extraneous build files

christycylee:
`mold` uses `cmake`, the `BUCK` files from `zstd` are not needed.


rui314:
Let's remove them for now. Do you mind if I ask you to fix it in upstream (https://github.com/facebook/zstd) so that the issue is resolved completely?

christycylee:
> Do you mind if I ask you to fix it in upstream

For sure!

rui314:
I'm not sure what these build files are for, but if they are malformed and cannot be used, you may want to remove them from upstream instead of fixing them. If they are unusable, these files have no users.

For tracking purpose, refer to this PR when you make a new PR for the upstream.

PR #1437, opened 2025-04-02
Ignore -m llvm for compatibility

christycylee:
Ignores -mllvm,<arg> for compatibility.

Before: 
```
mold: fatal: unknown -m argument: llvm
```


rui314:
Does my above change work for you?

christycylee:
This works for the case `-mllvm <opt>` but not `-mllvm,<opt>`. We can get rid of the comma from our side so this works great, thanks!

PR #1444, opened 2025-04-24
Print MOLD_VERSION after GNU ld compatibility string too.

satmandu:
As per discussion in https://github.com/rui314/mold/discussions/784#discussioncomment-12936908

I'm not sure if this breaks other scripts that look for mold versions, but this might work around scripts that look for GNU ld versions from looking at the output of `$LD --version`.

### DIFF
--- a/src/config.cc
+++ b/src/config.cc
@@ -4,9 +4,9 @@ namespace mold {
 
 std::string get_mold_version() {
   if (mold_git_hash.empty())
-    return "mold "s + MOLD_VERSION + " (compatible with GNU ld)";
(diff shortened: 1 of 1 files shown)


rui314:
It could make sense to claim that mold is compatible with a specific version of GNU ld (e.g. "compatible with GNU ld 2.37"), but claiming to be compatible with nonexistent version of GNU ld doesn't make sense to me. Overall, I think this is more like an issue of glibc rather than mold; they could have updated their configure script so that it can compile wlth mold, but they didn't. Unless there are lots of programs that need this change, I don't think I want to do this.

Issue #1451, opened 2025-05-04
R_RISCV_CALL_PLTs against weak undef symbols are incorrectly turned into dead loops


ziyao233:
Test `z-dynamic-undefined-weak.sh` hangs on riscv64 when running with Clang,

```
$ TEST_CC="clang -fuse-ld=mold" ./test/z-dynamic-undefined-weak.sh 
Testing z-dynamic-undefined-weak ... + cat
+ clang -fuse-ld=mold -o out/test/riscv64/z-dynamic-undefined-weak/a.o -c -xc - -fPIC
clang: warning: argument unused during compilation: '-fuse-ld=mold' [-Wunused-command-line-argument]
+ clang -fuse-ld=mold -B. -shared -o out/test/riscv64/z-dynamic-undefined-weak/b.so out/test/riscv64/z-dynamic-undefined-weak/a.o
+ clang -fuse-ld=mold -B. -shared -o out/test/riscv64/z-dynamic-undefined-weak/c.so out/test/riscv64/z-dynamic-undefined-weak/a.o -Wl,-z,dynamic-undefined-weak
+ clang -fuse-ld=mold -B. -shared -o out/test/riscv64/z-dynamic-undefined-weak/d.so out/test/riscv64/z-dynamic-undefined-weak/a.o -Wl,-z,nodynamic-undefined-weak
+ cat
+ clang -fuse-ld=mold -o out/test/riscv64/z-dynamic-undefined-weak/e.o -c -xc -
clang: warning: argument unused during compilation: '-fuse-ld=mold' [-Wunused-command-line-argument]
+ clang -fuse-ld=mold -B. -o out/test/riscv64/z-dynamic-undefined-weak/exe1 out/test/riscv64/z-dynamic-undefined-weak/e.o out/test/riscv64/z-dynamic-undefined-weak/b.so
+ out/test/riscv64/z-dynamic-undefined-weak/exe1
+ grep 1 # Hangs here
```

Debugger shows the program is looping in b.so at the call to foo(), which is incorrectly turned into a dead loop,

```
(gdb) disassemble 
Dump of assembler code for function bar:
   0x0000003ff7fd05e4 <+0>:     addi    sp,sp,-32
   0x0000003ff7fd05e6 <+2>:     sd      ra,24(sp)
   0x0000003ff7fd05e8 <+4>:     sd      s0,16(sp)
   0x0000003ff7fd05ea <+6>:     addi    s0,sp,32
[...]

PR #1452, opened 2025-05-04
RISC-V: Correctly relocate R_RISCV_CALL_PLTs against weak undef symbols

ziyao233:
Previously, all R_RISCV_CALL* relocations, regardless through or not through PLT entries, are turned into dead loops if they're against an undefined weak symbol.

This isn't desired in case of R_RISCV_CALL_PLT where the undefined symbol may be satisified at runtime. Let's correctly relocate the symbol in this case.

Fixes: e76f7c0d474e ("Add an initial support of RISC-V")
Closes: https://github.com/rui314/mold/issues/1451


ziyao233:
Already

PR #1454, opened 2025-05-07
Add support for static libatomic.a

snakethatlovesstaticlibs:
It looks like it's possible to accidentally link against `libatomic.so` even when `MOLD_MOSTLY_STATIC` is set. This PR is an attempt at a fix for this

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -277,7 +277,11 @@ int main() {
 }" HAVE_FULL_ATOMIC_SUPPORT)
 
 if(NOT HAVE_FULL_ATOMIC_SUPPORT)
-  target_link_libraries(mold PRIVATE atomic)
+  if (MOLD_MOSTLY_STATIC)
+    target_link_options(mold PRIVATE "-Wl,--push-state,-Bstatic" "-latomic" "-Wl,--pop-state")
+  else()
+    target_link_libraries(mold PRIVATE atomic)
+  endif()
 endif()
 
 # Add -pthread


rui314:
It looks like even with this change, mold is still be linked against libatomic.so because mimalloc requires it. I could tweak mimalloc's cmakefiles to not do that, but I wonder if I really need to. Isn't libatomic.so available on essentially all systems?

snakethatlovesstaticlibs:
> mold is still be linked against libatomic.so because mimalloc requires it

Ah I did not catch that, good catch.

We're using mold as the linker for builds managed by bazel, and some of our bazel configurations restrict access to system libraries. I'm trying to do another pass to see whether we can give that build step access to `libatomic.so`, but was hoping that it could be statically linked as well

rui314:
A mold binary linked with `-DMOLD_MOSTLY_STATIC` still needs the following dynamic libraries other than `libatomic.so`, so it shouldn't be too hard to add one more dynamic library to your whitelist?

```
 0x0000000000000001 (NEEDED)             Shared library: [libdl.so.2]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libpthread.so.0]
 0x0000000000000001 (NEEDED)             Shared library: [librt.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [ld-linux-x86-64.so.2]
```

snakethatlovesstaticlibs:
That's fair, we'll try it this way. Thank you for taking a look

PR #1495, opened 2025-08-04
Fix #1494 as-needed link issue

### LINKED ISSUE #1494: --no-as-needed not working in case: -lfoo -Wl,--no-as-needed -lfoo -Wl,--as-needed
`-Wl,--no-as-needed does not work in the following scenarios:

When main.cpp does not directly depend on libfoo.so but you want to force a link to foo, use the following parameters:
-foo -Wl,-no-as-needed -lfoo -Wl,--as-needed

The readelf -d output shows no forced linking (which is possible with the ld and gold linkers).

If you change it to
-Wl,-no-as-needed -lfoo -Wl,--as-needed -foo
, then forced linking is possible.

Reproduce: W/ gcc-9, x86_64, Ubuntu
```
cat > foo.cpp << EOF
void Foo() {}
EOF
cat > main.cpp << EOF
int main() { return 0; }
EOF

gcc -o libfoo.so -shared -fPIC foo.cpp
gcc -B/path/to/ld main.cpp -L./ -lfoo -Wl,--no-as-needed -lfoo -Wl,--no-as-needed && readelf -d ./a.out
```

a3377626:
* Fix #1494: -lfoo -Wl,--no-as-needed -lfoo -Wl,--as-needed

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -201,7 +201,6 @@ static void read_input_files(Context<E> &ctx, std::span<std::string> args) {
 
   ReaderContext rctx;
   std::vector<ReaderContext> stack;
-  std::unordered_set<std::string_view> visited;
 
   tbb::task_group tg;
   rctx.tg = &tg;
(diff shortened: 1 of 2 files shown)


rui314:
Thanks for the patch! However, the order of files on the command line is important, so we can't casually reverse it. I'll fix this for you.

PR #1507, opened 2025-08-26
 --print-gc-sections and --print-icf-sections can now take =FILE to s…

pedronavf:
`--print-gc-sections` and `print-icf-sections` now take a file (`--print-gc-sections=/tmp/foo.txt`), which makes it easier to get that information without as it doesn't require redirection.

I added to `print-gc-sections` a message with the total bytes saved, similar to the one in ICF, and added subtotal reports. For the `gc-sections` sizes we report the same metric used for ICF (`contents.size()`).

### DIFF
--- a/src/cmdline.cc
+++ b/src/cmdline.cc
@@ -137,9 +137,9 @@ Options:
   --pie, --pic-executable     Create a position-independent executable
     --no-pie, --no-pic-executable
   --pop-state                 Restore the state of flags governing input file handling
-  --print-gc-sections         Print removed unreferenced sections
+  --print-gc-sections[=FILE]  Print, or save in FILE, removed unreferenced sections
     --no-print-gc-sections
-  --print-icf-sections        Print folded identical sections
+  --print-icf-sections[=FILE] Print, or save in FILE, folded identical sections
     --no-print-icf-sections
   --push-state                Save the state of flags governing input file handling
   --quick-exit                Use quick_exit to exit (default)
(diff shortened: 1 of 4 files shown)


rui314:
When we add a new command-line option, we need to ensure that Unix linkers don't diverge too much unnecessarily, so I'm looping in @MaskRay. What do you think?

pedronavf:
Ah, fair enough. It's good to see that there is coordination between linkers. I could work on an lld patch if lld is interested, @maskray.

We are interested in knowing the space saved by ICf and GC, but also to know which functions were considered identical, to help debug crashes/conflicts, or unused. Redirecting the linker's stdout is complicated on some build systems and requires splitting by tool.

MaskRay:
Thanks for tagging me. I agree this is useful. GNU ld feature request: https://sourceware.org/bugzilla/show_bug.cgi?id=33331

(For lld we have a strict requirement for tests on all new features, and the test to modify is located at lld/test/ELF/gc-sections-print.s. Building llvm-project requires powerful hardware, so I can handle the implementation if you are not up to the task.)

rui314:
Thanks @MaskRay for sharing your thoughts and submitting the feature request to GNU ld!

I'm fine with adding new command-line flags, but I'm a bit concerned about displaying the number of saved bytes, because it might give users the wrong impression. The exact number of saved bytes should include the bytes used for padding to satisfy alignment requirements, so simply summing the removed sections doesn't provide an accurate result.

pedronavf:
@rui314 I am counting bytes using the same method that ICF uses: just adding the section contents, which is like the lower bound, right? At some point I was adding the size of the section header, the padding and the size of strings, somnething like this

```
	const auto &sh = isec->shdr();

    // payload contributes to output only if not NOBITS
    u64 payload = (sh.sh_type == SHT_NOBITS) ? 0ull : (u64)isec->sh_size;

    u64 align   = std::max<u64>(1, 1ull << isec->p2align);

    // conservative upper bound on layout contribution
    u64 section_size = (payload + (align - 1)) & ~(align - 1);
```

The total size would also have to take into account neighboring dead sections, less resolcations, hashes  (and debug symbols?). I figured it was too much work for this non-essential metric.

Since ICF was reporting the sum of `content.size()` as bytes saved, I figured it was a reasonable number. We can also kill this number.

mold is so fast that we could link twice, one with `--gc-sections` and one without it and report the difference :D 

@MaskRay If you have the time, by all means do the implementation yourself. It will be quicker that way. I do have the hardware to work with clang, if necessary.

rui314:
I merged the commit but make a follow-up commit to keep the output from the mold command as before. I didn't want to make two independent changes (adding =FILE to the command line *and* change the log output format) at the same time.

PR #1508, opened 2025-08-26
Fix GCC package name for RHEL-based distros in install-build-deps.sh

PowellNGL:
I can only confirm this myself on Fedora and Oracle Linux, but gcc-c++ seems to be the correct package name on every RHEL-based distro (see here: https://pkgs.org/search/?q=gcc-c%2B%2B)


rui314:
I don't remember why I chose that name over gcc-c++, but it looks like gcc-c++ is the correct name indeed. Thanks!

PR #1515, opened 2025-09-11
New section-order feature: noinit

APokorny:
I am consulting another team, working on a micro controller project, trying to move them to mold. They do have a lot of code and icf does help. But these manually placed uninitialized arrays usually eat up all gains by icf. So I need this to remove another linker script.. 

Allow the user to override the section types for custom input sections This is rather common in micro controller projects: to make data structures, that would otherwise be grouped to BSS, magically appear in a special location, the structure is placed inside a custom section which then gets defaulted to SHT_PROGBITS. This then leads to an output section that would be all zeros in your final binary. In gnu linker scripts this could be avoided by adding a NOLOAD attribute to the section.

The use case above could be easily implemented in a driver - simply by setting a pointer or writing a custom allocator.
Still every CPU vendor will give you overly complicated linker scripts examples. This feature only makes sense for supporting older frameworks, and future MCU vendor code.

### DIFF
--- a/src/cmdline.cc
+++ b/src/cmdline.cc
@@ -518,6 +518,7 @@ parse_section_order(Context<E> &ctx, std::string_view arg) {
                std::regex_constants::optimize;
   static std::regex re1(R"(^\s*(TEXT|DATA|RODATA|BSS)(?:\s|$))", flags);
   static std::regex re2(R"(^\s*([a-zA-Z0-9_.][^\s]*|EHDR|PHDR)(?:\s|$))", flags);
+  static std::regex re2_noinit(R"(^\s*noinit:([a-zA-Z0-9_.][^\s]*)(?:\s|$))", flags);
   static std::regex re3(R"(^\s*=(0x[0-9a-f]+|\d+)(?:\s|$))", flags);
   static std::regex re4(R"(^\s*%(0x[0-9a-f]+|\d*)(?:\s|$))", flags);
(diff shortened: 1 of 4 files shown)


rui314:
It's a bit surprising to me that the compiler doesn't automatically set `SHT_NOBITS` for `.lazy_vendor` if the variable defined in the file is uninitialized, like this:

`__attribute__((section(".lazy_vendor"))) char dma_buffer[256];`


as opposed to this:

`__attribute__((section(".lazy_vendor"))) char dma_buffer[256] = { 1, 2, 3, ... };`

I wasn't aware of that compiler behavior. That seems like a bug, or at least a very peculiar behavior, but there might be some backward compatibility issue or something behind it.

As for your patch, I think enabling additional directives inside the `--section-order` command-line option has a risk to making the command line option more complicated. Unlike the linker script, I think we’d prefer to add more separate command-line options to the linker, rather than expanding the functionality of `--section-order`. So, how about this:

1. Add a new command-line flag, say, `--noinit <section-name>`, to set `SHT_NOBIT`S for that section, or
2. Do it automatically. We could define a new command-line option, say, `--promote-to-bss` or something like that, to scan all data sections to check if they are entirely zero, and promote all-zero but allocated sections to `SHT_NOBITS`. This should be easier to maintain.

APokorny:
Yes (2) sounds great.

rui314:
I implemented (2) as `--zero-to-bss`. Let me know if it works for you.

PR #1549, opened 2026-02-11
Fix AS_NEEDED parsing

jmakovicka:
When `INPUT` or `GROUP` contain `AS_NEEDED`, pick the library name inside `AS_NEEDED` instead of `AS_NEEDED` token itself.

Observed with current development GCC 16 `libgcc_s_asneeded.so` containing

```
INPUT ( AS_NEEDED ( -lgcc_s ) )
```


jmakovicka:
duplicate of https://github.com/rui314/mold/pull/1546

PR #1558, opened 2026-03-17
Fix mingw ERROR macro collision

cerisier:
Rename the InputSection action enum value ERROR to ACT_ERROR.

When building with mingw-based Windows toolchains, Windows headers may be pulled in through transitive includes (via tbb for instance). In that case wingdi.h defines ERROR as a macro, which collides with the enum value in input-sections.cc and breaks the build.

Rename the enum value to avoid the macro conflict.

## Context
As part of the [llvm](https://registry.bazel.build/modules/llvm) Bazel C/C++ toolchain module, i'm working on bringing support for choosing arbitrary linkers.
Since the toolchain is a fully hermetic cross-platform cross-compiling toolchain, it needs to build any dependencies from source to run on the platform where the compiler toolchain will be invoked. 

I authored Bazel build files for mold and am able build it from source as part of the end user build and wire it as a linker override for all platforms.
The only issue I encountered is for windows since our toolchain can currently only cross-compile for windows using mingw.

This is the only patch I had to do to make mold compile for windows using a mingw toolchain. 
If supporting mingw is an acceptable goal, this patch was the only one necessary !

### DIFF
--- a/src/input-sections.cc
+++ b/src/input-sections.cc
@@ -121,7 +121,8 @@ void InputSection<E>::copy_contents_to(Context<E> &ctx, u8 *buf, i64 sz) {
   }
 }
 
-typedef enum : u8 { NONE, ERROR, COPYREL, PLT, CPLT } Action;
+// ERROR renamed to ACT_ERROR because mingw Windows headers already #define ERROR.
+typedef enum : u8 { NONE, ACT_ERROR, COPYREL, PLT, CPLT } Action;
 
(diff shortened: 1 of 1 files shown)


rui314:
Thanks. I submitted fd95adda95019fa60c2a1463aab2713fb074f631 instead as I think we can just undefine that macro before use.

cerisier:
Ah yeah much better ! Thanks 🙏

PR #1560, opened 2026-03-21
Various LTO compatibility fixes

tobim:
This branch contains 3 fixes. Each fix is structured as a pair of changes: first a dedicated regression test is added, then a follow-up change fixes the code so that the new test passes.

1. Fix COMDAT handling for mixed regular-object and LTO inputs. When an LTO object and a non-LTO object share the same COMDAT key, mold could let the LTO object claim COMDAT ownership even though the plugin symbol table does not necessarily enumerate all section-level helper symbols. That could cause needed non-LTO COMDAT sections to be discarded and leave dangling relocations. The fix preserves reachable non-LTO COMDAT members in that case.

2. Fix mixed-toolchain GCC fat LTO handling when the active LTO plugin is LLVM. In that case, mold should use the default object code from the GCC fat LTO object instead of treating it as GCC LTO input and failing.

3. Fix archive handling for incompatible LTO members, including thin archives. Previously, mold could fail eagerly while loading an archive even when the incompatible LTO member was never referenced. The fix preserves normal archive semantics by deferring the error until that archive member is actually selected. A setup leading to this situation can arguably be considered a build configuration bug, so I'd be fine dropping this one.

All three cases are handled correctly by lld.

### DIFF
--- a/src/archive-file.cc
+++ b/src/archive-file.cc
@@ -88,6 +88,7 @@ read_thin_archive_members(Context<E> &ctx, MappedFile *mf) {
     if ((begin - data) % 2)
       data++;
 
+    u8 *hdr_begin = data;
     ArHdr &hdr = *(ArHdr *)data;
     u8 *body = data + sizeof(hdr);
     u64 size = atol(hdr.ar_size);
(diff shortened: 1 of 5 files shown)


rui314:
Thanks! Looking good except for one thing: I don't think we should read a symbol table of an archive file. Since we ignore archive symbol table, it feels inconsistent. We should just ignore incompatible or unrecognizable files in archives.

I'll merge this change and then make a follow-up change.

PR #1568, opened 2026-04-18
[ci] Update GitHub Actions to their latest major releases

gruenich:
Fixes most deprecation warnings that Node20.js will stop working in June.


gruenich:
Further reading why this is necessary: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

gruenich:
@rui314 Any reason why this should not be merged?

PR #1576, opened 2026-05-07
Support -w / --no-warnings to suppress warnings

### LINKED ISSUE #1493: [Question] Way to suppress all linker warnings
Hi!
Question: Is there a supported way of suppressing all linker warnings like `-Wl,-w` (for BFD and LLD)?

bamo:
The Warn() machinery has consulted ctx.arg.suppress_warnings since 6d2dbd42 ("Preparation to support Apple ld64's -w option", March 2023), but no cmdline flag ever set it. Wire -w (the GNU/BFD spelling) and --no-warnings (the lld long form) to set the field.

Because Warn::Warn early-returns when the flag is true, before the ctx.arg.fatal_warnings branch sets has_error, -w also cancels --fatal-warnings - matching lld's documented behavior.

The "warnings only" semantics match GCC, clang, and lld: real errors still print and still fail the link. BFD ld's -w is broader and suppresses error messages too, which would hide the cause of link failures.

Fixes https://github.com/rui314/mold/issues/1493

### DIFF
--- a/docs/mold.md
+++ b/docs/mold.md
@@ -778,6 +778,9 @@ but as `-o magic`.
   directory, it is searched from library search paths for the sake of
   compatibility with GNU ld.
 
+* `-w`, `--no-warnings`:
+  Suppress warnings and cancel `--fatal-warnings`.
+
 * `--warn-common`, `--no-warn-common`:
   Warn about common symbols.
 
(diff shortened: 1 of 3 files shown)


rui314:
Thanks!

PR #1577, opened 2026-05-08
Don't emit PT_INTERP for -shared outputs

### LINKED ISSUE #1572: mold should not add .interp when -shared is used
This doesn't break anything, but I noticed this as a difference between lld and mold. It seems mold adds an `.interp` section to `.so` files (at least for my tests in chrome). I've verified that no input .o files contain the section, and that we pass `-shared`. AI suggested adding `--no-dynamic-linker`, but that didn't make it go away either.

bamo:
When --dynamic-linker is passed alongside -shared, mold currently emits a .interp section and a PT_INTERP program header. Both lld and BFD ld silently drop --dynamic-linker for -shared outputs because shared libraries are loaded by the dynamic linker rather than bootstrapped by the kernel, so PT_INTERP serves no purpose there.

Gate .interp creation on (!shared || pie) so we keep it for regular executables, -pie executables, and -shared -pie runnable shared libraries (the libc.so.6-style case), and drop it only for the pure shared-library case. The PT_INTERP program header at output-chunks.cc:219 is conditional on ctx.interp being non-null, so the segment vanishes automatically with no second edit.

Fixes https://github.com/rui314/mold/issues/1572

### DIFF
--- a/src/passes.cc
+++ b/src/passes.cc
@@ -158,7 +158,7 @@ void create_synthetic_sections(Context<E> &ctx) {
   if (ctx.shdr)
     ctx.shstrtab = push(new ShstrtabSection<E>);
 
-  if (!ctx.arg.dynamic_linker.empty())
+  if (!ctx.arg.dynamic_linker.empty() && (!ctx.arg.shared || ctx.arg.pie))
     ctx.interp = push(new InterpSection<E>);
   if (ctx.arg.build_id.kind != BuildId::NONE)
     ctx.buildid = push(new BuildIdSection<E>);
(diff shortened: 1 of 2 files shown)


bamo:
btw, I empirically tested all four cases on lld 18.1.3 and BFD ld 2.42 (Ubuntu 24.04 / aarch64). mold's behavior after the fix is consistent with both linkers on every case, with one exception: lld rejects `-shared -pie` outright (`ld.lld: error: -shared and -pie may not be used together`), whereas mold and BFD both accept it and produce a runnable shared library (the libc.so.6 shape). The regression test pins that more permissive BFD-parity behavior.

PR #1579, opened 2026-05-12
--gdb-index: use DW_FORM_rnglistx index instead of iterating all offset entries

### LINKED ISSUE #1578: GDB warning: .gdb_index address table has a range (...) that overlaps with an earlier range, ignoring .gdb_index
## Versions

| Component | Version |
|-----------|---------|
| mold (buggy) | 2.41.0 |
| lld (no bug) | any |
| Compiler | clang 21.1.8 |
| Build flags | `-O0 -fno-inline -ggdb3 -ggnu-pubnames -fuse-ld=mold -Wl,--gdb-index` |
| Target | x86-64 ELF, DWARF 5 |
| GDB | 17.1 (Debian) |

---

## Minimal Reproducer
File `repro.sh`:
```c++
void may_throw(int x) {
    if (x < 0) throw x;
}

struct Guard {
    int val;
    explicit Guard(int v) : val(v) {}
    ~Guard() { may_throw(val); }
};

int work(int x) {
    int r = 0;
    {
        int a = x + 1;
        {
            Guard g(a);
            {
                int b = x * 2;
                may_throw(b);
                r = b;
            }
            r += a;
        }
        r += x;
    }
    return r;
}

int main() { return work(3); }
```

File `run.sh`:
```bash
#!/bin/bash
set -ex

apt update
apt install -y clang mold gdb

clang++ --version
mold --version
gdb --version

cd /root/
clang++ -O0 -fno-inline -ggdb3 -ggnu-pubnames -fuse-ld=mold -Wl,--gdb-index repro.cc -o repro
gdb -batch -ex quit ./repro
```

Docker command:
```
docker run -it --rm -v $PWD:/root/ debian:sid bash /root/run.sh
```

Logs:
```console
$ docker run -it --rm -v $PWD:/root/ debian:sid bash /root/run.sh
+ apt update
...
+ apt install -y clang mold gdb
...
+ clang++ --version
Debian clang version 21.1.8 (7+b1)
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/lib/llvm-21/bin
+ mold --version
mold 2.41.0 (compatible with GNU ld)
+ gdb --version
GNU gdb (Debian 17.1-4) 17.1
Copyright (C) 2025 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
+ cd /root/
+ clang++ -O0 -fno-inline -ggdb3 -ggnu-pubnames -fuse-ld=mold -Wl,--gdb-index repro.cc -o repro
+ gdb -batch -ex quit ./repro

warning: .gdb_index address table has a range (0x1aca - 0x1af7) that overlaps with an earlier range, ignoring .gdb_index
```

## Investigation
*This report is AI-generated.*
<details>
  <summary>Click to expand</summary>

# mold Bug: `.gdb_index` Address Table Overlaps Due to Incorrect `DW_FORM_rnglistx` Handling

## Summary

When mold links with `--gdb-index` and the DWARF5 compile unit uses `DW_AT_ranges`
with `DW_FORM_rnglistx`, mold reads **all** range-list entries in the CU's
`.debug_rnglists` contribution — including those belonging to child DIEs such as
`DW_TAG_lexical_block` and `DW_TAG_subprogram` — instead of reading only the
single entry indexed by the `DW_FORM_rnglistx` value.  The child DIE ranges are
subsets of (or otherwise overlap with) the CU's own ranges, so GDB detects an
overlap when loading the index and discards the entire `.gdb_index`:

```
warning: .gdb_index address table has a range (0x1aca - 0x1af7) that overlaps
         with an earlier range, ignoring .gdb_index
```

lld does not exhibit this behaviour.

### The faulty code

In `src/gdb-index.cc`, `read_address_ranges()` handles the CU root DIE's
`DW_AT_ranges` attribute.  For DWARF 5, there are two possible forms:

* `DW_FORM_sec_offset` — an absolute byte offset into `.debug_rnglists`
* `DW_FORM_rnglistx` — an index into the CU's offset array inside
  `.debug_rnglists`

The `DW_FORM_sec_offset` case is handled correctly (lines 452-453):
```cpp
if (ranges.form == DW_FORM_sec_offset) {
    read_rnglist_range<E>(vec, buf + ranges.value, addrx, low_pc.value);
```

The `DW_FORM_rnglistx` case (the `else` branch, **lines 454-464**) is wrong:
```cpp
} else {
    if (rnglists_base == -1)
        Fatal(ctx) << "--gdb-index: missing DW_AT_rnglists_base";

    u8 *base = buf + rnglists_base;
    i64 num_offsets = *(U32<E> *)(base - 4);   // reads offset_entry_count
    Offset *offsets = (Offset *)base;

    for (i64 i = 0; i < num_offsets; i++)       // ← iterates ALL entries
        read_rnglist_range<E>(vec, base + offsets[i], addrx, low_pc.value);
}
```

`ranges.value` holds the index (the unsigned LEB128 operand of `DW_FORM_rnglistx`)
but is **never used** in this branch.  Instead the code reads `num_offsets` from
`*(U32<E> *)(base - 4)`, which is the `offset_entry_count` field of the
`.debug_rnglists` contribution header immediately preceding the offset array.

### What `offset_entry_count` actually counts

The DWARF 5 `.debug_rnglists` format (§7.28) allocates one offset-table entry
for **each DIE in the CU** that uses `DW_FORM_rnglistx` for its `DW_AT_ranges`
attribute — not just the CU root.  This includes `DW_TAG_subprogram`,
`DW_TAG_lexical_block`, `DW_TAG_inlined_subroutine`, etc.

Schematically, for the minimal reproducer (one function, three nested scopes,
EH landing pads making each scope non-contiguous):

```
offset_entry_count = N   (e.g. 5: CU root + subprogram + 3 lexical_blocks)
offsets[0]  →  range list for DW_TAG_compile_unit
offsets[1]  →  range list for DW_TAG_subprogram  (work)
offsets[2]  →  range list for outer  DW_TAG_lexical_block
offsets[3]  →  range list for middle DW_TAG_lexical_block
offsets[4]  →  range list for inner  DW_TAG_lexical_block
```

The CU's `DW_AT_ranges` attribute is `DW_FORM_rnglistx` with value 0 (pointing
to `offsets[0]`), but mold reads entries 0 through 4 and adds all of them to the
`.gdb_index` address table.

### Why the ranges overlap

Nested lexical blocks are, by definition, subsets of their enclosing scope.
When `work()` has an EH landing pad, the compiler generates:

| DIE | Address ranges |
|-----|----------------|
| CU / subprogram | `[0x1a80, 0x1aca)` + `[0x1aca, 0x1af7)` |
| outer lexical_block | `[0x1a90, 0x1aca)` + `[0x1aca, 0x1af7)` |
| middle lexical_block | `[0x1a98, 0x1aca)` + `[0x1aca, 0x1af8)` |
| inner lexical_block | `[0x1aa0, 0x1aca)` + `[0x1aca, 0x1af7)` |

All of the lexical-block ranges are strict subsets of the CU's ranges.  Once the
CU range `[0x1a80, 0x1aca)` is inserted into the address table, inserting the
outer block's `[0x1a90, 0x1aca)` (which is inside it) is an overlap.

GDB's `create_addrmap_from_gdb_index()` (`gdb/dwarf2/read-gdb-index.c:1439`)
calls `addrmap_mutable::set_empty(lo, hi-1, cu_unit)`, which returns `false`
if any byte in `[lo, hi-1]` is already mapped.  On the first conflict it prints
the warning and returns `false`, which causes the caller to ignore the entire
`.gdb_index` and fall back to a full linear DWARF scan.

### Why it triggers specifically in DWARF 5 with clang

With DWARF 5 enabled (clang 14+, `-ggdb3` on modern Debian), clang uses
`DW_FORM_rnglistx` for all range-list references and adds `DW_AT_rnglists_base`
to the CU root DIE.  This causes the buggy `else` branch in mold to execute.
With DWARF 4 (`-gdwarf-4`), ranges use `DW_FORM_sec_offset` and the
correctly-handled `DW_FORM_sec_offset` branch is taken instead.

### Why RAII + exceptions are the precipitating factor

A `DW_TAG_lexical_block` normally has contiguous code and is represented with
`DW_AT_low_pc` + `DW_AT_high_pc` (no rnglist entry, so it does not inflate
`offset_entry_count`).  The block becomes non-contiguous — and is thus
represented with `DW_AT_ranges` / `DW_FORM_rnglistx` — only when the compiler
emits EH landing pads whose addresses fall outside the normal code range of the
block.  This happens when:

1. The block contains a variable with a non-trivial destructor (RAII), **and**
2. A call inside the block can throw (so the compiler must generate a cleanup
   landing pad for the destructor).

Both conditions are met by `Guard` in the reproducer.

---

## Correct Fix

Replace the loop that iterates all `num_offsets` entries with a single lookup
using `ranges.value` (the actual index from `DW_FORM_rnglistx`):

```cpp
// src/gdb-index.cc — inside read_address_ranges(), the else branch (~line 454)

// BEFORE (wrong: reads every entry in the offset table, including child DIEs)
u8 *base = buf + rnglists_base;
i64 num_offsets = *(U32<E> *)(base - 4);
Offset *offsets = (Offset *)base;
for (i64 i = 0; i < num_offsets; i++)
    read_rnglist_range<E>(vec, base + offsets[i], addrx, low_pc.value);

// AFTER (correct: reads only the entry the CU's DW_AT_ranges points to)
u8 *base = buf + rnglists_base;
Offset *offsets = (Offset *)base;
read_rnglist_range<E>(vec, base + offsets[ranges.value], addrx, low_pc.value);
```

This mirrors how GDB itself resolves a `DW_FORM_rnglistx` attribute
(`gdb/dwarf2/read.c`, `read_rnglist_index()`):

```c
/* Offset in .debug_rnglists of the offset for RNGLIST_INDEX.  */
ULONGEST start_offset =
    rnglist_base + rnglist_index * cu->header.offset_size;
```

GDB uses the index to find exactly one entry in the table; it does not loop over
all entries.

---

## Impact

- GDB silently falls back to slow linear DWARF scanning on every load of the
  binary.
- For large C++ binaries with many RAII scopes and exception-handling paths,
  startup time in GDB can increase significantly.
- The binary is still correct and fully debuggable; only the index is unusable.
- The warning is printed once per GDB session.

</details>

bamo:
A compile unit's DW_AT_ranges attribute in DW_FORM_rnglistx form holds an index into the offset table at DW_AT_rnglists_base; the corresponding entry locates the range list for that DIE. mold was ignoring the index and iterating every offset, picking up range lists belonging to child DIEs (subprograms, lexical_blocks, inlined_subroutines) that use DW_FORM_rnglistx as well.

The resulting .gdb_index address table contained child-DIE ranges nested inside the CU's range. GDB 17 detects the overlap, prints "warning: .gdb_index address table has a range (...) that overlaps with an earlier range, ignoring .gdb_index", and discards the index. Older GDBs accept the bloated table without complaint -- the duplicate entries all point to the same CU, so lookups still resolve correctly - which is why this went undetected for years.

The bug only triggers when a CU contains a non-contiguous lexical_block, which clang emits for scopes with EH landing pads from throwing RAII destructors. Replace the loop with offsets[ranges.value], matching the DW_FORM_sec_offset branch and GDB's own read_rnglist_index.

The new test invokes clang++ directly since DW_FORM_rnglistx is a clang-specific behavior; gcc through 13 uses DW_FORM_sec_offset, which mold has always handled correctly. The test skips cleanly if clang++ isn't installed or doesn't emit rnglistx for the source.

Reported by @MikeWang000000

Fixes https://github.com/rui314/mold/issues/1578

### DIFF
--- a/src/gdb-index.cc
+++ b/src/gdb-index.cc
@@ -456,11 +456,9 @@ read_address_ranges(Context<E> &ctx, const Compunit &cu) {
         Fatal(ctx) << "--gdb-index: missing DW_AT_rnglists_base";
 
       u8 *base = buf + rnglists_base;
-      i64 num_offsets = *(U32<E> *)(base - 4);
       Offset *offsets = (Offset *)base;
-
-      for (i64 i = 0; i < num_offsets; i++)
-        read_rnglist_range<E>(vec, base + offsets[i], addrx, low_pc.value);
+      read_rnglist_range<E>(vec, base + offsets[ranges.value], addrx,
+                            low_pc.value);
     }
     return vec;
   }
--- a/test/gdb-index-dwarf5-rnglistx.sh
+++ b/test/gdb-index-dwarf5-rnglistx.sh
@@ -1,0 +1,58 @@
+#!/bin/bash
+. $(dirname $0)/common.inc
+
+# Bug only triggers on DW_FORM_rnglistx, which clang emits but gcc through 13 doesn't.
+[ $MACHINE = $(uname -m) ] || skip
+clang++ --version >& /dev/null || skip
+echo 'int main() {}' | clang++ -gdwarf-5 -g -fexceptions -c -o /dev/null -xc++ - 2>/dev/null || skip
+
+# Throwing destructor + nested scopes => non-contiguous lexical_blocks => rnglistx.
+cat <<'EOF' > $t/a.cc
+void may_throw(int x) { if (x < 0) throw x; }
+struct Guard {
+  int val;
+  explicit Guard(int v) : val(v) {}
+  ~Guard() { may_throw(val); }
+};
+int work(int x) {
+  int r = 0;
+  { int a = x + 1;
+    { Guard g(a);
+      { int b = x * 2; may_throw(b); r = b; }
+      r += a; }
+    r += x; }
+  return r;
+}
+int main() { return work(3); }
+EOF
+
+clang++ -c -o $t/a.o $t/a.cc -gdwarf-5 -ggdb3 -fexceptions -O0 -fno-inline
+readelf --debug-dump=info $t/a.o | grep -E 'DW_AT_ranges.*\(index:' > /dev/null || skip
+
+clang++ -B. -o $t/exe $t/a.o -gdwarf-5 -ggdb3 -ggnu-pubnames -fexceptions \
+  -O0 -fno-inline -Wl,--gdb-index
+
+readelf -WS $t/exe | grep -F .gdb_index
+
+readelf --debug-dump=gdb_index $t/exe |
+  awk '/^Address table:/ { in_tab=1; next }
+       /^Symbol table:/ { in_tab=0 }
+       in_tab && NF==3 && $1 ~ /^[0-9a-f]+$/ { print $1, $2 }' |
+  sort > $t/addrtab
+
+# 16-char zero-padded hex; lexicographic compare == numeric compare.
+awk '{
+    if (NR > 1 && $1 < prev_hi) {
+      printf "overlap: previous range ends 0x%s, current starts 0x%s\n", prev_hi, $1
+      exit 1
+    }
+    prev_hi = $2
+  }' $t/addrtab
+
+# A wrong-index regression would be overlap-free but miss main's address.
+main_addr=$(readelf -s $t/exe | awk '$NF == "main" { print $2; exit }')
+awk -v a="$main_addr" '
+  BEGIN { found = 0 }
+  $1 <= a && a < $2 { found = 1; print "found main 0x" a " in [0x" $1 ", 0x" $2 ")"; exit }
+  END { if (!found) { print "main 0x" a " not covered"; exit 1 } }
+' $t/addrtab


bamo:
CI blocked by https://github.com/rui314/mold/pull/1580

PR #1581, opened 2026-05-13
chore(tests): add missing +x for 2 tests

marxin:


bamo:
the msys build failure here will be fixed by https://github.com/rui314/mold/pull/1580

PR #1586, opened 2026-05-19
Fix ambiguous readelf option in gdb-index-dwarf5 test

andreas-schwab:
The binutils readelf command now accepts a --debug-dir=<DIR> option which
makes --debug=gdb_index ambiguous.


rui314:
Thanks!

PR #1587, opened 2026-05-19
perf: cache InputSection name to avoid per-call strlen

bamo:
InputSection::name() implicitly converts a `const char *` to string_view, which calls strlen on the shstrtab entry at every call. Precompute the name as a string_view once in the constructor.

Linking clang Debug is ~1.3% faster; mlir-opt Debug is ~2.6% faster.


rui314:
Thanks!

PR #1589, opened 2026-05-19
perf: pre-fault ConcurrentMap pages with MAP_POPULATE

bamo:
ConcurrentMap::insert reads ent.key before writing, so on an empty slot the read first installs a PTE pointing at the system zero page, then the CAS triggers wp_page_copy and a cross-CPU TLB shootdown. Under heavy parallel insertion this shows up as ~3% of link time in smp_call_function_many_cond plus a much larger inclusive cost in asm_exc_page_fault.

Pass MAP_POPULATE to pre-fault the whole map in the mmap() thread. Reads no longer install zero-page PTEs and writes no longer trigger CoW. Linking clang-23 Debug is ~5% faster, llvm-tblgen ~7%, clang-tblgen ~9%; max RSS rises by 0.1-0.6%. Falls back cleanly on platforms that don't define MAP_POPULATE.

### DIFF
--- a/lib/lib.h
+++ b/lib/lib.h
@@ -395,14 +395,19 @@ public:
     this->nbuckets = std::max<i64>(MIN_NBUCKETS, bit_ceil(nbuckets));
     i64 bufsize = sizeof(Entry) * this->nbuckets;
 
-    // Allocate a zero-initialized buffer. We use mmap() if available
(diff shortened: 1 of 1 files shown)


rui314:
Nice. Thanks!

Issue #336, opened 2022-02-09
Perhaps a dumb question: how to use mold for chromium.


Alex313031:
Hello, Im working on a chromium fork called Thorium > https://github.com/Alex313031/Thorium and would like to test using mold instead of lld during linking of the .so files and executables. How would I do this? I know this was one of the inspirations of the creator when making mold. Do I need to install or build/install mold first, or is it included in a chromium checkout? Can I just modify the main BUILD.gn to use LDFLAGS+= [ "-fuse-ld=mold", ]?

matu3ba:
Please explain what you tried from the instructions in the README.

Alex313031:
@matu3ba Okay, I have compiled it with stripping, -03, lto, and -march=haswell and -mtune=haswell for max performance :)
I then set -fuse-ld=/usr/local/bin/mold in the main BUILD.gn for Thorium.
Seems to work, but now another question. How do I pass llvm flags to it, as it doest recognize the mllvm option. Is mold one on one compatible with llvm flags, or is there something else I need to do?

For example, my main BUILD.gn normally has some loop optimizations I added like "-Wl,-mllvm,-aggressive-ext-opt" and "-Wl,-mllvm,-enable-loop-distribute",
I have tried omitting the -Wl, ommitting the -mllvm, and ommiting both, but mold doesn't recognize them, wheras lld does, and I have the same set as cflags for clang, which also recognizes it.

rui314:
Are you trying LTO? LTO is probably too new to try. We don't and unlikely to support mllvm options as they are LLVM-specific options.

monarc99:
i tried to compile chromium with mold. I use this patch to remove the mllvm options.

https://github.com/stha09/chromium-patches/blob/master/chromium-100-compiler.patch

After that the compile stops with the error:


```

ninja: Entering directory `out/Release'
[2040/50333] LINK ./chrome_sandbox
FAILED: chrome_sandbox 
[...]

PR #1594, opened 2026-05-26
ICF: don't fold sections with different alignments

justanotheranonymoususer:
LeafHasher/LeafEq and compute_digest hashed and compared section contents but not p2align, so two sections with identical bytes but different alignment requirements could be folded together. The follower then resolved to the leader's address, which may not satisfy the follower's alignment, and an aligned SIMD load (e.g. movaps) against it would SIGSEGV.

Include p2align in the hash and equality check so such sections are kept distinct.

---

This fixes the following issue:

Building chromium with mold causes a `mksnapshot` segfault:
```
[1920/1920] ACTION //v8:run_mksnapshot_default(//build/toolchain/linux/unbundle:default)
ninja: job failed: python3 ../../v8/tools/run.py ./mksnapshot --turbo_instruction_scheduling --stress-turbo-late-spilling --target_os=linux --target_arch=x64 --embedded_src gen/v8/embedded.S --embedded_variant Default --random-seed 314159265 --startup_blob snapshot_blob.bin --no-native-code-counters
Return code is -11
ninja: subcommand failed
```

Reported here: https://github.com/rui314/mold/issues/336#issuecomment-2128636987

---

This PR is heavily AI-assisted, but reproduced locally (issue and fix). Please review carefully.

### DIFF
--- a/src/icf.cc
+++ b/src/icf.cc
@@ -161,6 +161,7 @@ template <typename E>
 struct LeafHasher {
   size_t operator()(InputSection<E> *isec) const {
     u64 h = hash_string(isec->contents);
+    h = combine_hash(h, isec->p2align);
     for (FdeRecord<E> &fde : isec->get_fdes()) {
       u64 h2 = hash_string(fde.get_contents(isec->file).substr(8));
       h = combine_hash(h, h2);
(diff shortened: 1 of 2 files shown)


agrieve:
I suspect #1600 is solving the same problem in a slightly different way.

rui314:
Thank you for your pull request. I think @agrieve's one would be better, as we want to merge sections with identical contents but just want to choose the largest alignment requiremnt.

PR #1599, opened 2026-06-03
icf: Do not fold relocations pointing to preemptible symbols

### LINKED ISSUE #1598: Incorrect Identical Code Folding for shim
Here's a gemini explanation of a crashing chrome test when linked with mold:
 
* `libclearkeycdm.so` allocates video frame memory locally using its own PartitionAlloc instance (via `base::UncheckedMalloc`).  
* When freeing this memory, it calls `base::UncheckedFree`, which should route to the local allocator shim.  
* However, `mold`'s ICF incorrectly folded `base::UncheckedFree` into `SkData::operator delete` because they appeared structurally identical (both being simple tail-calls/jumps).  
* `SkData::operator delete` jumps to the dynamic `operator delete` (PLT), which resolves to the host process's (`browser_tests`) PartitionAlloc.  
* This resulted in memory allocated by the CDM's local PartitionAlloc being freed by the host's PartitionAlloc, triggering a `PA_NOTREACHED` (`FreeInUnknownRoot`) allocator mismatch crash.

agrieve:
If a relocation points to a preemptible (imported) symbol, it should not be considered identical to a relocation pointing to a local symbol, even if their targets currently have identical bodies. Preemptible symbols can be overridden at runtime, so we must preserve their unique identity.

Fixes #1598

### DIFF
--- a/src/icf.cc
+++ b/src/icf.cc
@@ -255,7 +255,7 @@ static Digest compute_digest(Context<E> &ctx, InputSection<E> &isec) {
   auto hash_symbol = [&](Symbol<E> &sym) {
     InputSection<E> *isec = sym.get_input_section();
 
-    if (!sym.file) {
+    if (!sym.file || sym.is_imported) {
       hash('1');
       hash((u64)&sym);
     } else if (SectionFragment<E> *frag = sym.get_frag()) {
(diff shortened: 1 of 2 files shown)


rui314:
Thanks!

PR #1601, opened 2026-06-04
Allow PC-relative relocations against absolute symbols in DSOs and PIEs.

### LINKED ISSUE #1573: mold cannot link v8 with x86: R_386_PC32 relocation at offset 0xe6 against symbol `' can not be used; recompile with -fPIC
Building:
```
autoninja libchrome.so
```

with `args.gn`:
```
target_os = "android"
target_cpu = "x86"
use_mold = true
```

yields:
```
mold: error: obj/v8/v8_base_without_compiler/liftoff-compiler.o:(.text): R_386_PC32 relocation at offset 0xe6 against symbol `' can not be used; recompile with -fPIC
mold: error: obj/v8/v8_base_without_compiler/liftoff-compiler.o:(.text): R_386_PC32 relocation at offset 0x119 against symbol `' can not be used; recompile with -fPIC
mold: error: obj/v8/v8_base_without_compiler/liftoff-compiler.o:(.text): R_386_PC32 relocation at offset 0x119 against symbol `' can not be used; recompile with -fPIC
mold: error: obj/v8/v8_base_without_compiler/liftoff-compiler.o:(.text): R_386_PC32 relocation at offset 0xe4 against symbol `' can not be used; recompile with -fPIC
mold: error: obj/v8/v8_base_without_compiler/liftoff-compiler.o:(.text): R_386_PC32 relocation at offset 0xe4 against symbol `' can not be used; recompile with -fPIC
mold: error: obj/v8/v8_base_without_compiler/liftoff-compiler.o:(.text): R_386_PC32 relocation at offset 0xe4 against symbol `' can not be used; recompile with -fPIC
mold: error: obj/v8/v8_base_without_compiler/liftoff-compiler.o:(.text): R_386_PC32 relocation at offset 0xfe against symbol `' can not be used; recompile with -fPIC
mold: error: obj/v8/v8_base_without_compiler/liftoff-compiler.o:(.text): R_386_PC32 relocation at offset 0xe4 against symbol `' can not be used; recompile with -fPIC
mold: error: obj/v8/v8_base_without_compiler/liftoff-compiler.o:(.text): R_386_PC32 relocation at offset 0xfe against symbol `' can not be used; recompile with -fPIC
mold: error: obj/v8/v8_base_without_compiler/liftoff-compiler.o:(.text): R_386_PC32 relocation at offset 0xe4 against symbol `' can not be used; recompile with -fPIC
```

Links fine with lld, and I've verified that -fPIC is used.

agrieve:
Historically, mold reported an error for PC-relative relocations against absolute symbols in shared libraries and PIEs, because they are marked as imported (preemptible) and dynamic linkers do not support PC-relative dynamic relocations.

However, if the absolute symbol is defined locally, it does not need a dynamic relocation if we assume it is not preempted (matching lld behavior).

This change updates Symbol::is_absolute() to recognize locally defined symbols as absolute even if they are marked as imported for preemption. It also updates the scan_pcrel decision table to allow Absolute symbols (action NONE) in Shared objects and PIEs.

Added a test case test/pcrel-absolute.sh to verify this behavior.

Fixes #1573

### DIFF
--- a/src/input-sections.cc
+++ b/src/input-sections.cc
@@ -183,8 +183,8 @@ void InputSection<E>::scan_pcrel(Context<E> &ctx, Symbol<E> &sym,
   // linker generally does not support PC-relative relocations.
   static Action table[][4] = {
     // Absolute  Local    Imported data  Imported code
-    {  ERROR,    NONE,    ERROR,         PLT    },  // Shared object
-    {  ERROR,    NONE,    COPYREL,       CPLT   },  // Position-independent exec
+    {  NONE,     NONE,    ERROR,         PLT    },  // Shared object
+    {  NONE,     NONE,    COPYREL,       CPLT   },  // Position-independent exec
     {  NONE,     NONE,    COPYREL,       CPLT   },  // Position-dependent exec
   };
 
--- a/src/mold.h
+++ b/src/mold.h
@@ -3442,8 +3442,8 @@ inline bool Symbol<E>::is_absolute() const {
   if (is_remaining_undef_weak())
     return true;
 
-  return !is_imported && !get_frag() && !get_input_section() &&
-         !get_output_section();
+  return file && !file->is_dso && !esym().is_undef() &&
+         !get_frag() && !get_input_section() && !get_output_section();
 }
 
 template <typename E>
--- a/test/pcrel-absolute.sh
+++ b/test/pcrel-absolute.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+. $(dirname $0)/common.inc
+
+# Only run on x86_64 and i386 for simplicity of assembler syntax
+[[ $MACHINE = x86_64 ]] || [[ $MACHINE = i386 ]] || skip
+
+# Define absolute symbol
+cat <<EOF | $CC -c -o $t/a.o -xassembler -
+.globl foo
+foo = 0
+EOF
+
+# Use PC-relative relocation to it
+cat <<EOF | $CC -fPIC -c -o $t/b.o -xassembler -
+.text
+.globl bar
+bar:
+  .long foo - .
+EOF
+
+# Try to link into shared library.
+# Should succeed with our fix.
+$CC -B. -shared -o $t/libtest.so $t/a.o $t/b.o


rui314:
Allowing PC-relative relocations against absolute symbols is wrong, isn’t it? Let’s take your test as an example. `bar + &bar` must evaluate to 0, since that is what the assembly expression `bar: .long foo - .` means, but at runtime we’ll get something different for PIC/PIE. Unless we know the runtime address of `bar` at link-time, we can't compute a PC-relative address for `bar` at link-time.

I’m more interested in knowing what motivated you to make this change. Is there existing code that relies on this behavior?

agrieve:
Similar to the V8 one, this can be fixed in code:

```
--- a/third_party/protobuf/src/google/protobuf/metadata_lite.h
+++ b/third_party/protobuf/src/google/protobuf/metadata_lite.h
@@ -119,7 +119,7 @@
   template <typename T>
   PROTOBUF_NDEBUG_INLINE void MergeFrom(const InternalMetadata& other) {
     if (other.have_unknown_fields()) {
-      DoMergeFrom<T>(other.unknown_fields<T>(nullptr));
+      DoMergeFrom<T>(other.PtrValue<Container<T>>()->unknown_fields);
     }
   }
```

Fixing in clang would be great as well, but this was the path of least resistance :).

rui314:
I tested this, and it does indeed look like a bug in the LLVM i386 backend. LLVM does not emit a PC-relative relocation against the null absolute symbol for other targets. It was very likely just neglected; i386 does not get much attention these days.

It looks like the following single-line change can fix the issue on the LLVM side:

```
diff --git a/llvm/lib/Target/X86/X86Subtarget.cpp b/llvm/lib/Target/X86/X86Subtarget.cpp
index 44d71e006496..181e190f1fb2 100644
--- a/llvm/lib/Target/X86/X86Subtarget.cpp
+++ b/llvm/lib/Target/X86/X86Subtarget.cpp
@@ -241,7 +241,7 @@ bool X86Subtarget::isLegalToCallImmediateAddr() const {
   // FIXME: I386 PE/COFF supports PC relative calls using IMAGE_REL_I386_REL32
   // but WinCOFFObjectWriter::RecordRelocation cannot emit them.  Once it does,
   // the following check for Win32 should be removed.
-  if (Is64Bit || isTargetWin32())
+  if (Is64Bit || isTargetWin32() || isPositionIndependent())
     return false;
   return isTargetELF() || TM.getRelocationModel() == Reloc::Static;
```

Given that
- this is a relocation usage error in LLVM,
- emitting an error from the linker for this unrepresentable relocation use case is logically correct, and
- it occurs only with the LLVM i386 backend,

I do not think it is a good idea to silence the error in mold.

Can you try the above change on the LLVM side?

agrieve:
Thanks, giving it a go: https://github.com/llvm/llvm-project/pull/202370

rui314:
Since https://github.com/llvm/llvm-project/pull/202370 has been merged to LLVM, I'll close this bug.

agrieve:
Nice work!

PR #1606, opened 2026-06-09
pre-compute is_debug_section flag on InputSection

aarchitecture:
`get_tombstone()` is called for every non-alloc relocation in every arch backend. i replaced the `name.starts_with(".debug_")` string prefix comparision since that result is invariant for the lifetime of the `InputSection`. it now precomputes three boolean flags in the `InputSection` constructor, which are `is_debug_section`, `is_debug_line`, and `is_debug_loc_ranges`. these replace their respetcive string comparisons.


rui314:
Performance optimizations should be backed by benchmark measurements. In general, we do _not_ optimize code without evidence that it matters, in order to avoid premature optimization.

aarchitecture:

PR #1607, opened 2026-06-09
hash-based CIE dedup

aarchitecture:
`uniquify_cies()` is called for CIE records in every object file during ICF. the old implementation does linear scan over all seen CIE records, which is quadratic, so i replaced it with an `std::unordered_map` keyed on `cie.get_contents()`, which buckets records by their raw content hash first. only records in the same bucket fall through to the full `cie_equals()` comparison.

### DIFF
--- a/src/icf.cc
+++ b/src/icf.cc
@@ -69,6 +69,7 @@
 
 #include <array>
 #include <cstdio>
+#include <unordered_map>
 #include <fstream>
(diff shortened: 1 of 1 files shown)


rui314:
I believe `cies` is usually very small -- typically contains just one, isn't it?

aarchitecture:

PR #1608, opened 2026-06-09
parallelize flatten in create_output_sections

aarchitecture:
`create_output_sections()` contains a serial loop over `ctx.osec_pool` that flattens per-object-file `members_vec` into a flat `members` vector for each output section. i've replaced the range-for loop with `tbb::parallel_for` since each `OutputSection` has its own independent `members_vec` and `members`; trivial to parallelize.


aarchitecture:

PR #1611, opened 2026-06-18
build: add convienence flags for selecting the C++ stdlib on Linux

a5ehren:
* MOLD_USE_LIBCXX configures Clang builds to use LLVM libc++. This is already the default on macOS and BSD.
* MOLD_USE_LIBCXX and MOLD_STATIC_LIBCXX require Clang; libc++ builds with GCC are unsupported.
* Added MOLD_STATIC_* flags for several libraries to provide finer control than MOLD_MOSTLY_STATIC.

Note that MOLD_MOSTLY_STATIC + Clang enables MOLD_USE_LIBCXX by default. This breaks dist.sh until my other PR is accepted.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,14 +43,17 @@
 # of the mold linker of the same version will have the exact same set of
 # features and behave exactly the same.
 
-cmake_minimum_required(VERSION 3.14)
+cmake_minimum_required(VERSION 3.14...4.3)
 project(mold VERSION 2.41.0)
 
 include(CMakeDependentOption)
+include(CheckCXXSourceCompiles)
 include(CheckSymbolExists)
 include(GNUInstallDirs)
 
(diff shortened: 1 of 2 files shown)


rui314:
What's the motivation for using libc++ over libstdc++?

a5ehren:
Some people think it’s faster (does not show up in mold benchmarks) but I am doing it here to simplify the build container in dist.sh - moving to Debian 10 and using static libc++ lets you get rid of the entire gcc build stage.

rui314:
I can merge this, but honestly do you think we need to support gcc + libc++ use case? If we limit it to clang it would need much simpler.

a5ehren:
Agree, I can’t think of anyone that would want to use libc++ with GCC, just included it to be complete. I’ll take it out and update.

a5ehren:
Updated

PR #1612, opened 2026-06-18
build: update dist.sh by refreshing all distro images 

a5ehren:
* update x86-64 to Debian 10. This has glibc 2.28 which matches the oldest actually used distro these days (rhel 8) and gcc-8.5 is new enough to bootstrap clang and libc++, which saves the whole gcc build step
* set the riscv64 target to use debian Trixie, which was the first stable release for that arch
* bullseye (arm,ppac,s390), trixie (riscv), and the loongarch sid snapshot all have clang-19 available
* make sure all images install the libc++-19-dev package to get libc++ static version

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,14 +43,17 @@
 # of the mold linker of the same version will have the exact same set of
 # features and behave exactly the same.
 
-cmake_minimum_required(VERSION 3.14)
+cmake_minimum_required(VERSION 3.14...4.3)
 project(mold VERSION 2.41.0)
 
 include(CMakeDependentOption)
+include(CheckCXXSourceCompiles)
 include(CheckSymbolExists)
 include(GNUInstallDirs)
 
 add_executable(mold)
+# CXX_SCAN_FOR_MODULES exists in CMake 3.28+, and mold does not use C++ modules.
+set_property(TARGET mold PROPERTY CXX_SCAN_FOR_MODULES OFF)
 target_compile_features(mold PRIVATE cxx_std_20)
 
 if(MINGW)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -133,20 +136,88 @@ if(MOLD_USE_MSAN)
...
+    find_path(MOLD_LIBCXX_INCLUDE_DIR
+      NAMES string
+      HINTS ${MOLD_LIBCXX_INCLUDE_HINTS}
+      PATHS /usr/include/c++/v1 /usr/local/include/c++/v1)
+    if(NOT MOLD_LIBCXX_INCLUDE_DIR)
+      message(FATAL_ERROR "Could not find libc++ headers for MOLD_USE_LIBCXX")
+    endif()
+
+    target_compile_options(mold PRIVATE $<$<COMPILE_LANGUAGE:CXX>:-nostdinc++>)
+    target_compile_features(mold PRIVATE cxx_std_23)
```
a5ehren: this is an annoying thing that falls out of trying to use libcxx with GCC - some of the keywords used are implemented as part of GCC's c++23 support instead of c++20 like they are in clang

rui314:
Thanks, merged. But I don't think we need fine-grained flags like MOLD_STATIC_LIBSTDCXX to control how each library is linked. MOLD_MOSTLY_STATIC is for ./dist.sh, so as long as it works, we are fine. So I'll remove these flags.

a5ehren:
No problem, I didn't know if anyone could use them but it's easy enough to do during cmake config.

rui314:
I'll revert these changes, as they broke the s390x and ppc64le cross builds. I spent a few hours trying to fix the breakages but couldn't get them working. Since the original dist.sh worked just fine, I want to keep it as is.

a5ehren:
Oh, the qemu ones? I’ll try them and see if I can fix, and add them to the CI workflow.

a5ehren:
Oh, they quit publishing ppc64le and s390x for Bullseye after the 20240904 images. I totally get it if you don't want to mess with it anymore, but the fix is just to split these into their own block:

```
  ppc64le | s390x)
  # Debian 11 (Bullseye) was initially released on August 14th, 2021
  #
  # Note that this snapshot is the final Bullseye image released for ppc64le and s390x
  cat <<EOF | $image_build
FROM mirror.gcr.io/library/debian:bullseye-20240904@sha256:8ccc486c29a3ad02ad5af7f1156e2152dff3ba5634eec9be375269ef123457d8
ENV DEBIAN_FRONTEND=noninteractive TZ=UTC
RUN sed -i -e '/^deb/d' -e 's/^# deb /deb /g' /etc/apt/sources.list && \
  echo 'Acquire::Retries "10"; Acquire::http::timeout "10"; Acquire::Check-Valid-Until "false";' > /etc/apt/apt.conf.d/80-retries && \
  apt-get update && \
  apt-get install -y --no-install-recommends build-essential gcc-10 g++-10 clang-16 libc++-16-dev cmake && \
  ln -sf /usr/bin/clang-16 /usr/bin/clang && \
  ln -sf /usr/bin/clang++-16 /usr/bin/clang++ && \
  rm -rf /var/lib/apt/lists
EOF
  ;;
```

rui314:
GCC bootstrap is a one time thing, so I'm not interested in eliminating it unless there's another reason to do so.

a5ehren:
Fair enough, no problem.

PR #1617, opened 2026-07-02
Add (default on) --as-needed-gc flag to remove DSOs used only by GC'ed sections

agrieve:
`lld` does this by default. `bfd` and `gold` do not. See: https://crbug.com/529102889

This flag makes mold behave like `lld` when both `--as-needed` and `--gc-sections` are used. If a DSO is only referenced by sections that are garbage-collected, the DSO's `DT_NEEDED` entry will be omitted.


agrieve:
Note #1618 is the same but without a flag (I'm leaning towards that being better, but thought I'd leave it up to you)

PR #1618, opened 2026-07-02
Remove DSOs used only by GC'ed sections

agrieve:
lld does this by default. See: https://crbug.com/529102889

If a DSO is only referenced by sections that are garbage-collected, the DSO's DT_NEEDED entry will be omitted.


agrieve:
Same as #1617 except without an opt-out (wasn't sure what was better)

rui314:
I think this new behavior doesn't cause any practical issue so I'm fine with that.

PR #1620, opened 2026-07-12
Don't pull archive members for versioned DSO undefineds

### LINKED ISSUE #1619: PIE crashes at startup when linked with mold because libstdc++.a members are pulled into the executable
When linking a trivial C++ program with `mold` and `-lopenmpt`, the resulting PIE crashes during dynamic-linker initialization, before `main()` is reached. The same program linked with `bfd` or `lld` works fine.

The crash happens because `mold` extracts members from `libstdc++.a` into the PIE to satisfy undefined symbols from `libopenmpt.so`, while `libopenmpt.so` also loads `libstdc++.so.6` at runtime. The two copies of `libstdc++` conflict during initialization.

Used gcc 16.1.1+r346+g4e03491b401d-3, mold: 2.41.0-1 and libopenmpt 0.8.7-1 in archlinux.

## Minimal reproducer

```
echo "int main() { return 0; }" > /tmp/repro.cpp
c++ -O2 -g /tmp/repro.cpp -lopenmpt -fuse-ld=mold -o /tmp/repro_mold
/tmp/repro_mold
```

It will crash with SIGSEGV.

## Possiable cause

libopenmpt.so has undefined C++ symbols `__cxa_throw`. When mold processes the link command, it finds libstdc++.a in the GCC private directory before any shared version is considered, and pulls archive members into the PIE to satisfy those symbols.

At runtime, libopenmpt.so has DT_NEEDED libstdc++.so.6, so the dynamic linker also loads libstdc++.so.6. The PIE now contains a static copy of libstdc++ and a dynamic copy of libstdc++ simultaneously. The static copy initializes first and corrupts state, so when libstdc++.so initializes, std::ios_base::Init dereferences an invalid this pointer (0x43).

wszqkzqk:
SharedFile::parse() used to set versyms[i] to VER_NDX_GLOBAL for all undefined symbols in a DSO, discarding their real version index. Because of that, a versioned DSO undefined symbol such as foo@VERSION was resolved to an unversioned symbol in an archive, pulling in the archive member and statically linking it into the output. This could cause crashes at runtime when the DSO reference was actually meant to be resolved by another DSO at load time.

Preserve the version index for DSO undefined symbols and skip archive extraction in mark_live_objects() when the DSO undefined symbol is versioned. The dynamic linker will resolve it from the DSO's DT_NEEDED dependencies at runtime.

Fixes https://github.com/rui314/mold/issues/1619

Tested on Arch Linux x86_64.

### DIFF
--- a/src/input-files.cc
+++ b/src/input-files.cc
@@ -1386,10 +1386,9 @@ void SharedFile<E>::parse(Context<E> &ctx) {
 
     this->elf_syms2.push_back(esyms[i]);
 
-    // resolve_symbols only consults versyms[] for defined symbols
-    // (see SharedFile::resolve_symbols), so VER_NDX_GLOBAL is fine
-    // for undefined entries.
-    this->versyms.push_back(esyms[i].is_undef() ? (u16)VER_NDX_GLOBAL : ver);
+    // Keep the version index for undefined symbols as well; mark_live_objects
+    // needs it to avoid pulling archive members for versioned DSO undefineds.
+    this->versyms.push_back(ver);
 
     std::string_view name = this->symbol_strtab.data() + esyms[i].st_name;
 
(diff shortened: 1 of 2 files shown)


wszqkzqk:
Sorry, I forgot musl.

wszqkzqk:
> Sorry, I forgot musl.

Fixed now.

PR #1625, opened 2026-07-22
Simplify ShardedMap

rui314:
Fold the `on_create` callback into the main `gather`/`insert` interface: `insert()` now returns `std::pair<T *, bool>` and `gather()`'s callback receives `(payload, value, key, created)`, so callers initialize newly-created values themselves behind `if (created)`. This removes the second `gather()` overload and the `OnCreate` template parameter threaded through `Shard::insert`.

Also store `T` directly in the shard map instead of behind a `T *` into a side deque; `std::unordered_map` keeps value addresses stable across rehashing, so pointer stability is unchanged and the pool is redundant.

Performance properties are preserved: keys are still hashed exactly once (in `add()`), and `gather()` still pre-reserves each shard.

Testing:
- All 5694 tests pass (Release build).
- A/B benchmark vs baseline 1d306dc0, alternating runs, medians:
  - chrome-debug-x86_64: 2.76s → 2.77s
  - chrome-release-x86_64: 1.075s → 1.07s (10 runs each)
  - clang-release: 0.15s → 0.14s
- `--perf` phase times for gather_symbols/resolve_symbols on chrome-debug-x86_64 match within noise.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/lib/lib.h
+++ b/lib/lib.h
@@ -528,18 +528,15 @@ public:
 // add(), which only appends the key and a caller-supplied payload to
 // a thread-local buffer binned by the key's hash. gather() then ends
 // the phase: it deduplicates the recorded keys into per-shard hash
-// tables and invokes `callback(payload, value)` for each record,
-// where `value` is the T created for (or found for) the record's key.
+// tables and invokes `callback(payload, value, key, created)` for
+// each record, where `value` is the T created for (or found for) the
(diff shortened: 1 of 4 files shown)


rui314:
Opened prematurely; closing.

PR #1641, opened 2026-08-26
Support -z rewrite-endbr on arm64

### LINKED ISSUE #1552: Support -z rewrite-endbr on aarch64
Currently, mold only supports `-z rewrite-endbr` on x86_64 (the `ENDBR64` landing pad), but aarch64 has something similar called [BTI](https://developer.arm.com/community/arm-community-blogs/b/architectures-and-processors-blog/posts/enabling-pac-and-bti-on-aarch64). It would be nice if mold could support this.

I gave a try at implementing it, based on 17f0d85efb427681f70db4f46a88a17dc660649f, but for some reasons the `main` function keeps getting its landing pad removed, and I can't figure out why.

jvoisin:
So far -z rewrite-endbr was implemented only for x86-64, where it rewrites unnecessary ENDBR64 landing pads (Intel CET) with NOPs for functions whose addresses are not taken. arm64 has an equivalent control-flow-integrity feature, BTI, whose landing pad is `bti c`.

This commit generalizes rewrite_endbr into a template shared by both targets: the landing pad and NOP byte patterns are selected per-arch (ENDBR64/NOP on x86-64, `bti c`/`nop` on arm64). The whole-program analysis that preserves landing pads for address-taken functions is unchanged.

Unlike x86-64, arm64 needs range extension thunks, and a thunk reaches its target with an indirect branch (`br x16`). A function that is only called directly can therefore still be entered indirectly once its call is routed through a thunk, so we must keep its landing pad. We preserve the landing pad of every thunked symbol; the thunk symbol set has already been narrowed to the ones that actually need a thunk by remove_redundant_thunks().

Note that on arm64 with recent glibc, `main`'s landing pad is legitimately removed: the CRT passes a local __wrap_main stub (whose landing pad is kept) to __libc_start_main and reaches main through a direct branch, which does not require a landing pad.

Also note that code built with -mbranch-protection=standard (BTI + pointer authentication) starts functions with `paciasp`, which doubles as a landing pad, so there is no separate `bti c` to rewrite and this option has no effect on such code.

This was locally tested on an Apple M2 machine running Asahi Linux.

Fixes #1552.


rui314:
I happened to revisit this topic very recently, as I think this could be an interesting research area. `-z rewrite-endbr` can indeed remove many landing pads from the output binary. In some cases, I saw it remove more than 90% of them.

But the real question is how much this actually improves security. So I counted the total number of landing pads in an entire process image, rather than in a single executable or shared library. Unfortunately, the results didn't look nearly as good. If a binary is dynamically linked against libc and other libraries, those DSOs bring in a large number of additional valid indirect-branch targets. As a result, even when `-z rewrite-endbr` removes most landing pads from the main executable, the total number of landing pads in the process may decrease by only a few percent.

In other words, I think focusing on landing pads in a single linker output is somewhat missing the bigger picture, which I didn't realize when I first implemented it. What we really want is to minimize the set of valid indirect-branch targets across the entire process image in memory.

NOPout and NOPoutNG (https://ieeexplore.ieee.org/document/11418050) address essentially the same problem at load time. NOPout removes unused landing pads from shared libraries, while NOPoutNG starts with landing pads disabled and enables them when they are found to be needed. One thing I don't particularly like about this approach is that it still requires patching executable text at load time, potentially in many places.

I think a cleaner direction is to stop putting landing pads directly at the beginning of every global function, and instead separate the function's indirect-call entry point from the function body. Florian Weimer proposed something along these lines in 2020 on the x86-64 ABI mailing list (https://groups.google.com/g/x86-64-abi/c/iQWEW-iW8DQ). I think that is the right direction: we should disable indirect calls by default, and have the dynamic linker consider the entire process image and create code stubs (similar to PLT entries) for indirect calls only for functions whose addresses are taken anywhere in the process.

jvoisin:
> But the real question is how much this actually improves security. So I counted the total number of landing pads in an entire process image, rather than in a single executable or shared library. Unfortunately, the results didn't look nearly as good. If a binary is dynamically linked against libc and other libraries, those DSOs bring in a large number of additional valid indirect-branch targets. As a result, even when -z rewrite-endbr removes most landing pads from the main executable, the total number of landing pads in the process may decrease by only a few percent.

What if every library/executable in the address space of the process were using `-z rewrite-endbr`? Shouldn't it remove ~90% of the landing pads?

I do like Weimer's approach, and I think it's the right one. In the meantime, reducing the amount of landing pads by a significant percentage is still interesting when paired with type-based CFI, like [FineIBT](https://dustri.org/b/paper-notes-fineibt.html), to reduce the amount of same-type functions.

rui314:
> What if every library/executable in the address space of the process were using -z rewrite-endbr? Shouldn't it remove ~90% of the landing pads?

That's a good question, and I don't have the numbers on hand to answer it quickly. But currently, all functions exported from a shared library must start with a landing pad, because a program using the library may take the address of any exported function. So the attack surface is huge.

I think we should merge this change, as we already have the option and there may be valid uses for it. That being said, this option could give users who enable it a false sense of security. So I think we should explore a better approach at the same time.

I'll review this tomorrow. Thank you for your patience!

PR #1647, opened 2026-09-02
Guard parse_sframe() against a zero-size .sframe section

### LINKED ISSUE #1646: parse_sframe() reads past a zero-size .sframe section, reported as "corrupted" (Scrt1.o on Ubuntu 25.10 / binutils 2.45)
### Description

mold 2.42.0 fails with `corrupted .sframe section` when linking against the
system `Scrt1.o` on Ubuntu 25.10 (binutils 2.45 / GNU as 2.45), even for a
trivial program:

```
$ cat repro.c
int main(void) { return 0; }

$ gcc repro.c -fuse-ld=mold
mold: fatal: /usr/lib/x86_64-linux-gnu/Scrt1.o:(.sframe): corrupted .sframe section
```

This is on 2.42.0, which is *after* #1511 was fixed by d0aa95ab ("Support
the SFrame v3 stack-unwinding format"), so it's a different, newer bug —
not a lack of SFrame v3 support.

### Root cause

`ObjectFile<E>::parse_sframe()` (`src/input-files.cc`) reads the section's
raw bytes and immediately casts them to `SFrameHeader<E>`, with no check
that the section is actually large enough to hold one:

```cpp
std::string_view data = this->get_string(ctx, isec->shdr());
const SFrameHeader<E> &hdr = *(const SFrameHeader<E> *)data.data();

if (hdr.magic != SFRAME_MAGIC)
  Fatal(ctx) << *isec << ": corrupted .sframe section";
```

On this toolchain, `Scrt1.o`'s `.sframe` section has `sh_size == 0` — GNU as
2.45 emits an `SHT_GNU_SFRAME` section header for it, but with zero content,
instead of either omitting the section entirely or writing a minimal valid
header (`num_fdes == 0`). Compare against the other CRT startup objects on
the same system, where `.sframe` is populated normally:

```
$ readelf -SW /usr/lib/x86_64-linux-gnu/crt1.o  | grep -i sframe
  [ 6] .sframe           GNU_SFRAME      0000000000000000 0000e0 000033 00   A  0   0  8
  [ 7] .rela.sframe      RELA            0000000000000000 0003b0 000018 18   I 14   6  8

$ readelf -SW /usr/lib/x86_64-linux-gnu/gcrt1.o | grep -i sframe
  [ 6] .sframe           GNU_SFRAME      0000000000000000 000158 000055 00   A  0   0  8
  [ 7] .rela.sframe      RELA            0000000000000000 000728 000030 18   I 14   6  8

$ readelf -SW /usr/lib/x86_64-linux-gnu/Scrt1.o | grep -i sframe
  [ 6] .sframe           GNU_SFRAME      0000000000000000 0000a0 000000 00   A  0   0  8
```

`Scrt1.o`'s `.sframe` is 0 bytes and has no `.rela.sframe`. Because
`get_string()` just returns the raw `[sh_offset, sh_offset + sh_size)`
slice of the file with no bounds check against `sizeof(SFrameHeader<E>)`,
`data.data()` still points at a valid file offset (the start of whatever
section follows), and `parse_sframe()` happily reads a `SFrameHeader<E>`
out of those unrelated bytes. The `magic` field of that garbage doesn't
match `SFRAME_MAGIC`, so it's reported as "corrupted" — but the section
isn't corrupted, it's just empty, and mold reads past its end.

Environment:
- Ubuntu 25.10
- GNU as / GNU ld (Ubuntu Binutils) 2.45
- gcc (Ubuntu 15.2.0-4ubuntu4) 15.2.0
- mold 2.42.0 (also reproduces built from current `main`)

### Fix

Treat a `.sframe` section shorter than `sizeof(SFrameHeader<E>)` the same
way the existing version check already treats an old-format section: skip
it instead of reading it. Patch, and a branch with it applied and verified
against this exact repro:

https://github.com/MrSplinterRat/mold/tree/fix-empty-sframe-section

```diff
--- a/src/input-files.cc
+++ b/src/input-files.cc
@@ -805,6 +805,15 @@ void ObjectFile<E>::parse_sframe(Context<E> &ctx) requires supports_sframe<E>
     isec->kill();
 
     std::string_view data = this->get_string(ctx, isec->shdr());
+
+    // Some assemblers (e.g. GNU as 2.45 on Scrt1.o) emit a placeholder
+    // SHT_GNU_SFRAME section with sh_size == 0 instead of either omitting
+    // the section or writing a valid empty header (num_fdes == 0). Treat
+    // that the same as "nothing to contribute" rather than reading past
+    // the end of the section as if it were a real header.
+    if (data.size() < sizeof(SFrameHeader<E>))
+      continue;
+
     const SFrameHeader<E> &hdr = *(const SFrameHeader<E> *)data.data();
 
     if (hdr.magic != SFRAME_MAGIC)
```

With this applied, the same repro links and runs cleanly:

```
$ gcc repro.c -B<dir-with-ld.mold-pointing-at-patched-binary> -fuse-ld=mold -o repro
$ ./repro; echo $?
0
```

Happy to open this as a PR if that's preferred over the patch inline here.

MrSplinterRat:
Fixes #1646.

### Problem

`ObjectFile<E>::parse_sframe()` casts a `.sframe` section's raw bytes
straight into `SFrameHeader<E>` with no check that the section actually
holds enough bytes for one. On Ubuntu 25.10 (GNU as / GNU ld 2.45), GNU as
emits a placeholder `SHT_GNU_SFRAME` section with `sh_size == 0` for
`/usr/lib/x86_64-linux-gnu/Scrt1.o` — instead of either omitting the
section or writing a minimal valid header (`num_fdes == 0`). mold then
reads a "header" out of whatever bytes happen to follow the (empty)
section in the file, its `magic` field doesn't match `SFRAME_MAGIC`, and
the link aborts with a misleading error:

```
$ gcc repro.c -fuse-ld=mold
mold: fatal: /usr/lib/x86_64-linux-gnu/Scrt1.o:(.sframe): corrupted .sframe section
```

Full repro, root-cause writeup and a `readelf` comparison against the
(correctly populated) `.sframe` in `crt1.o`/`gcrt1.o` on the same system
are in #1646.

### Fix

Skip `.sframe` sections shorter than `sizeof(SFrameHeader<E>)` the same
way the existing version check already skips an old-format section,
instead of reading past their end.

### Testing

Built and verified against the exact repro from #1646:

```
$ gcc repro.c -fuse-ld=mold -o repro   # was: mold: fatal: ...: corrupted .sframe section
$ ./repro; echo $?
0
```

Also ran the full `ctest` suite against the patched build: **100% tests
passed out of 491** (476 run, 15 skipped, 0 failed). `test/sframe.sh` and
`test/arch-x86_64-sframe-unwind.sh` are among the skipped ones on this
machine — their `readelf --sframe` probe for a Version-3-capable `as`
doesn't match here, so they don't exercise real `.sframe` content in this
environment — but nothing else regressed.

### DIFF
--- a/src/input-files.cc
+++ b/src/input-files.cc
@@ -805,6 +805,15 @@ void ObjectFile<E>::parse_sframe(Context<E> &ctx) requires supports_sframe<E> {
     isec->kill();
 
     std::string_view data = this->get_string(ctx, isec->shdr());
+
+    // Some assemblers (e.g. GNU as 2.45 on Scrt1.o) emit a placeholder
+    // SHT_GNU_SFRAME section with sh_size == 0 instead of either omitting
+    // the section or writing a valid empty header (num_fdes == 0). Treat
+    // that the same as "nothing to contribute" rather than reading past
+    // the end of the section as if it were a real header.
+    if (data.size() < sizeof(SFrameHeader<E>))
+      continue;
+
     const SFrameHeader<E> &hdr = *(const SFrameHeader<E> *)data.data();
 
     if (hdr.magic != SFRAME_MAGIC)


rui314:
I thought this was a bug report rather than a pull request, so I already fixed it and committed the fix myself. Sorry about that! I'll close this PR since it's already fixed, but I really appreciate the contribution.

PR #1650, opened 2026-09-05
Assign visibility to Symbol objects for linker-synthesized symbols

ziyao233:
When creating linker-synthesized symbols, scope/visibility properties are only applied to corresponding ELF symbols, not the Symbol objects. However, to_output_esym() decides their scope/visibility solely based on properties of the Symbol objects.

Thus after 35f4a334a0d0 ("Do not demote non-exported globals to local symbols in .symtab") removes the logic that demotes non-exported global symbols to local ones, linker synthesized symbols become STB_GLOBAL and STV_DEFAULT in the result binary, breaking ABI checkers of some applications.

Assign STV_HIDDEN to Symbol objects as well when creating linker-synthesized symbols, and copy the visibility when converting them to the ELF form in to_output_esym(). RISC-V-specific "__global_pointer$" needs specical treatment since it's required to be exported in dynamic-linked executables with GP-relocations.

### DIFF
--- a/src/output-chunks.cc
+++ b/src/output-chunks.cc
@@ -2161,6 +2161,7 @@ to_output_esym(Context<E> &ctx, Symbol<E> &sym, u32 st_name, U32<E> *shn_xindex)
     // Linker-synthesized symbols
     shndx = osec->shndx;
     esym.st_value = sym.get_addr(ctx);
+    esym.st_visibility = sym.visibility;
   } else if (SectionFragment<E> *frag = sym.get_frag()) {
     // Section fragment
     shndx = frag->get_output_section(ctx).shndx;
--- a/src/passes.cc
+++ b/src/passes.cc
@@ -955,6 +955,7 @@ void add_synthetic_symbols(Context<E> &ctx) {
 
     Symbol<E> *sym = get_symbol(ctx, name);
     sym->value = 0xdeadbeef; // unique dummy value
+    sym->visibility = STV_HIDDEN;
     obj.symbols.emplace_back(sym);
     return sym;
   };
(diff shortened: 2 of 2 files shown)


ziyao233:
After making symbols STV_HIDDEN and STB_LOCAL, binutils objdump doesn't print labels for them anymore, breaking the ppc64le tests...

ld.lld tends to make synthesized symbols hidden, which seems to match mold's original intention as well, but ld.bfd only makes them local. Not sure whether we should make them STV_DEFAULT, too, or find a way to force objdump to print labels for hidden local symbols.

ziyao233:
> After making symbols STV_HIDDEN and STB_LOCAL, binutils objdump doesn't print labels for them anymore, breaking the ppc64le tests...
> 
> ld.lld tends to make synthesized symbols hidden, which seems to match mold's original intention as well, but ld.bfd only makes them local. Not sure whether we should make them STV_DEFAULT, too, or find a way to force objdump to print labels for hidden local symbols.

So binutils considers hidden, local symbols with STT_NOTYPE not suitable for disassembling...,

```C
// opcodes/ppc-dis.c

static bool
ppc_symbol_is_valid (asymbol *sym,
                     struct disassemble_info *info ATTRIBUTE_UNUSED)
{
  elf_symbol_type * est;

  if (sym == NULL)
    return false;

  est = elf_symbol_from (sym);

  /* Ignore ELF hidden, local, no-type symbols.
     These are generated by annobin.  */
  if (est != NULL
      && ELF_ST_VISIBILITY (est->internal_elf_sym.st_other) == STV_HIDDEN
      && ELF_ST_BIND (est->internal_elf_sym.st_info) == STB_LOCAL
      && ELF_ST_TYPE (est->internal_elf_sym.st_info) == STT_NOTYPE)
    return false;

  return true;
}
```

rui314:
Thank you for your PR. 35f4a334a0d0a5f93eced1b521df8ca3eccdb2cf was meant to change the visibility of synthecized symbols. They had been in local in .symtab for years. Rather than assigning visibility to them, I think we should simply restore the original behavior.

rui314:
450832db6b7afaed621c9708bc733afd2edf821b should restore the original behavior.

PR #1660, opened 2026-09-19
Use `<[T]>::as_chunks` instead of `<[T]>::chunks_exact`

noritada:
This commit simply makes minor improvements to the code in terms of type.

Since `<[T]>::chunks_exact` returns a slice, the information about the chunk size is lost from the type, so even if you write `p[3]` as `p[4]` when `p` has a length of 4, the code will still compile.
Using `<[T]>::as_chunks` makes it clear that the type is a fixed-length array (`&[u8; 4]`), so references to out-of-bounds indices will be caught at compile time.


rui314:
Thanks!

PR #1662, opened 2026-09-20
Read the loop-invariant relocation addresses once

nmoinvaz:
**Read the loop-invariant relocation addresses once**

Applying a relocation reads two addresses that cannot change while the loop runs, the input section's address in its output section and the GOT base, yet both were read again for every relocation. `InputSection::offset` is an `AtomicU64` and `Context` carries interior mutability, so it is not `Freeze` and `&Context` reaches LLVM without `noalias`, leaving no way to hoist either read past the stores into the output buffer. Reading them up front drops nine instructions per relocation, among them four dependent loads, an atomic load and a bounds check. Every target had the pattern, as did `apply_abs_rels`, which also re-read `apply_dynamic_relocs` and `pack_dyn_relocs_relr`.

**Compute the x86-64 GOT displacement only where it is used**

G was computed eagerly although only six of the fifty relocation kinds use it, so every relocation paid for an aux-table lookup and its bounds check. A closure moves that off the common path, as the other twelve targets already do.

**Skip x86-64 absolute relocations the output section applies**

`scan_abs_relocations` collects every `R_X86_64_64` in an allocated section and `apply_abs_rels` applies it, so `apply_reloc_alloc` resolved the symbol and computed S, P and G only to fall through an empty match arm. The skip is free, since `R_X86_64_NONE` is 0 and `R_X86_64_64` is 1 and LLVM folds the two tests into one unsigned compare. It stays on x86-64 alone: it transfers unchanged to i386, arm32, arm64, m68k, s390x and sh4, whose `R_ABS` arm is likewise empty or a bare `continue`, but ppc32, ppc64v2 and sparc64 share that arm with other kinds, and unlike the hoists this is not a pure common-subexpression change.

Measured on generated inputs sized so one relocation kind dominates one output section, median of fifteen single-threaded runs of that section's own `--perf` timer, on an Apple M5:

| Section | Relocations | Before | After |
|---------|-------------|--------|-------|
| `.text` | 7.0M `PC32`, `PLT32`, `REX_GOTPCRELX` | 91ms | 82ms |
| `.rodata` | 3.8M `R_X86_64_64` | 42ms | 21ms |

A whole link of the pointer-table inputs goes from 190ms to 174ms on one thread and 80ms to 76ms on ten. A link dominated by neither shape stays inside the noise.

Output is unchanged. Across thirty-three combinations of input set and link mode, covering `-pie`, `-shared` (with `-Bsymbolic` where the inputs need it), `-no-pie`, `-r`, `--gc-sections`, `--icf=all`, `--emit-relocs`, `--strip-all` and `-z pack-relative-relocs`, the linked files are byte for byte identical apart from the mold version string in `.comment` and the build ID that hashes it.


rui314:
Nice. Thanks!

PR #1663, opened 2026-09-20
Check symbol types without loading the owner's symbol entry

nmoinvaz:
`check_symbol_types` read `sym.esym(ctx)` for every symbol a file references but does not own, which loads the ELF symbol entry from the owning file's symbol table, a second cache miss per reference on top of the symbol lookup itself, only to get `st_type`. `Symbol` already caches that in `type_and_bind`, which `set_esym` keeps in step wherever a symbol is pointed at a new entry, and which `Symbol::ty` already relies on.

Measured on generated inputs whose 8.1M symbol references scatter across 1.5M symbols, median of nine runs of the pass's own `--perf` timer, on an Apple M5:

| Threads | Before | After |
|---------|--------|-------|
| 1 | 321ms | 146ms |
| 10 | 51ms | 26ms |

A link whose symbols stay cache-resident is unchanged, at 4ms either way.

The warning is unchanged: the scenario from `tests/warn-symbol-type.sh`, an object named `times` against a shared-library function, prints the same text with both builds. Across eighteen combinations of input set and link mode, all of which link, the outputs are byte for byte identical apart from the mold version string in `.comment` and the build ID that hashes it.


rui314:
Nice. How did you find it?

nmoinvaz:
Code analysis with Claude

PR #1664, opened 2026-09-20
Count the GOT's dynamic relocations in parallel

nmoinvaz:
`got::num_dynrels` walks every symbol in `got_syms` on one thread, a cache miss each in a large link, and it runs three times on the way to the final layout: `compute_section_headers` calls `update_shdr` on every chunk twice, and `update_reldyn` sizes `.rela.dyn` once more. At ten threads those two phases were the only ones before the copy that used a single CPU. The count is a pure function of each symbol, so `par_iter().filter().count()` replaces the loop over `got_syms`; the three TLS lists are short and stay serial.

Measured with `-pie` on generated inputs with 1.48M GOT entries, median of nine runs of each phase's own `--perf` timer at ten threads, on an Apple M5:

| Phase | Before | After |
|-------|--------|-------|
| `compute_section_headers` | 38ms | 10ms |
| `update_reldyn` | 20ms | 5ms |

On one thread `compute_section_headers` is unchanged, 72ms against 73ms.

Across eighteen combinations of input set and link mode, all of which link, the outputs are byte for byte identical apart from the mold version string in `.comment` and the build ID that hashes it.

`got::copy_buf` and `got::write_dynrels` walk the same list serially through `for_each_entry` and are left alone here. GOT indices are assigned in `got_syms` order, so each run of symbols owns a contiguous slice of the buffer and they could be split the same way.


rui314:
Thanks!

nmoinvaz:
Thanks for the quick turnaround!

PR #1666, opened 2026-09-20
Fix AArch64 TLSLE LDST relocations

artemkulyk:
mold could not link AArch64 objects using the load/store forms of the local-exec TLS offset:

    unknown relocation: R_AARCH64_TLSLE_LDST64_TPREL_LO12_NC

Found while linking ClickHouse for aarch64 (libloggers.a, ExtendedLogMessage.cpp.o); lld handles the whole family.

This adds the five relocations to the same dispatch as R_AARCH64_TLSLE_ADD_TPREL_LO12_NC and encodes bits 11..N of the offset for a 2^N-byte access.

The test stores and loads 1, 2, 4, 8 and 16-byte thread-local variables and checks the values.


rui314:
Thank you!

PR #1667, opened 2026-09-20
Warn on truncated .eh_frame records instead of panicking

artemkulyk:
Truncated .eh_frame records made mold panic (slice index out of range):

    .eh_frame: 03 00 00 00 00 00 00

The length check in parse_ehframe and the CIE augmentation parser in parse_fde_encoding now warn and drop the record instead of slicing past the end of the section.  lld reports the same condition as "corrupted .eh_frame".

The test covers a short record, a record whose length extends past the section, and a CIE too small to parse.


rui314:
C++ mold used to simply crash with a segmentation fault on corrupted input, and we considered that acceptable. With Rust, however, memory safety is much easier to achieve, so I think we should treat any segmentation fault as a security bug and pay more attention to safety.

That said, whether we should produce detailed error messages for every kind of corrupted input is a separate question. I believe there are many other places where malformed input can currently cause a panic, and making every one of them completely bulletproof seems a bit too tedious.

It might be better to have the panic handler report what mold was processing when the panic occurred. Let me think about this a little more.

PR #1669, opened 2026-09-21
Check a file's own symbol entry before looking the symbol up

nmoinvaz:
`check_duplicate_symbols` and `check_symbol_types` walk every global symbol reference of every file and looked each one up in the symbol table first, a cache miss each once the table outgrows the cache. Both then ran tests on the file's own ELF symbol entry that skip the reference regardless of what the lookup found. Most of a file's global entries are undefined references, so most lookups were wasted.

In `check_duplicate_symbols` the undefined, common and weak tests were already in the same `continue` condition as the ownership tests, listed after the lookup; they now run before it. In `check_symbol_types` a reference whose entry has no type can never mismatch, since the check requires both types to differ from `STT_NOTYPE`, so that test now comes before the lookup.

Measured on generated inputs whose 8.1M symbol references scatter across 1.5M symbols, median of eleven runs of each pass's own `--perf` timer, on an Apple M5:

| Pass | Threads | Before | After |
|------|---------|--------|-------|
| `check_symbol_types` | 1 | 149ms | 43ms |
| `check_symbol_types` | 10 | 25ms | 7ms |
| `check_duplicate_symbols` | 1 | 140ms | 46ms |
| `check_duplicate_symbols` | 10 | 29ms | 9ms |

On inputs whose symbols stay cache-resident both passes are unchanged:

| Pass | Threads | Before | After |
|------|---------|--------|-------|
| `check_symbol_types` | 1 | 4ms | 5ms |
| `check_duplicate_symbols` | 1 | 5ms | 4ms |

Both diagnostics still fire with the same text: a duplicate definition across two objects reports `duplicate symbol`, and the scenario from `tests/warn-symbol-type.sh`, an object named `times` against a shared-library function, reports `symbol type mismatch`. Across eighteen combinations of input set and link mode, all of which link, the outputs are byte for byte identical apart from the mold version string in `.comment` and the build ID that hashes it.


nmoinvaz:
I recommend to install Coderabbit AI or some other free AI bot on your repo.

